PR #28, opened 2022-01-19
Require more builds to pass before bors merges

msimberg:
Require all current GitHub actions linux builds to pass and the non-test CircleCI checks (i.e. building core and various static checks).


msimberg:
bors merge

PR #30, opened 2022-01-19
Fix specification of bors checks

msimberg:
The catchall `%` means that when _any_ (one) matching build has succeeded bors will decide to merge. What we're looking for is all matching builds to succeed. This changes the list to an explicit list of the required builds.


msimberg:
bors merge

PR #35, opened 2022-01-19
Remove spellcheck from required bors checks

msimberg:
It's a guideline and can be broken in certain cases (and it sometimes reports false positives). This removes it as a requirement when bors merging.


msimberg:
bors merge

PR #42, opened 2022-01-20
Fix missing inline for `pika::mpi::experimental::detail::error_message`

msimberg:
An inline keyword would fix it as well, but this moves the function into a
source file and `PIKA_EXPORT`s the function instead. I've also moved the other functions in `mpi_exception` to the source file.


msimberg:
Since there is currently no MPI builder:

bors merge

aurianer:
Thanks :)

PR #43, opened 2022-01-20
Fix reference file names for performance CI

msimberg:
They still had `local` in their names even though the tests themselves had `local` removed from their names.


msimberg:
bors merge

PR #51, opened 2022-01-21
Rename inspect_local to inspect

msimberg:
This only renames the binary/target. Folder names still contain `Local`, see #50.


msimberg:
bors merge

PR #61, opened 2022-01-21
Don't propagate private build flags to consumers

msimberg:
The private flags were mistakenly propagated out from modules.


msimberg:
bors merge

PR #62, opened 2022-01-21
Build header tests in parallel on CircleCI

msimberg:
I'm hoping this will slightly speed up the time spent in the header tests job, without OOM-ing.


msimberg:
Looks like the speedup is almost exactly 3x (and no OOMs...). Merging.

bors merge

PR #71, opened 2022-01-27
Don't try to take lock when getting work counts from MPI/CUDA polling

### LINKED ISSUE #41: Don't take a lock when getting the polling event counts for CUDA and MPI
Not strictly a bug, but taking the lock (especially trying to take the lock) means that shutdown can take longer or a very long time.

msimberg:
Fixes #41. Instead uses atomic counts for the vector/queue sizes. This means that `get_work_count` will never accidentally return zero if it can't take the lock. Related to https://github.com/eth-cscs/DLA-Future/issues/448.

Still waiting for some DLA-Future runs, but this at least improves (not necessarily fixes) the shutdown behaviour in miniapps.

### DIFF
--- a/libs/pika/async_mpi/src/mpi_future.cpp
+++ b/libs/pika/async_mpi/src/mpi_future.cpp
@@ -82,8 +82,7 @@ namespace pika { namespace mpi { namespace experimental {
         {
             get_requests_vector().push_back(req_callback.request);
             get_request_callback_vector().push_back(std::move(req_callback));
-            get_mpi_info().requests_vector_size_ =
-                static_cast<std::uint32_t>(get_requests_vector().size());
+            ++(get_mpi_info().active_requests_vector_size_);
 
             if constexpr (mpi_debug.is_enabled())
             {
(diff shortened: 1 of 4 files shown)


aurianer (approved):
LGTM thanks!

libs/pika/async_mpi/src/mpi_future.cpp
```diff
@@ -306,6 +305,8 @@ namespace pika { namespace mpi { namespace experimental {
                 // is only used for an assert
                 request_callback_vector[std::size_t(index)].request =
                     MPI_REQUEST_NULL;
+
+                --(detail::get_mpi_info().requests_vector_size_);
```
aurianer, line 309:
```suggestion
                --(detail::get_mpi_info().active_requests_vector_size_);
```
? as now we don't count the null requests in the vector anymore

aurianer (approved):

libs/pika/async_mpi/src/mpi_future.cpp
msimberg: Renamed the variable.

msimberg:
test cscs

msimberg:
bors merge

PR #77, opened 2022-01-27
Ignore spack files

msimberg:
When using `spack develop` spack creates various `spack-*` files and directories in the root of the repo that can be ignored.


msimberg:
bors merge

PR #81, opened 2022-01-31
Properly handle types of forwarded values

hkaiser:
This fixes use cases like the following:
```
    {

        std::atomic<bool> set_value_called{false};
        std::string str{"hello"};
        int x = 3;
        auto s = ex::just(str, x);
        auto f = [](std::string& str, int& x) {  // <-- note the reference arguments
            PIKA_TEST_EQ(str, std::string("hello"));
            PIKA_TEST_EQ(x, 3);
        };
        auto r = callback_receiver<decltype(f)>{f, set_value_called};
        auto os = ex::connect(std::move(s), std::move(r));
        ex::start(os);
        PIKA_TEST(set_value_called);
    }
```

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/just.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/just.hpp
@@ -84,7 +84,7 @@ namespace pika { namespace execution { namespace experimental {
                         [&]() {
                             pika::execution::experimental::set_value(
                                 PIKA_MOVE(os.receiver),
-                                PIKA_MOVE(os.ts).template get<Is>()...);
+                                PIKA_FORWARD(Ts, os.ts.template get<Is>())...);
                         },
                         [&](std::exception_ptr ep) {
                             pika::execution::experimental::set_error(


msimberg:
Are you sure this should be allowed? See https://brycelelbach.github.io/wg21_p2300_std_execution/std_execution.html#spec-execution.senders.just (specifically the `start_t` overload), and https://brycelelbach.github.io/wg21_p2300_std_execution/std_execution.html#design-sender-factory-just. At least I'm pretty sure the spec says the values should be moved, but then the next question is of course if you think the spec should change? I'd be worried that allowing references there allows all kinds of dangers (that one can opt in to with `reference_wrapper` already now).

hkaiser:
> Are you sure this should be allowed? See https://brycelelbach.github.io/wg21_p2300_std_execution/std_execution.html#spec-execution.senders.just (specifically the `start_t` overload), and https://brycelelbach.github.io/wg21_p2300_std_execution/std_execution.html#design-sender-factory-just. At least I'm pretty sure the spec says the values should be moved, but then the next question is of course if you think the spec should change? I'd be worried that allowing references there allows all kinds of dangers (that one can opt in to with `reference_wrapper` already now).

Fair points. Apparently, I misread the specs... And no, I'm not planning to propose changing the specs ;-)

PR #85, opened 2022-02-02
Set conservative timeout on build steps in Jenkins

msimberg:
To avoid jobs like these taking 13 hours: https://lisone.cscs.ch/blue/organizations/jenkins/ssl%2Fpika/detail/pika/26/pipeline.

The underlying problem of the filesystem quota being full should of course also be addressed...


msimberg:
bors merge

PR #91, opened 2022-02-03
Add warmup iterations to foreach performance test

msimberg:
This is a bit of a shot in the dark. The `perftests_report` utility already does one warmup iteration which I would've thought would be enough, but it's typically the `scheduler_executor` test on `foreach_report` which fails, and it's the first executor to run.

Maybe fixes #84.

PR #92, opened 2022-02-03
Update the reference of the stream bench for fork_join_executor

### LINKED ISSUE #84: Update performance test references
The current references seem to be somewhat too strict. Alternatively, can we slightly relax the criteria (without missing performance regressions).

aurianer:
Maybe fixes #84


msimberg:
launch perftests

aurianer:
launch perftests

aurianer:
The stream for the fork_join_executor looks at least more stable, will have a look at the parallel_executor now

msimberg:
launch perftests

msimberg:
Should we go ahead with this one as it is or should we still also update the foreach reference? It just about failed once, and failed once badly, but I'm not sure if the latter was with or without #91?

aurianer:
I think we can go ahead with it like that cause looks like the last run passed, if it fails more often I can still update it in a separate PR

msimberg:
bors merge

PR #95, opened 2022-02-04
Update the CMake MPI setup

aurianer:


msimberg (approved):
Thanks! Since we don't currently have any CI configurations with MPI did you test this at least on one system?

aurianer:
I tested the PR that was opened against the hpx repo but now that I tested this one after the daint upgrade there is a segfault that I have to investigate

aurianer:
Now tested and works :D

msimberg:
bors merge

PR #100, opened 2022-02-08
Update foreach benchmark reference for `scheduler_executor`

msimberg:
Updated with the run from https://github.com/pika-org/pika/pull/99#issuecomment-1032460258. Part of #84.

PR #101, opened 2022-02-08
Update reference foreach scheduler executor

### LINKED ISSUE #84: Update performance test references

aurianer:
I didn't take the very bad (---) but the bad (-) reference to update, so it still might fail once in a while
Maybe fixes #84


aurianer:
Closing, duplicate of #100

PR #105, opened 2022-02-09
Return thread data by const ref to avoid stray copies

biddisco:
The thread num tss data is used by the schedulers and does
not change frequently, returning by const ref saves a copy
in some cases.


libs/pika/threading_base/src/thread_num_tss.cpp
```diff
@@ -61,7 +61,7 @@ namespace pika { namespace threads { namespace detail {
         thread_nums_tss() = t;
     }
 
-    thread_nums get_thread_nums_tss()
+    const thread_nums& get_thread_nums_tss()
```
msimberg, line 64: How about replacing this with this definition instead: https://github.com/pika-org/pika/blob/79866e4c6458a258a9c1e4535ed87274f0f4f137/libs/pika/threading_base/src/thread_num_tss.cpp#L18-L23? I.e. keep the `get_thread_nums_tss` name, but but the `static thread_local` variable inside this function, and then use `get_thread_nums_tss` instead of `thread_nums_tss` in these spots: https://github.com/pika-org/pika/blob/79866e4c6458a258a9c1e4535ed87274f0f4f137/libs/pika/threading_base/src/thread_num_tss.cpp#L28?

Finally, you can replace these accesses: https://github.com/pika-org/pika/blob/0ab7223c4e08a9888bca38260ccb53db5bd0c2c0/libs/pika/schedulers/include/pika/schedulers/shared_priority_queue_scheduler.hpp#L328-L331 with `get_global_thread_num_tss` etc.

biddisco: I don't really know what you're trying to say.

msimberg: With your current change `get_thread_nums_tss` has exactly the same signature as `thread_nums_tss` and all `get_thread_nums_tss` does is call `thread_nums_tss`. So I'm suggesting to remove the unnecessary function call.

The second suggestion is that you would also avoid the potential copy by using the helper functions `get_global_thread_num_tss` etc.

libs/pika/threading_base/include/pika/threading_base/thread_num_tss.hpp
```diff
@@ -39,8 +27,57 @@ namespace pika { namespace threads { namespace detail {
         std::size_t thread_pool_num;
     };
 
-    PIKA_EXPORT void set_thread_nums_tss(const thread_nums&);
-    PIKA_EXPORT thread_nums get_thread_nums_tss();
+    inline static thread_local thread_nums thread_nums_tss_ = {
+        std::size_t(-1), std::size_t(-1), std::size_t(-1)};
```
msimberg, line 31: Much better, but this needs to stay in the source file. The thread numbers will otherwise be TU-specific.

biddisco: I should probably make the non `detail` namespace functions non inline, since they might be accessed from across a shared library boundary and the standard isn't clear on how an inline var behaves in that case. I'll push another patch just to be safe.

msimberg: TIL! Thanks. However, it needs to be `inline`, not `inline static`.

msimberg: And I'm pretty sure the functions have to stay as `inline`.

biddisco:
Internal stuff can use `detail::` inlined functions, public facing code should use the stuff compiled into the main library to ensure there's no chance of a multiple definition issue

msimberg:
> Internal stuff can use `detail::` inlined functions, public facing code should use the stuff compiled into the main library to ensure there's no chance of a multiple definition issue.

I'm not sure I follow the logic, but it looks good to me now.

> If possible, please trigger a perf report

The perftests are triggered automatically. It just doesn't compile at the moment so you're not getting a report. See e.g. https://github.com/pika-org/pika/runs/5128951660?check_suite_focus=true.

biddisco:
oops. I fixed that, but forgot to commit it.

biddisco:
I thought I'd better run tests to make sure that no errors are introduced, and I see quite a lot of segfaults on the following tests
```
 9/13 Test #452: tests.unit.modules.resource_partitioner.suspend_pool_external .......***Exception: SegFault 15.51 sec
      Start 453: tests.unit.modules.resource_partitioner.suspend_runtime
10/13 Test #453: tests.unit.modules.resource_partitioner.suspend_runtime .............***Exception: SegFault 15.96 sec
      Start 454: tests.unit.modules.resource_partitioner.suspend_thread
11/13 Test #454: tests.unit.modules.resource_partitioner.suspend_thread ..............***Exception: SegFault 12.75 sec
      Start 455: tests.unit.modules.resource_partitioner.suspend_thread_external
      Start 447: tests.unit.modules.resource_partitioner.scheduler_binding_check
 4/13 Test #447: tests.unit.modules.resource_partitioner.scheduler_binding_check .....***Exception: SegFault  2.36 sec
      Start 452: tests.unit.modules.resource_partitioner.suspend_pool_external
 9/13 Test #452: tests.unit.modules.resource_partitioner.suspend_pool_external .......***Exception: SegFault 16.21 sec
      Start 454: tests.unit.modules.resource_partitioner.suspend_thread
11/13 Test #454: tests.unit.modules.resource_partitioner.suspend_thread ..............***Exception: SegFault 12.72 sec
```
but this happens on the `main` branch as well as the thread_tss branch.

Are these known problems?

msimberg:
> Are these known problems?

Not to me or CI at least. Does it only happen with the resource tests or everything?

Unrelated to this PR but I did break the spack stuff for Jenkins here. I'll try to fix that ASAP so you get the rest of the configurations running here as well.

biddisco:
> Not to me or CI at least. Does it only happen with the resource tests or everything?

I ran a full test suite locally, only one thread suspend test failed, then I ran the `ctest -R tests.unit.modules.resource_partitioner` several times and saw multiple fails - switching back to main branch, I see basically the same fails, randomly, different ones each time I run the tests, but the same 'group" of fails. I will wipe my build tree and do a clean main branch build just in case.

msimberg:
> I will wipe my build tree and do a clean main branch build just in case.

Sounds good. Let us know how it goes in any case. It could be that you're using a configuration that we don't cover in CI.

#106 should fix the CI problems, and will hopefully be merged soon. Once that's in the tests can be retriggered here but this looks good to go already now.

msimberg:
test cscs

PR #104, opened 2022-02-09
Add direct dependency on APEX

msimberg:
Fixes #40. Also see #103.

Still to do:
- [x] make sure CI has APEX installed and available
- [x] add patch to APEX spack package (see https://github.com/spack/spack/pull/28844, the patch is already applied on our local spack instance)

This also adds a way to use a spack environment in Jenkins, which is now used for the APEX build. Currently the only spack build is `gcc-10-apex-spack`. I've left the `-spack` suffix while this is the only configuration using spack. If we move all of them to use spack in the future we can remove the suffix.

I've disabled all other CI configurations that had APEX enabled (it was mostly formal, they didn't actually have real APEX support anyway). The only one with APEX enabled is now the new spack configuration on Jenkins.

## Proposed Changes

- Adds a direct dependency on APEX and uses its `task_wrapper` type directly. APEX does not depend on pika.
- Adds the `annotation_check` test from HPX.
- Does light refactoring of headers and functions in the `external_timer` namespace.
- Moves `external_timer` into `pika::detail`.

## Checklist

Not all points below apply to all pull requests.

- [x] I have added a new feature and have added tests to go along with it.
- [ ] I have fixed a bug and have added a regression test.
- [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.

PR #106, opened 2022-02-10
Use even more specific spec for cmake and ninja in Jenkins CI

msimberg:
Hopefully fixes Jenkins CI. `ccache` will also be updated separately (#104), but it does not seem to be required right now.


msimberg:
test cscs

msimberg:
test cscs

PR #109, opened 2022-02-10
Add `when_all_vector` sender adaptor

### LINKED ISSUE #86: Add variant of `when_all` that accepts containers of senders and sends a container of values
`dataflow` recursively waits for any containers of futures (of containers). One can do e.g. this with `dataflow`: `dataflow(unwrapping([](vector<T>){...}), vector<future<T>>{...})`. There is no equivalent sender adaptor. `when_all` is variadic and requires all arguments to be senders themselves.

This feature is used in DLA-Future. A `when_all_vector(vector<sender_of<T>>)` would be sufficient as a starting point.

msimberg:
This is an adaptor like when_all, but specialized to take a vector of senders and signal the continuation when all senders in the vector have completed. It requires that the senders send a single type, or none at all. If they send nothing, the sender returned by when_all_vector also sends nothing.

Fixes #86, at least the minimum required case for DLA-Future.

Note that this currently only works with a vector of senders. Open questions for future work are:
1. Can this be optimized to allocate less? There are currently many separate heap allocations within this adaptor which could potentially be unified.
2. Can this be generalized in some reasonable way? It could handle any ranges or iterator/sentinel pairs. What container should it send in that case? It could handle nested senders, like dataflow.

Opening this PR for early CI/review feedback. This has _not_ yet been tested in DLA-Future.

## Checklist

- [x] I have added a new feature and have added tests to go along with it.
- [ ] I have fixed a bug and have added a regression test.

## To do before merging

- [x] Add `when_all(when_all_vector(x), when_all_vector(y), when_all_vector(z))` test case
- [ ] Rerun tests on Piz Daint (down for maintenance at the time of writing)

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/when_all_vector.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/when_all_vector.hpp
@@ -1,0 +1,342 @@
+//  Copyright (c) 2020 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#pragma once
+
+#include <pika/config.hpp>
+#include <pika/assert.hpp>
+#include <pika/concepts/concepts.hpp>
+#include <pika/datastructures/optional.hpp>
+#include <pika/datastructures/variant.hpp>
+#include <pika/execution/algorithms/detail/single_result.hpp>
+#include <pika/execution_base/operation_state.hpp>
+#include <pika/execution_base/receiver.hpp>
+#include <pika/execution_base/sender.hpp>
+#include <pika/functional/detail/tag_fallback_invoke.hpp>
+#include <pika/type_support/detail/with_result_of.hpp>
+#include <pika/type_support/pack.hpp>
+
+#include <atomic>
+#include <cstddef>
+#include <exception>
+#include <functional>
+#include <memory>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace pika::execution::experimental {
+    namespace when_all_vector_detail {
+        template <typename Sender>
+        struct when_all_vector_sender_impl
+        {
+            struct when_all_vector_sender_type;
+        };
+
+        template <typename Sender>
+        using when_all_vector_sender = typename when_all_vector_sender_impl<
+            Sender>::when_all_vector_sender_type;
+
+        template <typename Sender>
+        struct when_all_vector_sender_impl<Sender>::when_all_vector_sender_type
+        {
+            using senders_type = std::vector<Sender>;
+            senders_type senders;
+
+            explicit constexpr when_all_vector_sender_type(
+                senders_type&& senders)
+              : senders(PIKA_MOVE(senders))
+            {
+            }
+
+            explicit constexpr when_all_vector_sender_type(
+                senders_type const& senders)
+              : senders(senders)
+            {
+            }
+
+            // We expect a single value type or nothing from the predecessor
+            // sender type
+            using element_value_type = detail::single_result_t<
+                typename pika::execution::experimental::sender_traits<Sender>::
+                    template value_types<pika::util::pack, pika::util::pack>>;
+
+            static constexpr bool is_void_value_type =
+                std::is_void_v<element_value_type>;
+
+            // This is a helper empty type for the case that nothing is sent
+            // from the predecessors
+            struct void_value_type
+            {
+            };
+
+            // This sender sends a single vector of the type sent by the
+            // predecessor senders or nothing if the predecessor senders send
+            // nothing
+            template <template <typename...> class Tuple,
+                template <typename...> class Variant>
+            using value_types = Variant<std::conditional_t<is_void_value_type,
+                Tuple<>, Tuple<std::vector<element_value_type>>>>;
+
+            // This sender sends any error types sent by the predecessor senders
+            // or std::exception_ptr
+            template <template <typename...> class Variant>
+            using error_types = pika::util::detail::unique_concat_t<
+                typename pika::execution::experimental::sender_traits<
+                    Sender>::template error_types<Variant>,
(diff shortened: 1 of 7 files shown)


msimberg:
test cscs

msimberg:
test cscs

libs/pika/execution/tests/unit/algorithm_when_all_vector.cpp
```diff
@@ -0,0 +1,218 @@
...
+        senders.emplace_back(ex::just(42.0));
+        senders.emplace_back(ex::just(43.0));
+        senders.emplace_back(ex::just(44.0));
+        auto s = ex::when_all_vector(std::move(senders));
+        auto r = error_callback_receiver<decltype(check_exception_ptr)>{
+            check_exception_ptr, set_error_called};
+        auto os = ex::connect(std::move(s), std::move(r));
+        ex::start(os);
+        PIKA_TEST(set_error_called);
+    }
```
aurianer, line 212: Should we also add a test with a sender that is not just (like with a longer task) or does it not make sense?

aurianer (approved):
LGTM thanks a lot!

libs/pika/execution/tests/unit/algorithm_when_all_vector.cpp
msimberg: I added some tests that actually spawn tasks to the `thread_pool_scheduler` test over here: https://github.com/pika-org/pika/blob/d439f181006f3358500390740680ea053b5dfa6c/libs/pika/executors/tests/unit/thread_pool_scheduler.cpp#L661-L774. Do you think that covers it?

libs/pika/execution/include/pika/execution/algorithms/when_all_vector.hpp
```diff
@@ -0,0 +1,323 @@
...
+            using senders_type = std::vector<Sender>;
+            senders_type senders;
+
+            explicit constexpr when_all_vector_sender_type(
+                senders_type&& senders)
+              : senders(PIKA_MOVE(senders))
+            {
+                PIKA_ASSERT_MSG(this->senders.size() >= 1,
+                    "when_all_vector expects at least one element in the "
+                    "vector of senders");
```
msimberg, line 55: To do: this leaves it up to the caller to check that the vector isn't empty. Is that necessary?

aurianer: I'm fine with leaving it out

teonnik:
Hey, thanks for this! I have a couple of questions.

Can one pass more than one vector<> of senders?

```
ex::when_all_vector(senders1, senders2, senders3)
```

What type should be used when declaring a vector of senders?

I noticed that `vector<ex::any_sender<>>` and `vector<ex::unique_any_sender<>>` were used in the code.

msimberg:
>     1. Can one pass more than one vector<> of senders?

Currently no, but would be possible to add. Is that something that would be useful for your work? If yes, would it be essential or "nice to have"?

>     2. What type should be used when declaring a vector of senders?
> 
> 
> I noticed that `vector<ex::any_sender<>>` and `vector<ex::unique_any_sender<>>` were used in the code. For DLAF, since `TIle<>` is a move-only type, I presume `ex::unique_any_sender<>` is the right wrapper.

Yes, if you want to wrap different move-only types `unique_any_sender` is the right thing to use (it should maybe be renamed `move_only_any_sender` like `move_only_function`...). For copyable types `any_sender`. And of course if all the senders are of the same type you should just use that type directly. Note that `Tile` being move-only doesn't necessarily mean that the sender can't be copyable, like e.g. `shared_future<Tile<...>>` or the sender returned from `split` (the sender-equivalent of `shared_future`).

teonnik:
> Edit: Do you expect the senders in different vectors to send different types? If yes, that would be when_all(when_all_vector(v1), when_all_vector(v2), when_all_vector(v3)) because the sent values necessarily need to be stored in different vectors anyway. 

Yes. I have a kernel passed to `pika::dataflow()` in DLAF that takes three vectors in total: two `std::vector<pika::future<Tile<T, Device::CPU>>>` where `T` is `double` or `float` and a `std::vector<pika::future<Tile<std::ptrdiff_t, Device::CPU>>>`.

If I can do something like this:

```
void mykernel(const std::vector<Tile<T, Device::CPU>>&,const std::vector<Tile<T, Device::CPU>>&, const std::vector<Tile<SizeType, Device::CPU>>&) {...}

when_all(when_all_vector(v1), when_all_vector(v2), when_all_vector(v3)) | mykernel(Policy<>(...)) | start_detached()
```

then I don't need any further extension to the currently implemented functionality.

msimberg:
> > Edit: Do you expect the senders in different vectors to send different types? If yes, that would be when_all(when_all_vector(v1), when_all_vector(v2), when_all_vector(v3)) because the sent values necessarily need to be stored in different vectors anyway.
> 
> Yes. I have a kernel passed to `pika::dataflow()` in DLAF that takes three vectors in total: two `std::vector<pika::future<Tile<T, Device::CPU>>>` where `T` is `double` or `float` and a `std::vector<pika::future<Tile<std::ptrdiff_t, Device::CPU>>>`.
> 
> If I can do something like this:
> 
> ```
> void mykernel(
>     const std::vector<Tile<T, Device::CPU>>& v1,
>     const std::vector<Tile<T, Device::CPU>>& v2, 
>     const std::vector<Tile<SizeType, Device::CPU>> & v3) {
> ...
> }
> 
> when_all(when_all_vector(v1), when_all_vector(v2), when_all_vector(v3)) | 
>     mykernel(Policy<>(...)) | 
>     start_detached()
> ```
> 
> then I don't need any further extension to the currently implemented functionality.

It *should* work, but I'm at the moment not explicitly testing for that case. It's a good test case to add.

msimberg:
test cscs

msimberg:
@teonnik FYI I think this test covers your use case: https://github.com/pika-org/pika/blob/cecbd3f480eed751c41f0f58d685e2e6151d2cff/libs/pika/execution/tests/unit/algorithm_when_all_vector.cpp#L165-L202.

msimberg:
bors merge

PR #111, opened 2022-02-11
Fix `split`/`ensure_started` shared state reference counting

msimberg:
Fixes the reference counting for the shared state of `split`/`ensure_started`. It was being held with an additional unnecessary reference count by the `split_receiver` which is connected to the predecessor sender. This made the shared state stay alive longer than necessary.

Additionally it resets the operation state (resulting from connecting the predecessor sender to `split`/`ensure_started` to `split_receiver`) as soon as `set_x` has been called on `split_receiver`. This releases resources as early as possible, since the operation state is no longer needed after `set_x` has been called.

This was triggered by the work in https://github.com/eth-cscs/DLA-Future/pull/475 where the RAII nature of `Tile` revealed deadlocks unless the operation state is released as early as possible.

Note that currently P2300 does not mandate that the operation state has to be released this early. I don't yet know if it should mandate that or not, but if it won't this kind of usage requires a non-standard `ensure_started` which has this guarantee.

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/split.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/split.hpp
@@ -156,23 +156,21 @@ namespace pika { namespace execution { namespace experimental {
 
                 struct split_receiver
                 {
-                    pika::intrusive_ptr<shared_state> state;
+                    shared_state& state;
 
                     template <typename Error>
                     friend void tag_invoke(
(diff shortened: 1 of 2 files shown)


msimberg:
test cscs

aurianer (approved):
LGTM thanks a lot!

msimberg:
bors merge

PR #110, opened 2022-02-10
Reschedule threads on the worker thread where they were suspended again

msimberg:
This used to work but the changes seem to have been lost at some point. This also adds a test to verify the behaviour with the static scheduler (static since we want to guarantee that a thread runs where it's scheduled).

## Any background context you want to provide?

This is not super critical but I noticed this was missing when working on DLA-Future (https://github.com/eth-cscs/DLA-Future/pull/475).

## Checklist

Not all points below apply to all pull requests.

- [x] I have added a new feature and have added tests to go along with it.
- [ ] I have fixed a bug and have added a regression test.
- [ ] I have added a test using random numbers; I have made sure it uses a seed, and that random numbers generated are valid inputs for the tests.

PR #112, opened 2022-02-11
Don't allow yielding with uncaught exceptions

msimberg:
This is also related to the work in https://github.com/eth-cscs/DLA-Future/pull/475. The underlying bug is fixed separately, but this fixes a side effect that was revealed by the bug, which I don't see a reason to allow, namely yielding pika threads with uncaught exceptions.

This was triggered by a test:
- throwing an exception
- the destructor for an unsatisfied promise running
- the destructor trying to construct an exception
- the exception construction trying to collect a stacktrace
- the stacktrace collection being launched on another thread
- the exception construction waiting for the completion of the stacktrace by suspending/yielding
- the exception construction thread waking up on another worker thread (this happens sometimes with the stealing schedulers, also see #110)

Since the yield happens with an uncaught exception the value of `std::uncaught_exceptions()` is one on the worker thread where the yield happened. When the thread is resumed and the exception is eventually handled libstdc++ will decrement the count for `std::uncaught_exceptions()`. However, since it's now on another thread and the count is thread local the count will be decremented to `-1` on the new worker thread, while the count remains `1` on the first worker thread. This leads to all kinds of inconsistencies.

This adds assertions that `std::uncaught_exceptions()` is `0` before yielding since I don't see a legitimate use case (at this point) for allowing it.

Additionally, I've made the stacktrace collection happen inline instead of on a new pika thread to make it safe to construct pika exceptions while there are uncaught exceptions. This last part I'm slightly unsure about since I'm guessing there was a reason for putting the stacktrace collection on a separate pika thread. However, so far it seems to be safe. CI may tell me otherwise. @biddisco do you have any memory of why the stacktrace collection might have happened on a separate pika thread?

### DIFF
--- a/libs/pika/coroutines/include/pika/coroutines/detail/context_base.hpp
+++ b/libs/pika/coroutines/include/pika/coroutines/detail/context_base.hpp
@@ -193,6 +193,15 @@ namespace pika { namespace threads { namespace coroutines { namespace detail {
         //         resumed.
         void yield()
         {
+            // Misbehaved threads may try to yield while handling an exception.
+            // This is dangerous if the thread can migrate to other worker
+            // threads since the count for std::uncaught_exceptions may become
+            // inconsistent (including negative). If at any point in the future
+            // there is a legitimate use case for yielding with uncaught
+            // exceptions this assertion can be revisited, but until then we
+            // prefer to be strict about it.
+            PIKA_ASSERT(std::uncaught_exceptions() == 0);
+
             //prevent infinite loops
             PIKA_ASSERT(m_exit_state < ctx_exit_signaled);
             PIKA_ASSERT(running());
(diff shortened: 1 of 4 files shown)


msimberg:
test cscs

biddisco:
> @biddisco do you have any memory of why the stacktrace collection might have happened on a separate pika thread?

No. That seems a bit ridiculous. Stacktrace collection should be done inline IMHO.

msimberg:
> > @biddisco do you have any memory of why the stacktrace collection might have happened on a separate pika thread?
> 
> No. That seems a bit ridiculous. Stacktrace collection should be done inline IMHO.

That's what I thought as well. Let's go with this change then. I think we'll find out soon enough what we've missed if there is something that we've missed.

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

Issue #23, opened 2022-01-18
Fix or disable remaining failing tests using timed suspension


msimberg:
The following (and possibly a few more) tests fail for various reasons after disabling timed suspensions. They need to be dealt with before the first release.

- tests.unit.modules.synchronization.shared_mutex.shared_mutex1
- tests.unit.modules.threading.condition_variable2
- tests.unit.modules.threading.stop_token_cb1
- tests.unit.modules.threading.thread
- tests.performance.local.tls_overhead

msimberg:
Things got better with #45, but there are still problems so I'm reopening this.

msimberg:
This is close to done, but may still need some work. Reopening.

PR #113, opened 2022-02-14
Attempt to fix deadlocks in `condition_variable`

msimberg:
Attemps to fix the deadlocks that seem to happen occasionally in the `condition_variable2/4` tests. The deadlocks seem related to `stop_token` in the sense that previously if a stop request comes in at a bad moment compared to when a `stop_callback` is created, the `stop_callback` will execute but will not mark itself as executed. While removing the callback it will then wait (in vain) for the callback to first have been executed (which it has, but it has not been marked as executed).

The first three commits of this PR slightly refactor the deadlock detection mechanism we have to first print a warning at a lower iteration count. This also reduces the limit at which an exception is thrown when deadlock detection is enabled. I hope I've not reduced the limit too far but it's still user-configurable, is only enabled by default in debug builds, and we can still increase the default again. The current limit ends up stopping after around 10 seconds (with nothing else in the queues).

The last commit attempts to fix the deadlock itself by marking the callback as executed after executing it.

Part of #23.

### DIFF
--- a/libs/pika/config/include/pika/config.hpp
+++ b/libs/pika/config/include/pika/config.hpp
@@ -61,7 +61,12 @@
 ///////////////////////////////////////////////////////////////////////////////
 /// By default, enable minimal thread deadlock detection in debug builds only.
 #if !defined(PIKA_SPINLOCK_DEADLOCK_DETECTION_LIMIT)
-#  define PIKA_SPINLOCK_DEADLOCK_DETECTION_LIMIT 1073741823
+#  define PIKA_SPINLOCK_DEADLOCK_DETECTION_LIMIT 10000000
+#endif
+
+/// Print a warning about potential deadlocks after this many iterations.
+#if !defined(PIKA_SPINLOCK_DEADLOCK_WARNING_LIMIT)
+#  define PIKA_SPINLOCK_DEADLOCK_WARNING_LIMIT 1000000
 #endif
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/libs/pika/execution_base/include/pika/execution_base/detail/spinlock_deadlock_detection.hpp
+++ b/libs/pika/execution_base/include/pika/execution_base/detail/spinlock_deadlock_detection.hpp
@@ -15,6 +15,8 @@ namespace pika { namespace util { namespace detail {
     PIKA_EXPORT void set_spinlock_break_on_deadlock_enabled(bool enabled);
     PIKA_EXPORT bool get_spinlock_break_on_deadlock_enabled();
     PIKA_EXPORT void set_spinlock_deadlock_detection_limit(std::size_t limit);
+    PIKA_EXPORT void set_spinlock_deadlock_warning_limit(std::size_t limit);
     PIKA_EXPORT std::size_t get_spinlock_deadlock_detection_limit();
+    PIKA_EXPORT std::size_t get_spinlock_deadlock_warning_limit();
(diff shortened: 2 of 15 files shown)


msimberg:
test cscs

msimberg:
This has run once without the actual fix, and this is an example of the output this gives now on potential deadlocks: https://cdash.cscs.ch/test/2709735. Pushing the actual fix now.

msimberg:
test cscs

biddisco (approved):
I have looked through this and it seems fine, but I don't really understand the solution or the problem it solves, so my review is only a check that there doesn't appear to be anything bad in here.

msimberg:
Sigh: https://cdash.cscs.ch/test/2724281. I still think this is an improvement, but it's clearly not a full fix... I'll still have a short look at that new failure.

msimberg:
I've added one more change that seems to fix `stop_token` hopefully for good.

The locking was faulty in that it could occasionally let two threads enter the locked region concurrently. See https://github.com/pika-org/pika/pull/113/commits/6c9e5699b83bc5631ce0ce9ef143ed4f953ba0a4 for details.

msimberg:
test cscs

msimberg:
test cscs

msimberg:
There's definitely more to this so I'm still not going to close #23 but I think this is an improvement so I'm going to merge it. To be continued...

bors merge

aurianer (approved):
LGTM thanks!

hkaiser:
FWIW, this was fixed in HPX recently as well (see https://github.com/STEllAR-GROUP/hpx/pull/5749), in particular some significant changes to stop_token:  
- https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R70-R83, 
- https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R94-R108
- https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R120-R121

(the last change was applied here as well).

msimberg:
> FWIW, this was fixed in HPX recently as well (see [STEllAR-GROUP/hpx#5749](https://github.com/STEllAR-GROUP/hpx/pull/5749)), in particular some significant changes to stop_token:
> 
>     * https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R70-R83,
> 
>     * https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R94-R108
> 
>     * https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R120-R121
> 
>     * https://github.com/STEllAR-GROUP/hpx/pull/5749/files#diff-f8adc309a1446f848370feb4b83c32389e3213005833e2c2dc4a418a32b6cb58R129-R151
> 
> 
> (the last two changes were partially applied here as well).

Thanks @hkaiser!

Issue #63, opened 2022-01-21
Enable sanitizers in CI


msimberg:
This is potentially very important for debugging. We currently only test with lsan (leak sanitizer). We should try to add tsan (thread sanitizer), msan (memory sanitizer), asan (address sanitizer), and ubsan (undefined behaviour sanitizers). Enabling them with heavy suppression files would be a good start, just to allow consumers of pika to enable sanitizers.

- [x] lsan
- [x] tsan
- [x] asan
- [x] ubsan

# msan

This would require recompiling all dependencies with `-fsanitize=memory`, including the standard library: https://github.com/google/sanitizers/wiki/MemorySanitizer#using-instrumented-libraries. This is something to consider doing with the CSCS CI pipelines and spack.

msimberg:
With #1042 all tests and examples, except for an explicit set of blacklisted tests, run through asan, lsan, ubsan, and tsan on merges. msan is still not used at all.

msimberg:
Closing in favour of a more specific issue #1124.

PR #115, opened 2022-02-16
More sanitizers

msimberg:
This is meant for #63. I have done _zero_ work to attempt to fix any leaks etc. This currently only adds various new CI sanitizer configurations (MemorySanitizer and ThreadSanitizer) as well as one which runs tests inside valgrind. I'm simply opening this as a draft to see how bad the situation is.

We might be able to get these running by heavily suppressing the current errors. That would at least be useful to ensure we don't introduce new errors and would let us gradually fix what is there.


msimberg:
Given that this PR has not made any progress in its current form, I'm closing this. I've updated #63 with a list of all modules by sanitizer. I suggest we try to take the same approach as with #802 instead.

Issue #22, opened 2022-01-18
Remove future functionality


msimberg:
Including `hpx::async/apply/dataflow/when_all`. This can be done once DLA-Future has been completely ported to use senders. The cleanup itself for this issue is not complicated and only requires removing functionality. What functionality is currently missing on the sender/receiver side?
- #13
- #14
- Anything else?

PR #126, opened 2022-02-18
Remove leftover tests and a few `is_action` uses

msimberg:
Removes a few distributed test files that were still hanging around. Also removes the use of `is_action` in many places. Replaces it with an always-false trait. I've left the use of `is_action` in the executor/dataflow functionality only because of laziness since the full files will be removed for #22 which will take any last uses of `is_action` with them.

Related to #8.


msimberg:
test cscs

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

PR #128, opened 2022-02-22
Thread-local streams in `cuda_pool`

msimberg:
After the initial attempt of having a global pool of streams in `cuda_pool` I'm reverting it back to what DLA-Future used originally, i.e. thread-local streams. To avoid having to create the `cuda_pool` after the pika runtime I simply create a vector with `hardware_concurrency * num_streams_per_thread`. pika worker threads will have their thread number set and index into the vector with the thread number. non-pika threads will have their thread number set to `-1` and will all use the same "thread-local" streams, but at least they will fall back to a valid stream. I create one vector of streams instead of using `thread_local` streams to avoid calling CUDA API functions at arbitrary points when the `thread_local`s are accessed the first time.

I've also moved the `thread_local` cuBLAS/cuSOLVER handles to a cpp file to make sure there is really only one handle per thread, instead of one per function instantiation (the `thread_local` handles were defined in a function template). These are true `thread_local`s unlike the streams because of thread-safety issues.

Because of the move from a global pool of streams to thread-local streams I moved the access of the "next stream" to just before CUDA functions are called in the receiver, instead of having it already in the sender constructor. This is because otherwise a full DLA-Future algorithm may end up using streams for only one worker thread because the senders are created on the same worker thread.

Finally, I've done two small cleanups:
- Add more `PIKA_NO_UNIQUE_ADDRESS`.
- Rename the internal `then_with_cuda` to `then_with_cuda_stream` because I think it describes it better.

### DIFF
--- a/libs/pika/async_cuda/CMakeLists.txt
+++ b/libs/pika/async_cuda/CMakeLists.txt
@@ -52,6 +52,7 @@ set(async_cuda_sources
     cusolver_exception.cpp
     cusolver_handle.cpp
     get_targets.cpp
+    then_with_stream.cpp
 )
 
 if(PIKA_WITH_HIP AND TARGET roc::hipblas)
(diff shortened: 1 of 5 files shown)


msimberg:
test cscs

msimberg:
test cscs

msimberg:
@aurianer I'm going to go ahead and merge this with the knowledge that this works in DLA-Future, but a review is _very_ welcome even after merging.

msimberg:
bors merge

PR #140, opened 2022-03-10
Don't unwrap reference_wrapper in bind_front

### LINKED ISSUE #130: `bind_front` should not unwrap `reference_wrapper`
We can either update it to be conformant, or wait until we require C++20 and remove our own implementation. I'm unsure if we actually rely on this behaviour somewhere within pika.

msimberg:
Fixes #130.

Flyby: replaced `::type` and `::value` uses with `_t` and `_v` in the `bind_front.hpp` file.


msimberg:
test cscs

PR #142, opened 2022-03-15
Check that MPI polling has been registered also when using senders

msimberg:
Minor debugging help. It was previously only checked when using futures, not when using senders. Helps with https://github.com/eth-cscs/DLA-Future/pull/497.


msimberg:
test cscs

aurianer (approved):

PR #143, opened 2022-03-15
Separate `split` and `ensure_started`

### LINKED ISSUE #93: Un-unify `split` and `ensure_started`
Currently `split` and `ensure_started` share the implementation with the only difference being when `start` is called. However, this places the unnecessary restriction on `ensure_started` that it sends `const&`s instead of moving the values to the receiver.

msimberg:
Fixes #93. The two share no code now. As an optimization `split` could use the shared state of `ensure_started`, but I'm leaving that for maybe-in-the-future.


msimberg:
test cscs

msimberg:
test cscs

msimberg:
bors merge

PR #144, opened 2022-03-17
Fix SFINAE in `then_with_stream`

### LINKED ISSUE #136: Investigate if `then_with_stream` and friends require const references instead of plain references in callables
It should allow both, but it seems like it doesn't (see https://github.com/eth-cscs/DLA-Future/pull/460/commits/3d7f7ca0b926b6838b75670dbdc0f9386470b7e0).

msimberg:
The checks for invocability were not correctly testing with lvalue references. This meant that functions which take references to noncopyable types would incorrectly be diagnosed as not invocable. This also adds a regression test. Fixes #136.

## Checklist

- [ ] I have added a new feature and have added tests to go along with it.
- [X] I have fixed a bug and have added a regression test.

### DIFF
--- a/libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
+++ b/libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
@@ -209,8 +209,8 @@ namespace pika::cuda::experimental::detail {
         }
 
         template <typename... Ts>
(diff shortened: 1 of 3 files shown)


msimberg:
test cscs

msimberg:
test cscs

msimberg:
bors merge

PR #145, opened 2022-03-17
Remove serialization support

### LINKED ISSUE #7: Remove serialization support
Serialization is not used by DLA-Future.

aurianer:
Fixes #7. Removes the serialization module and helpers

@msimberg I remember that you told me to keep some of the tests that were testing some other facilities but I cannot remember which files we were talking about, could you refresh my memory?


libs/pika/functional/include/pika/functional/detail/function_registration.hpp
```diff
@@ -26,14 +26,7 @@ namespace pika { namespace util { namespace detail {
...
     {
-        static char const* call()
-#ifdef PIKA_HAVE_AUTOMATIC_SERIALIZATION_REGISTRATION
-        {
-            return debug::type_id<F>::typeid_.type_id();
-        }
-#else
-            = delete;
-#endif
+        static char const* call() = delete;
```
msimberg, line 29: This whole file could possibly be deleted. Have you checked where it's still used?

msimberg:
> @msimberg I remember that you told me to keep some of the tests that were testing some other facilities but I cannot remember which files we were talking about, could you refresh my memory?

I think I may have been referring to some tests testing both serialization and the more basic functionality (like `function`), and in those cases the non-serialization tests should still be kept. It may however be that those tests are all separated into separate tests and then my comment doesn't apply anymore.

libs/pika/functional/tests/unit/function_object_size.cpp
```diff
@@ -1,238 +0,0 @@
-////////////////////////////////////////////////////////////////////////////////
```
msimberg, line 1: This test for example tests non-serialization functionality as well and should stay (but the serialization functions can be removed of course).

aurianer: good it was probably the one we talked about, I will restore the non-serialization functionalities :ok_hand:

libs/pika/functional/include/pika/functional/detail/function_registration.hpp
aurianer: Indeed it was used in unique_function and function in preprocessor macros that are in fact unused, removing..

aurianer:
test cscs

msimberg (approved):
Looks good, thanks!

msimberg:
bors merge

PR #149, opened 2022-03-24
Enable examples as tests again in Jenkins CI

msimberg:
I have no idea why they were disabled, but I'm guessing I made a typo/thinko. They should be enabled.


aurianer (approved):
LGTM thanks!

msimberg:
test cscs

msimberg:
test cscs

msimberg:
test cscs

msimberg:
Note that after enabling the examples as tests again the `guided_pool` example started failing with assertions about thread/pool numbers in the `shared_priority_queue_scheduler`. The changes in #105 seem to work with clang, but not gcc, i.e. gcc was creating multiple instances of the thread-local structs with one being used while setting the numbers and another while getting them. Maybe due to how `inline constexpr` is handled across shared libraries/executables? In any case I've partially reverted the changes and moved the thread-local variables back into a cpp file and things are back to normal.

PR #125, opened 2022-02-18
Update jenkins configs to only use spack

msimberg:
Fixes #121.

This is also WIP. I'm currently building some of the specs and this needs jenkins to be back in action before it's useful, but I'm starting the work. Comments on structure welcome.

To do:
- [x] change perftests to use spack environment

PR #151, opened 2022-03-25
Add basic annotation test for sender adaptors

### LINKED ISSUE #103: Check annotations for senders and sender adaptors
There is a test to check annotations with `async`, `dataflow`, `then`, etc., i.e. for the future-based adaptors. There is currently no test for senders and sender adaptors. We should check that things are set up reasonably. Annotations should already be enabled for tasks created with `thread_pool_scheduler` (https://github.com/pika-org/pika/blob/0ab7223c4e08a9888bca38260ccb53db5bd0c2c0/libs/pika/executors/include/pika/executors/thread_pool_scheduler.hpp#L144), but this needs to be tested.

msimberg:
Fixes #103. From what I can tell everything is in place.

### DIFF
--- a/libs/pika/threading_base/tests/unit/CMakeLists.txt
+++ b/libs/pika/threading_base/tests/unit/CMakeLists.txt
@@ -28,6 +28,8 @@ foreach(test ${tests})
 endforeach()
 
 if(PIKA_WITH_APEX)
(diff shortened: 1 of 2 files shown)


msimberg:
test cscs

msimberg:
Note that I've apparently broken the APEX build while working on #125, so this can wait a bit until that is ready and the annotations actually get tested.

msimberg:
test cscs

msimberg:
bors merge

PR #154, opened 2022-03-25
Replace typedefs with using

aurianer:
I thought of doing that in one go instead of piece by piece
EDIT: my scripts missed some, I have to figure out a way to catch those ^^'

### DIFF
--- a/examples/1d_stencil/1d_stencil_1.cpp
+++ b/examples/1d_stencil/1d_stencil_1.cpp
@@ -33,10 +33,10 @@ double dx = 1.;        // grid spacing
 struct stepper
 {
     // Our partition type
-    typedef double partition;
+    using partition = double;
 
     // Our data for one time step
-    typedef std::vector<partition> space;
+    using space = std::vector<partition>;
 
     // Our operator
     static double heat(double left, double middle, double right)
--- a/examples/1d_stencil/1d_stencil_1_omp.cpp
+++ b/examples/1d_stencil/1d_stencil_1_omp.cpp
@@ -48,10 +48,10 @@ double dx = 1.;        // grid spacing
 struct stepper
 {
     // Our partition type
-    typedef double partition;
+    using partition = double;
 
     // Our data for one time step
-    typedef std::vector<partition> space;
+    using space = std::vector<partition>;
 
     // Our operator
     static double heat(double left, double middle, double right)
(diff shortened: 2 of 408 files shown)


examples/quickstart/sort_by_key_demo.cpp
```diff
@@ -51,6 +51,5 @@ int pika_main()
 ///////////////////////////////////////////////////////////////////////////////
 int main(int argc, char* argv[])
 {
-    return pika::init(
-        pika_main, argc, argv);    // Initialize and run pika
+    return pika::init(pika_main, argc, argv);    // Initialize and run pika
```
msimberg, line 54: Unrelated to this PR, but we can probably have the clang-format check enabled for all source files now (including .cu files), right?

msimberg (commented):
Excellent idea. Could you try enabling https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-using.html in our clang-tidy config after this as well?

examples/quickstart/sort_by_key_demo.cpp
msimberg: And related to clang-format, I wouldn't mind bumping the clang version if needed.

aurianer:
Is that okay if I look into the rest of the typedefs in another PR? Cause it conflicts quite much with the cleanup PRs
Otherwise any idea about the codespell issue? Is it possible to ignore that file or that url in particular?

aurianer:
> Excellent idea. Could you try enabling https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-using.html in our clang-tidy config after this as well?

I opened #169 to remember that

examples/quickstart/sort_by_key_demo.cpp
aurianer: I opened #168

msimberg:
> Is that okay if I look into the rest of the typedefs in another PR? Cause it conflicts quite much with the cleanup PRs

Absolutely.

> Otherwise any idea about the codespell issue? Is it possible to ignore that file or that url in particular?

I wouldn't want to ignore the whole file, and the check is not required. As far as I know it's only possible to have either a global whitelist of spellings, or ignore files, but not possible to have a whitelist for a specific file/line. I think we can just ignore it.

msimberg (approved):

msimberg:
bors merge

PR #157, opened 2022-03-28
Fix compilation with precompiled headers

msimberg:
The `filesystem` module does not have a `pika/modules/filesystem.hpp` header (and I don't think it needs one), so I've removed it from the precompiled header.

### DIFF
--- a/cmake/pika_setup_vc.cmake
+++ b/cmake/pika_setup_vc.cmake
@@ -87,4 +87,8 @@ if(PIKA_WITH_DATAPAR_VC AND NOT TARGET Vc::vc)
   pika_info("Found Vc (vectorization):" ${Vc_INCLUDE_DIR} "- version:"
             ${Vc_VERSION_STRING}
   )
+
+  if(NOT PIKA_FIND_PACKAGE)
+    target_link_libraries(pika_base_libraries INTERFACE Vc::vc)
+  endif()
 endif()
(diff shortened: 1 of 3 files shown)


msimberg:
test cscs

msimberg:
To do before merging: the `pika/modules/assertion.hpp` header has also been removed... Enable precompiled headers on some CI configuration? How does it interact with header tests?

msimberg:
Precompiled headers do _not_ get used for the header tests, so this is ready to go.

msimberg:
test cscs

cmake/pika_setup_vc.cmake
```diff
@@ -87,4 +87,8 @@ if(PIKA_WITH_DATAPAR_VC AND NOT TARGET Vc::vc)
   pika_info("Found Vc (vectorization):" ${Vc_INCLUDE_DIR} "- version:"
             ${Vc_VERSION_STRING}
   )
+
+  if(NOT PIKA_FIND_PACKAGE)
+    target_link_libraries(pika_base_libraries INTERFACE Vc::vc)
```
aurianer, line 92: How was it working before? ^^'

aurianer (approved):
LGTM thanks!

cmake/pika_setup_vc.cmake
msimberg: This was (and still is) linked directly to the module that uses Vc as well. This makes it easier (and possible) to use the same precompiled headers for all modules (Vc sets architecture-specific flags for consumers).

msimberg:
bors merge

PR #156, opened 2022-03-28
Update `transform_mpi`

msimberg:
Updates `transform_mpi` according to #138. Also applies the fix done in https://github.com/pika-org/pika/pull/144 to `then_with_stream` to `transform_mpi`. Also applies some `S`-to-`sender`-style renamings.

~As a flyby I've enabled some of the MPI tests on CircleCI. This still needs work.~

To do:
- [x] simplify the `set_value` helpers to remove `pika::visit`.

PR #162, opened 2022-03-30
Store values sent from predecessor to `then_with_stream` in operation state

### LINKED ISSUE #138: `then_with_stream` should keep arguments alive in operation state
Currently `transform_stream` gives references to the values sent by the predecessor sender, and then moves the values into a callback to release the values once the CUDA event completes. This works in DLA-Future because `Tile`s have reference semantics so moving the values away does not invalidate the underlying data pointed to by the `Tile`s. However, in the more general case the callable may actually need to reference the actual data after returning, in which case moving the values to the event callback doesn't work. We should store the values sent by the predecessor sender in the operation state of `then_with_stream` instead to give them a stable location.

Edit: This also applies to `transform_mpi`.

msimberg:
Fixes #138 together with #156. The changes here are very similar to in #156, i.e. instead of storing the callable, the successor receiver, the scheduler, and the values sent from the predecessor in the `then_with_cuda_stream_receiver` they are now stored in the operation state. This simplifies the handling of the callback as we can just capture a reference to the operation state in the event callback.

### DIFF
--- a/libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
+++ b/libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
@@ -16,7 +16,9 @@
 #include <pika/async_cuda/custom_gpu_api.hpp>
 #include <pika/async_cuda/custom_lapack_api.hpp>
 #include <pika/async_cuda/detail/cuda_event_callback.hpp>
+#include <pika/datastructures/optional.hpp>
 #include <pika/datastructures/tuple.hpp>
+#include <pika/datastructures/variant.hpp>
 #include <pika/execution/algorithms/detail/partial_algorithm.hpp>
 #include <pika/execution/algorithms/then.hpp>
 #include <pika/execution_base/receiver.hpp>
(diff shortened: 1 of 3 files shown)


msimberg:
test cscs

msimberg:
test cscs

msimberg:
test cscs

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
```diff
@@ -85,268 +87,108 @@ namespace pika::cuda::experimental::detail {
...
+                    pika::execution::experimental::set_value(
+                        PIKA_MOVE(op_state.receiver));
+                }
+                else
+                {
+                    PIKA_UNREACHABLE;
+                }
             },
-            stream.get());
+            op_state.result);
```
msimberg, line 106: As pointed out by @aurianer this could be simplified. We can assert that the variant holds a `pika::monostate` and access that member directly.

msimberg:
test cscs

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
msimberg: I've made the changes here. Please have a look @aurianer. It should be significantly more straightforward to understand now.

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
```diff
@@ -85,268 +87,108 @@ namespace pika::cuda::experimental::detail {
...
+                }
+                else
+                {
+                    pika::execution::experimental::set_value(
+                        PIKA_MOVE(op_state.receiver), PIKA_MOVE(result));
+                }
             },
-            stream.get());
+            op_state.result);
     }
```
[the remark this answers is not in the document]
msimberg: This is also now simplified.

aurianer (approved):
LGTM thanks!

msimberg:
test cscs

msimberg:
test cscs

msimberg:
bors merge

PR #165, opened 2022-03-31
Remove `-w` flag to disable all warnings with CUDA

### LINKED ISSUE #161: Don't leak `-w` nvcc flag to consumers
We currently set `-w` (i.e. `--disable-warnings`) in the `Cuda::cuda` interface target, which ends up in consumers' targets. https://github.com/pika-org/pika/blob/009f6687a80100a6c54855e0388080abbd51def6/cmake/pika_setup_cuda.cmake#L50. We should not do this.

msimberg:
(Eventually) fixes #161. Not tested, will run CI to see what this does initially.

### DIFF
--- a/cmake/pika_setup_cuda.cmake
+++ b/cmake/pika_setup_cuda.cmake
@@ -45,12 +45,7 @@ if(PIKA_WITH_CUDA AND NOT TARGET Cuda::cuda)
   )
 
   if(NOT PIKA_WITH_CLANG_CUDA)
-    if(NOT MSVC)
-      target_compile_options(
-        Cuda::cuda INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:-w>
-      )
-    else()
-      # Windows
+    if(MSVC)
       set(CUDA_PROPAGATE_HOST_FLAGS OFF)
       target_compile_options(
         Cuda::cuda
--- a/cmake/pika_setup_cuda.cmake
+++ b/cmake/pika_setup_cuda.cmake
@@ -44,7 +44,11 @@ if(PIKA_WITH_CUDA AND NOT TARGET Cuda::cuda)
     Cuda::cuda PROPERTIES INTERFACE_POSITION_INDEPENDENT_CODE ON
   )
 
-  if(NOT PIKA_WITH_CLANG_CUDA)
+  if(PIKA_WITH_CLANG_CUDA)
+    if(NOT PIKA_FIND_PACKAGE)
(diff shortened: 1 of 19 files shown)


msimberg:
test cscs

msimberg:
There are quite a few warnings. I've attempted to fix some of them without testing in [af45a77](https://github.com/pika-org/pika/pull/165/commits/af45a77e7d51355677a21acfeda4f7eb5db90320) but there are still unfixed ones. One problem is that even if we were to disable warnings only within pika, the warnings would most likely show up for consumers because the warnings come in most cases from header files...

msimberg:
test cscs

cmake/pika_setup_cuda.cmake
```diff
@@ -92,9 +91,18 @@ if(PIKA_WITH_CUDA AND NOT TARGET Cuda::cuda)
...
+      if(PIKA_WITH_COMPILER_WARNINGS_AS_ERRORS)
+        target_compile_options(
+          pika_private_flags INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:--Werror
+                                       all-warnings>
+        )
+      endif()
+
+      target_compile_options(
+        pika_private_flags
+        INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:--display-error-number>
```
msimberg, line 104: This is a flyby that I've added. It means that an error code will be printed together with warnings/errors and in the future we can use those codes to suppress warnings locally.

msimberg:
test cscs

msimberg:
I learned a few things while doing this:
- There's a `#pragma hd_warning_disable` that can be used to disable warnings about "calling host-only function from host/device function". This comes up in quite a lot of places and is more or less a false positive. Any templated function that could run on the device is marked host/device, but if it takes a host-only function nvcc will issue a warning. However, we can in most cases ignore these. I've disabled the warning in all places that triggered a warning plus nearby host/device functions that I imagined could trigger a warning in the future.
- nvcc starting from 11.5 has `#pragma nv_diag_suppress <error_code>` and `#pragma nv_diagnostic push/pop` that can be used to locally suppress warnings. I haven't used it conditionally here since most of our supported versions still don't have those pragmas.
- Almost any function with `if constexpr` and `auto` return type triggers a bogus warning about no return from a non-void function. I've suppressed these with `__builtin_unreachable` for nvcc only.

aurianer (approved):
LGTM thanks!

Issue #172, opened 2022-04-05
Allow for more flexibility in inheriting annotations


msimberg:
Currently some annotations are inherited at the point of spawning a task. In some cases the annotations are not inherited at all (`execution::schedule/transfer`). I'm only considering sender adaptors for this and will not focus on backporting any changes decided on here to `async`/`dataflow`/`then`.

I think there are a few possibilities for determining the annotation for a task:
1. Explicitly specified annotation
2. Inheriting the annotation from the task that spawns the new task
3. Inheriting the annotation from the task that schedules the new task

By spawning I mean eagerly creating a pika thread and by scheduling I mean creating a sender that eventually creates a pika thread.

This is an example that captures the different possibilities that I can think of right now:
```
scoped_annotation ann("A"); // 1. The scheduling task has an annotation
schedule(with_annotation(thread_pool_scheduler{}, "B")) | // 2. The task is explicitly annotated "B"
  then(f1) | // 3. f1 is not annotated, do nothing or inherit "B" or "A"?
  transfer(thread_pool_scheduler{}) | // 4. Inherit "B" or "A"?
  then(annotated_function(f2, "C")) | // 5. Use "C"
  then(f3) | // 6. Inherit "C", "B", or "A"?
  start_detached();
```

In my opinion there is no discussion for 1., 2., or 5. but welcome other opinions.

In 3. I think there is no inheritance needed. The task calling f1 is already annotated with "B".

In 5. one possibility would be to change the annotation of the task to "C" so that continuations are also annotated "C". However, I think this goes against locally annotating f2, so I think "C" is not an option in 6.

That leaves 4. and whatever is the correct answer to 4. is also the answer for 6.

I can see a case for both "A" and "B" in 4. On one hand experience from DLA-Future shows that "A" is often what one would expect. On the other hand "B" is the simpler alternative (and close to the current behaviour) since one only needs to look at the context at the point of actually spawning a task. However, "simpler to implement" should be the main criterion if option "A" is more intuitive.

There is one additional situation to consider:
```
// 1. The scheduling task has no annotation or is a plain std::thread
schedule(with_annotation(thread_pool_scheduler{}, "B")) |
  then(f1) |
  transfer(thread_pool_scheduler{}) | // 2. Fall back to "B" or nothing?
  then(2) |
  start_detached();
```
If we use "inherit from the scheduling task" the label in 2. should be nothing/`<unknown>`. I think this also fits if one considers that the situation could be:
```
[...]

PR #173, opened 2022-04-05
Inherit annotation from spawning task as fallback in `thread_pool_scheduler`

msimberg:
This is a last-minute addition to 0.3.0. Though the change may not immediately look like it, the `thread_description` constructor gets the fallback annotation from the spawning thread if needed. Added some tests, but more thorough tests will need to be added for #172.


msimberg:
test cscs

msimberg:
test cscs

msimberg:
bors merge

Issue #16, opened 2022-01-18
Move everything except public functionality to detail namespace


msimberg:
The public API of pika is small: sender/receiver functionality, runtime initialization, what else?

Hidden functionality can then gradually be brought into the public namespace through `pika::experimental::` or directly into `pika::`.

PR #179, opened 2022-04-07
Clean up topology module

aurianer:
Fixes partially #16
- Moves all topology symbols into the detail namespace

TODO:
- [x] Mark `constexpr`, `noexcept` the functionalities when relevant

### DIFF
--- a/libs/pika/topology/src/topology.cpp
+++ b/libs/pika/topology/src/topology.cpp
@@ -1585,7 +1585,7 @@ namespace pika::threads::detail {
     ///////////////////////////////////////////////////////////////////////////
     struct hw_concurrency
     {
-        hw_concurrency()
+        hw_concurrency() noexcept
 #if defined(__ANDROID__) && defined(ANDROID)
           : num_of_cores_(::android_getCpuCount())
 #else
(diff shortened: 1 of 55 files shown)


aurianer:
test cscs

libs/pika/topology/src/topology.cpp
```diff
@@ -1579,34 +1576,32 @@ namespace pika { namespace threads {
     }
 
     ///////////////////////////////////////////////////////////////////////////
-    topology& create_topology()
+    topology& create_topology() noexcept
```
msimberg, line 1579: This shouldn't be noexcept because the `topology` constructor isn't noexcept (and it shouldn't be noexcept because it can throw an exception right there in the constructor body).

libs/pika/topology/src/topology.cpp
```diff
@@ -1579,34 +1576,32 @@ namespace pika { namespace threads {
...
     }
 
     ///////////////////////////////////////////////////////////////////////////
-    namespace detail {
-        struct hw_concurrency
-        {
-            hw_concurrency()
+    struct hw_concurrency
+    {
+        hw_concurrency()
```
msimberg, line 1588: This should maybe be noexcept... unsure about the android call, but I would guess `hwloc_hardware_concurrency` is a C function.

msimberg (approved):
Thanks!

msimberg:
bors merge

PR #180, opened 2022-04-07
Change reference counting again in ensure_started

### LINKED ISSUE #60: Potential bug in `thread_pool_scheduler`
Try to see if this can be reproduced: https://app.circleci.com/pipelines/github/pika-org/pika/133/workflows/00b0996c-601f-4f6c-8f1c-f3b533738a4a/jobs/1462/steps?invite=true#step-103-164.

### LINKED ISSUE #174: Potential bug in `ensure_started`
The following test seems to segfault with `ensure_started`, in `ensure_started`:
```
auto test_senders()
{
    ex::execute(
        ex::with_annotation(ex::thread_pool_scheduler{}, "0-execute"), [] {});

    {
        pika::scoped_annotation ann{"0-execute-parent-annotation"};
        ex::execute(ex::thread_pool_scheduler{}, [] {});
        ex::execute(ex::with_annotation(ex::thread_pool_scheduler{},
                        "0-execute-with-annotation-override"),
            [] {});
        ex::execute(ex::with_annotation(ex::thread_pool_scheduler{},
                        "0-this-should-not-become-an-annotation"),
            pika::annotated_function([] {}, "0-execute-annotated-function"));
    }

    auto s1 = ex::schedule(
        ex::with_annotation(ex::thread_pool_scheduler{}, "0-schedule"));
    auto s2 = ex::schedule(ex::thread_pool_scheduler{}) |
        ex::then(pika::annotated_function(
            [] { dummy_task(1000); }, "0-schedule-then"));
    auto s3 = ex::just() |
        ex::transfer(
            ex::with_annotation(ex::thread_pool_scheduler{}, "0-transfer"));

    auto s4 = [] {
        pika::scoped_annotation ann{"0-schedule-parent-annotation"};
        auto s1 =
            ex::schedule(ex::thread_pool_scheduler{}) | ex::ensure_started();
                                                     // ^ adding this segfaults, why?

        return ex::when_all(std::move(s1));
    }();

    return ex::when_all(std::move(s1), std::move(s2), std::move(s3));
}
```
Is there a bug in `ensure_started`? Without `ensure_started` started all is fine. This may in fact be the same bug as #60.

msimberg:
Keep one reference alive in the `ensure_started_receiver` in case the `ensure_started_sender` is discarded without being connected.

Fixes #174. I improved the situation for `split` in #111, but made it worse for `ensure_started`. If the sender from `ensure_started` is discarded (which is allowed) the shared state would be released immediately because the receiver from `ensure_started` only kept a reference to the shared state (not a counted reference). This makes the reference an `intrusive_ptr` again. After #143 `ensure_started` sends rvalues to connected receivers which means that the `ensure_started_receiver` holding onto stored values is less of a concern. ~Just in case, I'm still resetting the shared state after continuations have been signaled.~ (Edit: the extra reset is not necessary, and actively harmful) I'm not 100% confident that I'm not introducing yet another bug with this, but I'm again slightly more certain that this goes in the right direction. Until the next one...

I could eventually reproduce at least one segfault coming from the `ensure_started` tests in `thread_pool_scheduler_test`. This seems to fix those as well. I.e. possibly fixes #60.

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/ensure_started.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/ensure_started.hpp
@@ -152,21 +152,21 @@ namespace pika { namespace execution { namespace experimental {
 
                 struct ensure_started_receiver
                 {
-                    shared_state& state;
+                    pika::intrusive_ptr<shared_state> state;
 
                     template <typename Error>
                     friend void tag_invoke(set_error_t,
                         ensure_started_receiver&& r, Error&& error) noexcept
                     {
-                        r.state.v.template emplace<error_type>(
+                        r.state->v.template emplace<error_type>(
                             error_type(PIKA_FORWARD(Error, error)));
-                        r.state.set_predecessor_done();
+                        r.state->set_predecessor_done();
                     }
 
                     friend void tag_invoke(
                         set_done_t, ensure_started_receiver&& r) noexcept
                     {
-                        r.state.set_predecessor_done();
+                        r.state->set_predecessor_done();
                     };
 
                     // This typedef is duplicated from the parent struct. The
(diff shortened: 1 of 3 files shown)


msimberg:
test cscs

msimberg:
test cscs

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

PR #182, opened 2022-04-08
Remove enabler in `then`'s `set_value`

### LINKED ISSUE #181: User-provided lambdas with auto parameters are instantiated with receiver types
The following currently fails to compile with `auto` in the lambda because the lambda gets instantiated with a receiver type. An explicit type `int` compiles correctly.
```
sync_wait(transfer_just(thread_pool_scheduler{}, 3) | then([](auto x) { return x + 1; }));
```
Find out if this is "user error" and the lambda must be correctly constrained, or if the instantiation can be avoided in pika. Does the P2300 reference implementation compile correctly with `auto`?

havogt:
Fixes #181

The problem is that `tag_invoke` of any tag instantiates the enable_if, before the function is excluded from the overload set. For generic lambdas the body has to be instantiated to get the return type which is a hard error.
To my understanding there is nothing to disambiguate in the `then::tag_invoke(set_value, ...)` and it can be removed (alternatively the enabler can be put in a trailing return type).

## Any background context you want to provide?

Maybe this is a general problem of the tag_invoke mechanism? C++20 constraints have the same problem. I am wondering about compile time of codes with many tag_invoke overloads if all non-trailing constraints might have to be evaluated. Or am I missing something?

## Checklist

- [ ] I have added a new feature and have added tests to go along with it.
- [x] I have fixed a bug and have added a regression test.

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/then.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/then.hpp
@@ -91,8 +91,7 @@ namespace pika { namespace execution { namespace experimental {
                     });
             }
 
-            template <typename... Ts,
-                typename = std::enable_if_t<pika::is_invocable_v<F, Ts...>>>
+            template <typename... Ts>
             friend void tag_invoke(
                 set_value_t, then_receiver_type&& r, Ts&&... ts) noexcept
             {
--- a/libs/pika/execution/tests/unit/algorithm_then.cpp
+++ b/libs/pika/execution/tests/unit/algorithm_then.cpp
@@ -66,6 +66,17 @@ int main()
 
     {
         std::atomic<bool> set_value_called{false};
+        auto s = ex::then(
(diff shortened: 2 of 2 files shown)


msimberg (approved):
Thanks @havogt, much appreciated! I think this is a safe change. The reason we had that `enable_if` there in the first place was to satisfy GCC 7 (possibly 8 as well) which would in fact instantiate (or attempt to) instantiate e.g. `tag_invoke(set_value_t, ...`) even if the tag is different. However, my memory on this is hazy. clang never had problems without the `enable_if`. Since we now require GCC 9 at minimum this should be safe.

I think with C++20 one could argue that the lambda should in fact be constrained with `[](std::integral auto x) { ... }` (or `addable` or something similar), which also avoids at least this particular problem.

msimberg:
test cscs

msimberg:
bors merge

PR #188, opened 2022-04-11
Inherit annotations from scheduling task instead of spawning task in `transfer` for `thread_pool_scheduler`

### LINKED ISSUE #172: Allow for more flexibility in inheriting annotations

msimberg:
Fixes #172.

Changes the behaviour of `execute` and `schedule` for `thread_pool_sheduler` to take a fallback annotation from the calling context. Also customizes `schedule_from` for `thread_pool_scheduler` in order to get the fallback annotation from the context where `transfer` is called. The default `transfer` implementation otherwise calls `schedule` only when it's time to schedule the new task, which would inherit the wrong annotation without the `schedule_from` customization.

Open issues that will likely be left out from this PR:
- APEX develop seems to have changed the output format when screen output is enabled. I think we will have to change how the annotation tests check for correct output, but I'm still targeting this against the latest release of APEX.
- The `schedule_from` customization for `thread_pool_scheduler` currently relies on knowing the sender type of the default `schedule_from` implementation. We should think about how we can customize an operation but then also be able to dispatch to the default implementation (in this case we only want to add an annotation to the scheduler and then use the default implementation).
- The behaviour of annotations for other adaptors such as `bulk`, `when_all` etc. haven't really been considered. I think they are covered by the changes here but I'll add some more notes to #172 about those cases tomorrow.

### DIFF
--- a/libs/pika/executors/include/pika/executors/thread_pool_scheduler.hpp
+++ b/libs/pika/executors/include/pika/executors/thread_pool_scheduler.hpp
@@ -11,6 +11,7 @@
 #include <pika/coroutines/thread_enums.hpp>
 #include <pika/errors/try_catch_exception_ptr.hpp>
 #include <pika/execution/algorithms/execute.hpp>
+#include <pika/execution/algorithms/schedule_from.hpp>
 #include <pika/execution/executors/execution_parameters.hpp>
 #include <pika/execution_base/receiver.hpp>
 #include <pika/execution_base/sender.hpp>
@@ -135,9 +136,9 @@ namespace pika { namespace execution { namespace experimental {
         }
 
         template <typename F>
-        void execute(F&& f) const
+        void execute(F&& f, char const* fallback_annotation) const
         {
-            pika::util::thread_description desc(f, annotation_);
+            pika::util::thread_description desc(f, fallback_annotation);
             threads::thread_init_data data(
                 threads::make_thread_function_nullary(PIKA_FORWARD(F, f)), desc,
                 priority_, schedulehint_, stacksize_);
@@ -148,7 +149,7 @@ namespace pika { namespace execution { namespace experimental {
         friend void tag_invoke(
             execute_t, thread_pool_scheduler const& sched, F&& f)
         {
-            sched.execute(PIKA_FORWARD(F, f));
+            sched.execute(PIKA_FORWARD(F, f), sched.get_fallback_annotation());
         }
 
         template <typename Scheduler, typename Receiver>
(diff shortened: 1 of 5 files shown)


msimberg:
test cscs

aurianer:
bors merge

PR #193, opened 2022-04-13
Replace with slimmer docker image

### LINKED ISSUE #10: Use slimmer container image for CI
We currently still use https://hub.docker.com/r/stellargroup/build_env. We could do with a much lighter image without documentation tools and e.g. python.

aurianer:
Fixes #10

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ working_dir_default: &working_dir_default
 
 docker_default: &docker_default
     docker:
-      - image: stellargroup/build_env:10
+      - image: pikaorg/pika-ci-base:1
 
 defaults: &defaults
     <<: *working_dir_default
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,7 +193,6 @@ jobs:
                 -DPIKA_WITH_EXAMPLES=On \
                 -DPIKA_WITH_TESTS=On \
                 -DPIKA_WITH_TESTS_HEADERS=On \
-                -DPIKA_WITH_DATAPAR_VC=On \
                 -DPIKA_WITH_COMPILER_WARNINGS=On \
                 -DPIKA_WITH_COMPILER_WARNINGS_AS_ERRORS=On \
                 -DPIKA_WITH_DEPRECATION_WARNINGS=On \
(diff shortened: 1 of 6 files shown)


.github/workflows/linux_hip.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   build:
     name: github/linux/hip/fast
     runs-on: ubuntu-latest
-    container: stellargroup/hip_build_env:10
+    container: pikaorg/pika-ci-hip:1
```
msimberg, line 20: Would it possible/easier to have HIP in the base image as well? Size is definitely one argument... Just curious what you think. I don't mind either way.

aurianer: The argument to have them separate is that hip is is quite long to install and we would need to do it everytime we build and push the image. But it's not a so strong argument since we don't update the image that often so I don't mind. Let me know what you prefer :)

msimberg: Fair enough. Let's keep them separate at the moment (the title does say "slimmer image" after all...).

aurianer:
Technically this can go as the docker image is already on docker hub (since the push was done on the PR not at the merge) but I don't mind if we wait to merge the pika-ci-image PR before

msimberg:
bors try

msimberg:
Vc support needs to be removed before this can be merged.

aurianer:
bors try

.circleci/config.yml
```diff
@@ -497,11 +494,7 @@ jobs:
           name: Installing
           command: |
               ./bin/hello_world --pika:bind=none
-              # Disable Vc here as the target machine of the container image may
-              # not support the same vector intrinsics as this machine. The
-              # "Test Docker Image" below fails with "Illegal instruction" if VC
-              # is enabled, and other machines may fail similarly.
-              cmake -DPIKA_WITH_DATAPAR_VC=Off .
+              cmake .
```
msimberg, line 497: This can be completely removed.

msimberg (approved):
Just that one unnecessary `cmake` call to remove, let's merge this afterwards. Thanks @aurianer!

msimberg:
bors merge

Issue #12, opened 2022-01-18
Remove/replace remaining Boost includes


msimberg:
It would now be feasible to completely remove all dependencies on Boost. Some things would naturally remove Boost dependencies:
- #6 
- #7 
- #11

Other headers can likely be brought into `pika`.

msimberg:
Current situation is the following.

Main library, include counts per boost header:
```
> rg --no-heading --no-line-number --no-filename --no-messages '#include\W+boost' libs/pika/*/{include,src}/**/*.{cpp,hpp,cu} | sort | uniq -c | sort -r
      7 #include <boost/tokenizer.hpp>
      7 #include <boost/shared_array.hpp>
      5 #include <boost/version.hpp>
      2 #include <boost/variant.hpp>
      2 #include <boost/lockfree/queue.hpp>
      2 #include <boost/exception/exception.hpp>
      2 #include <boost/exception/diagnostic_information.hpp>
      2 #include <boost/dynamic_bitset.hpp>
      2 #include <boost/config.hpp>
      1 #include <boost/shared_ptr.hpp>
      1 #include <boost/range/irange.hpp>
      1 #include <boost/multi_array.hpp>
      1 #include <boost/lockfree/stack.hpp>
      1 #include <boost/lockfree/policies.hpp>
      1 #include <boost/lockfree/detail/tagged_ptr.hpp>
      1 #include <boost/lockfree/detail/prefix.hpp>
      1 #include <boost/iterator/iterator_categories.hpp>
      1 #include <boost/intrusive/slist.hpp>
      1 #include <boost/intrusive_ptr.hpp>
      1 #include <boost/functional/hash.hpp>
      1 //    #include <boost/detail/utf8_codecvt_facet.ipp>
      1 #include <boost/context/detail/fcontext.hpp>
      1 #include <boost/container/small_vector.hpp>
      1 #include <boost/array.hpp>
```
Main library, includes per pika file:
```
> rg --no-heading --no-line-number --no-messages '#include\W+boost' libs/pika/*/{include,src}/**/*.{cpp,hpp,cu} | sort | uniq
libs/pika/affinity/include/pika/affinity/parse_affinity_options.hpp:#include <boost/variant.hpp>
libs/pika/algorithms/include/pika/parallel/algorithms/copy.hpp:#include <boost/shared_array.hpp>
[...]

PR #202, opened 2022-04-20
Remove uses of `boost::shared_array` in favour of `std::shared_ptr<T[]>`

msimberg:
Part of #12.

I think we can do this, but I'm not 100% sure. Let's see what the CI configurations say.


msimberg:
bors try

msimberg:
Requiring clang 11 may be quite reasonable in the near future. Cray's clang is already on 12. It looks like DLA-Future would have no problems with requiring clang 11. The new LTS version of Ubuntu (22.04) also seems to include clang 11.

msimberg:
Agreed in meeting to require clang 11, pending on what underlying LLVM version ROCm is using in 4.3. If older than 11, we can most likely bump the ROCm version requirement as well.

msimberg:
ROCm 4.3.0 seems to be based on clang 13. Though probably a prerelease since officially clang 13 was released after ROCm 4.3.0. ROCm 4.3.0 was released significantly after clang 11 though so we should be safe there.

msimberg:
#239 updated the clang requirement to version 11 so we can now go ahead with this.

msimberg:
bors merge

msimberg:
bors merge

PR #214, opened 2022-05-02
Allow using `prefer` on scheduling properties

### LINKED ISSUE #208: Scheduling properties should not have a fallback `static_assert` when unsupported
We currently have a fallback case with a `static_assert(false)` and a helpful error message if a property is not supported by whatever someone is trying to apply it to: https://github.com/pika-org/pika/blob/7df15bf3cf8770b47462b9c553ce3d149a863deb/libs/pika/async_base/include/pika/async_base/scheduling_properties.hpp#L14-L38. This is nice for the case that one actually requires a property. However, it prevents the use of `prefer` since we can't distinguish between actually supporting a property and the fallback error case (the fallback is always callable, but fails to compile). The `prefer` case is more important than the error message, so we should change the scheduling properties here to just use the plain `tag` helper base class: https://github.com/pika-org/pika/blob/7df15bf3cf8770b47462b9c553ce3d149a863deb/libs/pika/async_base/include/pika/async_base/scheduling_properties.hpp#L41-L78.

This should possibly even go into 0.4.0.

msimberg:
Fixes #208.

## Checklist

- [X] I have fixed a bug and have added a regression test.

### DIFF
--- a/libs/pika/async_base/include/pika/async_base/scheduling_properties.hpp
+++ b/libs/pika/async_base/include/pika/async_base/scheduling_properties.hpp
@@ -8,72 +8,48 @@
 
 #include <pika/config.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
+#include <pika/functional/tag_invoke.hpp>
 
 namespace pika::execution::experimental {
-    namespace detail {
-        template <typename Tag, typename... Args>
-        struct property_not_supported
-        {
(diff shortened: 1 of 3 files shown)


msimberg:
bors try

msimberg:
bors try

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

PR #226, opened 2022-05-09
Reduce the number of iterations in `cross_pool_injection` test

msimberg:
Reduces the number of iterations, but hopefully still keeps the test useful for triggering bugs. With 1000 iterations the test takes around 2 minutes on CircleCI and is clearly the longest-running test we have. I'm hoping this will bring it down to around a minute.


msimberg:
bors merge

PR #227, opened 2022-05-09
Increase parallelism on CircleCI

msimberg:


aurianer (approved):
LGTM thanks!

aurianer:
bors merge

msimberg:
The CI environment probably has a resource constraint to 4 cores, but the environment actually reports 36 cores meaning that we can't use an "unconstrained ninja". I'll update this to use `ninja -j4` explicitly instead (was `-j3`).

PR #224, opened 2022-05-06
Format files with black and shfmt

msimberg:
Fixes #66. This also updates the clang-format requirement to version 14.

~Still needs a CI image bump.~

Related CI PR: https://github.com/pika-org/pika-ci-image/pull/10.

To do:
- [x] bump CI image version
- [x] add formatting steps to required bors jobs

PR #232, opened 2022-05-13
Fix unicode symbols in `plot.py`

msimberg:
They got garbled in https://github.com/pika-org/pika/pull/224, though it's not clearly black that is the cause of it. Running black again on the file did not change the symbols, so I'm not quite sure what happened.


msimberg:
launch perftests

msimberg:
bors merge

PR #242, opened 2022-05-18
Use process mask by default

### LINKED ISSUE #55: Make `--pika:use-process-mask` the default?

msimberg:
Fixes #55. Still needs testing in real-world situations but the basic change is simple. This replaces `--pika:use-process-mask` with `--pika:ignore-process-mask`. By default a process mask will be used (which will be the full machine if no specific bindings are set by e.g. slurm or `hwloc-bind`).

I've added some warning messages for the cases when the process mask forces the runtime to use only one worker thread or only one core.


msimberg:
bors try

msimberg:
bors try

msimberg:
@albestro, @rasolca, or @teonnik, if one of you has time to look at this I would very much appreciate it. In summary everything that was possible before is still possible, only the default behaviour changes with this PR. For the cases that there happens to be a process mask set and it happens to contain only one worker thread or core (possibly by mistake) I'm printing a warning. I'm not sure if there needs to be a way to silence the warning as well?

libs/pika/command_line_handling/src/command_line_handling.cpp
```diff
@@ -967,6 +961,37 @@ namespace pika::detail {
...
+                std::cerr
+                    << "The pika runtime will be started with only one worker "
+                       "thread because the process mask has restricted the "
+                       "available resources to only one thread or the system "
+                       "only has one core. If this is unintentional make sure "
+                       "the process mask contains the resources you need or "
+                       "use --pika:ignore-process-mask to use all resources. "
+                       "Use --pika:print-bind to print the thread bindings "
+                       "used by pika.\n";
+            }
```
rasolca, line 979: Will this warning be printed even when the user specifies `pika:threads=1`? I would find this annoying.

Maybe it also worth to print the message only if the system has more than one core.

msimberg: Yes, it would currently do that. These are good suggestions and I'll update this to take that into account.

msimberg: @rasolca I've udpated the conditions for printing the warning message. It is now printed only if:
1. The number of worker threads (or cores) is 1,
2. The default number of worker threads is bigger than one (i.e. what you would get without a process mask), and
3. `--pika:threads` (or `--pika:cores`) is not given as a command line argument.

I think that should make it print only in cases where one might be surprised that the number of worker threads is 1.

msimberg:
bors try

libs/pika/command_line_handling/src/command_line_handling.cpp
rasolca: Thanks!

PR #243, opened 2022-05-18
Remove `then_with_any_cuda` and related unused utilities

msimberg:

PR #247, opened 2022-06-01
Add back SFINAE to `then_with_cuda_stream_sender` for DLA-Future

msimberg:
#243 also removed the SFINAE prematurely from `then_with_cuda_stream_sender`. It will be possible to remove it once https://github.com/eth-cscs/DLA-Future/pull/459 is in, but before that the SFINAE is still required.


msimberg:
bors merge

Issue #191, opened 2022-04-12
Occasionally failing tests


msimberg:
This issue is meant to collect the names of failing tests together with links to the failures for investigation. This is a long-running issue and will likely not be closed by one or only a few PRs.

- [ ] `shared_mutex2`:
  - https://cdash.cscs.ch/test/9848329

msimberg:
I will attempt to look at the `std_thread_scheduler` failure for 0.6.0.

msimberg:
These are some of the more frequently failing tests on CDash: https://cdash.cscs.ch/testOverview.php?project=pika&date=2023-01-31&group=28.

PR #250, opened 2022-06-08
Attempt to wait for detached thread to finish in `std_thread_scheduler_test`

msimberg:
I think this fixes part of #191 (the `std_thread_scheduler_test` failures). The sender is still discarded immediately so that part is still somewhat tested, but we attempt to wait for the detached `std::thread` to avoid having it possibly start after exiting main.

### DIFF
--- a/libs/pika/executors/tests/unit/std_thread_scheduler.cpp
+++ b/libs/pika/executors/tests/unit/std_thread_scheduler.cpp
@@ -9,8 +9,10 @@
 
 #include <atomic>
 #include <chrono>
+#include <condition_variable>
 #include <cstddef>
 #include <exception>
+#include <mutex>
 #include <stdexcept>
 #include <string>
 #include <thread>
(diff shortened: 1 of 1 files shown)


msimberg:
This didn't quite work as well as I'd hoped. Still some work to do here.

msimberg:
I've reverted this in https://github.com/pika-org/pika/commit/e34406cf75d8d77b32fb2d822e37cb68836a79d5 as I think my analysis was bogus.

The problem smells a lot like https://sourceware.org/bugzilla/show_bug.cgi?id=28028 (and possibly https://sourceware.org/bugzilla/show_bug.cgi?id=22711), or may at least be related to it. The segfault comes from the malloc implementation, which doesn't rule out us corrupting it, but it is suspicious. Setting `GLIBC_TUNABLES=glibc.malloc.tcache_count=0` gets rid of the segfault, but again that may still be a side effect of us doing something wrong. Since we don't see this failure at all on the other CI configurations I do suspect a glibc bug but can't back it up at the moment with evidence.

Issue #64, opened 2022-01-21
Add tracy support


msimberg:

msimberg:
Will attempt this for 0.5.0, but it may require more time than a month.

msimberg:
The spack part of this is ongoing in https://github.com/spack/spack/pull/30588, but the pika support will not be in 0.5.0.

msimberg:
Part of this was implemented and merged in #252, but leaving open for the other open items in the issue description.

PR #252, opened 2022-06-09
Add basic Tracy support

msimberg:
Adds basic Tracy support, fixing part of #64. "basic" means that this can currently only be used with threads that never yield (stackful threads are ok if they don't yield, or stackless threads to be sure that they never yield). This is because of the way scoped annotations work in Tracy (but also in e.g. APEX, where we have to save and restore annotations on suspend and resume). Because of this limitation this is also not enabled in CI.

- `PIKA_WITH_TRACY` enables the use of Tracy. Tracy has to be installed separately, there is no fetchcontent support.
- I've updated the thread names to be of the form `pika/worker-thread/pool:default/global:1/local:1`.
- Every task invocation is wrapped in an annotation of the form `pika/task:0xabcd.../task_name`.

There are two independent next steps from this:
1. Start using stackless threads as much as possible since that is easier to do with senders.
2. Add support for fibers or saving/restoring the annotations on suspend/resume.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1000,6 +1000,21 @@ if(PIKA_WITH_APEX)
   endif()
 endif()
 
+pika_option(
+  PIKA_WITH_TRACY BOOL "Enable Tracy instrumentation support." OFF
+  CATEGORY "Profiling"
+)
+if(PIKA_WITH_TRACY)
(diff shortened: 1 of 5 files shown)


msimberg:
bors try

msimberg:
bors merge

PR #257, opened 2022-06-15
Clean up coroutines module

msimberg:
Part of #16.

Cleans up the `coroutines` module by moving nearly everything into the detail namespace. To keep things simple I've moved some related functionality in other modules to the `pika::threads::detail` namespace as well. Those modules are not considered "cleaned up" and will still be worked on as part of #16.

@albestro, @biddisco, @rasolca, @teonnik:

This introduces a breaking change for `thread_priority`, `thread_stacksize`, and `thread_schedule_hint`. I've moved them all to `pika::execution` (from `pika::threads`). I think this is a more appropriate namespace for them long term. They're clearly part of the public API since all three properties are reasonable things for users to customize (and DLA-Future uses at least `thread_priority` so they should not be in `detail`. I've chosen `pika::execution` because all of P2300 is proposed to be in `std::execution` and I see these as properties related to execution. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2079r2.html also puts `system_context` and `system_scheduler` in `std::execution`. Finally, there is no `std::threads` namespace. The main argument I see against putting them in `pika::execution` is that the properties may not apply to all execution contexts. However, we already know that `thread_priority` (maybe it should just be `priority` or `scheduling_priority`, but that's for another day) applies to at least pika threads and CUDA streams. The top-level `pika` namespace would be another option, but given that we already have `pika::execution` and the properties are related to execution I think it fits. Thoughts?


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors merge

msimberg:
bors merge

msimberg:
bors merge

msimberg:
bors merge

PR #262, opened 2022-06-16
Add `drop_value` sender adaptor

msimberg:
Adds a `drop_value` sender adaptor. It ignores any values sent by a predecessor sender. E.g.:
```
just(1, "hello", Tile{}) | drop_value() | then([](/* nothing here */) {...});
```
This is useful for https://github.com/eth-cscs/DLA-Future/pull/572.

I went for singular `value` instead of `values` to mirror `let_value/error` and `upon_error/stopped` (`then` is "`upon_value`") which all use singular and it mirrors the "channel name" (`value`, `error`, `stopped`). An alternative to `drop` would perhaps be `ignore`.

Note that I'm not adding `drop_error` and `drop_stopped` because 1. I don't want to encourage that and 2. it's not quite clear what they would do (sending no values in `set_value` is fine, but sending no error in `set_error` makes no sense and isn't allowed).

Opinions?

### DIFF
--- a/libs/pika/execution/CMakeLists.txt
+++ b/libs/pika/execution/CMakeLists.txt
@@ -12,6 +12,7 @@ set(execution_headers
     pika/execution/algorithms/detail/partial_algorithm.hpp
     pika/execution/algorithms/detail/predicates.hpp
     pika/execution/algorithms/detail/single_result.hpp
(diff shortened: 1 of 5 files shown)


msimberg:
bors try

msimberg:
bors merge

PR #273, opened 2022-06-23
Remove mimalloc version constraint

msimberg:
Not necessary since we are not relying on mimalloc APIs, but only linking to it for a faster malloc.


msimberg:
bors merge

msimberg:
bors merge

PR #274, opened 2022-06-23
Increase time tolerance in `wait_for_1751` test

msimberg:
This often fails with waiting times just above the tolerance (~1.15) so I'm increasing it.


msimberg:
bors merge

PR #282, opened 2022-06-30
Fix handling on reference types in sent by predecessors to `ensure_started` and `schedule_from`

msimberg:
Fixes handling of reference types sent to `ensure_started` and `schedule_from`. Other algorithms also need to be checked for the same problem, but I'm doing it in separate PRs.


msimberg:
Confirmed working by @teonnik.

msimberg:
bors merge

PR #286, opened 2022-07-01
Add priorities for CUDA polling

msimberg:
Slight refactoring of the CUDA polling implementation. The actual polling mechanism is completely unchanged. I've only refactored the queues and free functions into a class so that it's easier to create multiple instances. There are now two polling queues, one for normal priority streams and one for high priority streams. The interface for adding callbacks is almost unchanged. The only difference is that it now requires a `cuda_stream` which has an embedded `pika::execution::thread_priority` instead of having to query a `cudaStream_t`.

Performance is minimally improved with small blocksizes in DLA-Future, so this is primarily about refactoring the code to be slightly more manageable, not a performance improvement.


msimberg:
bors try

aurianer (approved):
LGTM thanks!

PR #287, opened 2022-07-01
Fix `--pika:ignore-process-mask`

msimberg:
The process mask was being taken into account for the _number_ of threads, but not for the actual affinity data (in `affinity_data`). This lead me to believe that things were working ok, but things were apparently slightly off. This meant that e.g. multiple ranks on a single node had overlapping thread bindings.

In an attempt to simplify things I've renamed some variables back to what they were before #242. Now everything internal is again called `use_process_mask` whereas _only_ the command line and configuration options are called `ignore_process_mask`, and the given options are translated to `use_process_mask` as soon as possible. I prefer `use_process_mask` internally since it's the "positive" form.


msimberg:
I should add that I would really like to be able to add tests for this but I'm lacking good ideas to do so.

msimberg:
bors merge

msimberg:
bors merge

PR #297, opened 2022-07-06
Fix Tracy integration

msimberg:
Fixes the build with Tracy, and since I managed to break the build in record time I'm also adding a simple GitHub actions workflow which builds examples and runs `hello_world`. We still can't run all executables because threads can't yield with the Tracy integration.

### DIFF
--- a/.github/workflows/linux_tracy.yml
+++ b/.github/workflows/linux_tracy.yml
@@ -1,0 +1,90 @@
+# Copyright (c) 2022 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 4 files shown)


.github/workflows/linux_tracy.yml
```diff
@@ -0,0 +1,81 @@
...
+on:
+  push:
+    # TODO: Uncomment before merging
+    # branches:
+    #   # Development and release branches
+    #   - main
+    #   - release**
+    #   # Bors branches
+    #   - trying
+    #   - staging
```
msimberg, line 18: To do before merging.

aurianer (approved):
LGTM thanks!

msimberg:
bors merge

PR #298, opened 2022-07-06
Fix compilation of `mpi_async_storage` test

msimberg:
Fixes things that were missed in #277.

@biddisco can you confirm that the mutex really was intended to be unused?

### DIFF
--- a/libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
+++ b/libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
@@ -423,8 +423,9 @@ int pika_main(pika::program_options::variables_map& vm)
         *reinterpret_cast<uint64_t*>(&local_recv_storage[i]) = temp;
     }
     //
-    uint64_t num_slots =
-        1024 * 1024 * options.local_storage_MB / options.transfer_size_B;
+    std::uint64_t num_slots = static_cast<std::uint64_t>(1024) *
+        static_cast<std::uint64_t>(1024) * options.local_storage_MB /
+        options.transfer_size_B;
     nws_deb<6>.debug(
         "num ranks ", nranks, ", num_slots ", num_slots, " on rank", rank);
     //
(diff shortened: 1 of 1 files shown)


aurianer (approved):
LGTM thanks!

msimberg:
> There's really no need to use `std::uint64_t` and friends in place of `uint64_t` - I specifically removed the `std::` because it jjust makes lines longer for no extra benefit.

I agree that it makes lines longer. However:

- The global namespace typedefs are not part of the C++ standard, they are only there because of how standard libraries implement `cstddef`
- We don't leave out the `std::` anywhere else (at least not without doing the below).

If you really want the shorter names let's do `using namespace std` or `using std::uint64_t` in the test.

Unrelated, what are these for and can they be removed: https://github.com/pika-org/pika/blob/71f122c2128f6519b215d43ed981d52365fe9492/libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp#L52-L53?

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -541,7 +538,7 @@ int main(int argc, char* argv[])
         "The total storage will be num_ranks * localMB");
 
     cmdline.add_options()("in-flight-limit",
-        pika::program_options::value<std::uint64_t>()->default_value(64),
+        pika::program_options::value<std::uint64_t>()->default_value(256),
```
msimberg, line 541: To do: what's a reasonable value that's big enough not to deadlock but to still test the throttling facility?

msimberg: @biddisco the test seems to have passed at least once now with the higher value. I'll rerun it a few times, but assuming they all pass do you think 256 is a decent value to have here as the default?

msimberg:
bors try

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
msimberg: It timed out once more, but also passed once more. This needs more investigation, but I'm going to merge this in any case since this is already an improvement.

msimberg:
bors merge

msimberg:
bors merge

Issue #184, opened 2022-04-11
Run MPI and CUDA sender continuations inline if request/event is done before adding continuation


msimberg:
It seems like in some situations it's possible that communication will complete either very quickly or be guaranteed to complete before a function returns (e.g. with small message sizes), and in those cases it's unnecessary to add a continuation to be polled. As an optimization, we could check if a request/event is done just before adding it to the polling loop, and if it is, just call the callback inline. I don't think CUDA would ever guarantee that a kernel completes inline, but for small kernels I could imagine them completing by the time we add a callback. The same optimization is potentially useful there as well.

msimberg:
This may (mostly speculation) also account for the difference between the current `yield_while` implementation in DLA-Future for MPI, and the polling version using the pika scheduler. With `yield_while` the predicate is checked before suspending, and if the request is already done no yielding will happen, effectively running the continuation inline.

PR #306, opened 2022-07-07
Eagerly check completion of CUDA kernels

msimberg:
Fixes the CUDA part of #184. However, this is untested and if it doesn't show any improvement (or worse, a performance drop) this shouldn't be merged.

Based on #286.

### DIFF
--- a/libs/pika/async_cuda/include/pika/async_cuda/custom_gpu_api.hpp
+++ b/libs/pika/async_cuda/include/pika/async_cuda/custom_gpu_api.hpp
@@ -70,6 +70,7 @@
     #define cudaStreamGetFlags hipStreamGetFlags
     #define cudaStreamNonBlocking hipStreamNonBlocking
     #define cudaStreamSynchronize hipStreamSynchronize
+    #define cudaStreamQuery hipStreamQuery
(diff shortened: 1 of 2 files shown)


msimberg:
bors try

msimberg:
This needed #325, but I will not have time to test this properly for 0.7.0. Moving to 0.8.0.

msimberg:
The result is that this seems to have mostly no effect but a small effect on some 256 blocksize configurations in DLA-Future. Since the change is small I think we can go with this change, even if the effect is quite minimal.

biddisco (approved):
Seems reasonable. Would be interesting to add a performance counter to see how often this is actually triggered in production (my guess is not often, but it doesn't cost much to check...)

aurianer:
bors merge

aurianer:
bors merge

msimberg:
> Seems reasonable. Would be interesting to add a performance counter to see how often this is actually triggered in production (my guess is not often, but it doesn't cost much to check...)

True. However, remember that there are no HPX-y performance counters anymore. We can of course add a simple counter specifically for this purpose.

PR #320, opened 2022-07-13
Add `when_all` test and handling for const reference error types

### LINKED ISSUE #284: Fix handling of `value_types` and `error_types` with references in all sender adaptors
See #282 for details.

These need to be checked and fixed if necessary:

- [x] `bulk` (#338)
- [x] `drop_value` (#338)
- [x] `ensure_started` (#338)
- [x] `let_error` (#338, #371)
- [x] `let_value` (#472)
- [x] `split` (#285)
- [x] `start_detached` (#474)
- [x] `sync_wait` (#285)
- [x] `then` (#338)
- [x] `transfer` (#338)
- [x] `when_all` (#320)
- [ ] `when_all_vector` (#484)

aurianer:
Partially fixes #284

### DIFF
--- a/libs/pika/execution/tests/include/algorithm_test_utils.hpp
+++ b/libs/pika/execution/tests/include/algorithm_test_utils.hpp
@@ -96,6 +96,42 @@ struct error_sender
     }
 };
 
+struct const_reference_error_sender
+{
+    template <template <class...> class Tuple,
+        template <class...> class Variant>
+    using value_types = Variant<Tuple<>>;
+
+    template <template <class...> class Variant>
+    using error_types = Variant<std::exception_ptr const&>;
+
+    static constexpr bool sends_done = false;
+
+    using completion_signatures =
+        pika::execution::experimental::completion_signatures<
+            pika::execution::experimental::set_error_t(std::exception_ptr&)>;
+
+    template <typename R>
+    struct operation_state
+    {
+        std::decay_t<R> r;
+        friend void tag_invoke(pika::execution::experimental::start_t,
+            operation_state& os) noexcept
+        {
+            auto const e = std::make_exception_ptr(std::runtime_error("error"));
+            pika::execution::experimental::set_error(std::move(os.r), e);
+        }
+    };
+
+    template <typename R>
+    friend operation_state<R> tag_invoke(
+        pika::execution::experimental::connect_t, const_reference_error_sender,
+        R&& r)
+    {
+        return {std::forward<R>(r)};
+    }
+};
+
 template <typename F>
 struct callback_receiver
 {
--- a/libs/pika/execution/tests/include/algorithm_test_utils.hpp
+++ b/libs/pika/execution/tests/include/algorithm_test_utils.hpp
@@ -66,7 +66,7 @@ struct error_sender
 
     using completion_signatures =
         pika::execution::experimental::completion_signatures<
-            pika::execution::experimental::set_value_t(),
+            pika::execution::experimental::set_value_t(Ts...),
             pika::execution::experimental::set_error_t(std::exception_ptr)>;
 
     template <typename R>
@@ -220,51 +220,6 @@ struct void_callback_helper
     void operator()(T&& t)
     {
         PIKA_INVOKE(std::move(f), std::forward<T>(t));
-    }
-};
-
-template <typename T>
-struct error_typed_sender
-{
-    template <template <class...> class Tuple,
-        template <class...> class Variant>
-    using value_types = Variant<Tuple<T>>;
-
-    template <template <class...> class Variant>
-    using error_types = Variant<std::exception_ptr>;
-
-    static constexpr bool sends_done = false;
-
-    using completion_signatures =
-        pika::execution::experimental::completion_signatures<
-            pika::execution::experimental::set_value_t(T),
-            pika::execution::experimental::set_error_t(std::exception_ptr)>;
-
-    template <typename R>
-    struct operation_state
-    {
-        std::decay_t<R> r;
-
-        friend void tag_invoke(pika::execution::experimental::start_t,
-            operation_state& os) noexcept
-        {
-            try
-            {
-                throw std::runtime_error("error");
-            }
-            catch (...)
-            {
-                pika::execution::experimental::set_error(
-                    std::move(os.r), std::current_exception());
-            }
-        };
-    };
-
-    template <typename R>
-    friend auto tag_invoke(
-        pika::execution::experimental::connect_t, error_typed_sender, R&& r)
-    {
-        return operation_state<R>{std::forward<R>(r)};
     }
 };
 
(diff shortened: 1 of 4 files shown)


aurianer:
bors try

libs/pika/execution/tests/include/algorithm_test_utils.hpp
```diff
@@ -96,6 +96,51 @@ struct error_sender
...
+
+    template <template <class...> class Variant>
+    using error_types = Variant<std::exception_ptr const&>;
+
+    static constexpr bool sends_done = false;
+
+    using completion_signatures =
+        pika::execution::experimental::completion_signatures<
+            pika::execution::experimental::set_value_t(),
+            pika::execution::experimental::set_error_t(std::exception_ptr&)>;
```
msimberg, line 114: Unsure if this will compile, but it would be the correct thing to do:
```suggestion
struct const_reference_error_sender
{
    template <template <class...> class Tuple,
        template <class...> class Variant>
    using value_types = Variant<>;

    template <template <class...> class Variant>
    using error_types = Variant<std::exception_ptr const&>;

    static constexpr bool sends_done = false;

    using completion_signatures =
        pika::execution::experimental::completion_signatures<
            pika::execution::experimental::set_error_t(std::exception_ptr&)>;
```
If the above doesn't compile do this:
```suggestion
struct const_reference_error_sender
{
    template <template <class...> class Tuple,
        template <class...> class Variant>
    using value_types = Variant<Tuple<>>;

    template <template <class...> class Variant>
    using error_types = Variant<std::exception_ptr const&>;

    static constexpr bool sends_done = false;

    using completion_signatures =
        pika::execution::experimental::completion_signatures<
            pika::execution::experimental::set_value_t(),
            pika::execution::experimental::set_error_t(std::exception_ptr&)>;
```

Since GitHub doesn't seem to highlight the changes, both variants untemplate the struct since you will never send a `T`. It's not necessary. The first variant changes the `value_types` to send nothing and removes the `set_value_t()` in `completion_signatures`. The second variant just sends no value, i.e. `Tuple<>` and keeps the `set_value_t()` in `completion_signatures`.

libs/pika/execution/tests/include/algorithm_test_utils.hpp
```diff
@@ -96,6 +96,51 @@ struct error_sender
...
+            {
+                throw std::runtime_error("error");
+            }
+            catch (...)
+            {
+                auto const& e = std::current_exception();
+                pika::execution::experimental::set_error(
+                    std::move(os.r), e);
+            }
+        }
```
msimberg, line 133: I forgot about `make_exception_ptr` (https://en.cppreference.com/w/cpp/error/make_exception_ptr):
```suggestion
            pika::execution::experimental::set_error(
                    std::move(os.r), std::make_exception_ptr(std::runtime_error("error")));
        }
```

msimberg (changes requested):
Some minor changes required, but looks good otherwise!

msimberg:
Merging is of course currently not going to work with bors since daint is down. Let's wait a few days to see if they can resolve the problems. Otherwise we can hopefully go with only the GitHub actions and CircleCI checks.

libs/pika/execution/tests/include/algorithm_test_utils.hpp
aurianer: So removing the `set_value` from the `completion_signatures` is working but the `Variant<>` fail on a `static_assert` checking that size of T is strictly greater than 0 so I'm going with the `Variant<Tuple<>>` solution

aurianer:
bors try

libs/pika/execution/tests/include/algorithm_test_utils.hpp
msimberg: Just for the record, can you put here which static assert that was?

libs/pika/execution/tests/include/algorithm_test_utils.hpp
```diff
@@ -66,7 +66,6 @@ struct error_sender
 
     using completion_signatures =
         pika::execution::experimental::completion_signatures<
-            pika::execution::experimental::set_value_t(),
```
msimberg, line 69: This should stay but as `set_value_t(Ts...)`.

libs/pika/execution/tests/include/algorithm_test_utils.hpp
msimberg: I'm only marking this unresolved as a reminder to write down which assertion failed. No further changes needed in this PR (for this comment).

msimberg:
Make sure you also check this failure: https://app.circleci.com/pipelines/github/pika-org/pika/2121/workflows/ea660918-a3d9-4f48-867d-b635e7be6915/jobs/21088/steps?invite=true#step-102-113.

libs/pika/execution/tests/include/algorithm_test_utils.hpp
msimberg: Also, if you have not only location of the static assertion, but the full instantiation stack that'd be helpful.

aurianer: https://gist.github.com/aurianer/8ba692582b878f57b9823a1800d2ef55

aurianer: https://github.com/pika-org/pika/blob/d723db02970b895f9f0c2f8e6b03fe1b101ec464/libs/pika/execution/include/pika/execution/algorithms/detail/single_result.hpp#L74-L75

aurianer:
bors try

libs/pika/execution/tests/include/algorithm_test_utils.hpp
msimberg: Thanks. I think that's a mostly a non-issue, though it could be made to work. Basically `when_all` expects that the predecessors actually send some values. If one of the predecessors only ever sends errors or stopped, `when_all` will also only ever send errors or stopped. It may useful to have this work in some generic contexts, but I think we can safely ignore it for now. The reference implementation may handle this already.

msimberg (approved):

msimberg:
bors merge

libs/pika/execution/tests/include/algorithm_test_utils.hpp
msimberg: In fact, since the CI build with the reference implementation builds without the `set_value_t()` in `completion_signatures`, it does handle this correctly.

msimberg:
bors merge

libs/pika/execution/tests/include/algorithm_test_utils.hpp
msimberg: `completion_signatures` is only used by the reference implementation and it's able to handle no values. Our implementation does not handle it and requires an empty `Tuple<>`.

PR #315, opened 2022-07-12
Preliminary support for using the P2300 reference implementation with CUDA

msimberg:
Part 1/n to fix #302. This adds changes that in principle enable the use of `PIKA_WITH_P2300_REFERENCE_IMPLEMENTATION=ON` with CUDA. In practice it means that these changes can be correctly parsed when compiling host code but Cray's clang fails with an ICE when parsing `ranges`. Even though this is only part of the solution, as long as I'm not breaking any existing configuration I'd like to get this in separately. I will continue the hunt for a working clang configuration separately.

PR #326, opened 2022-07-18
Fix use of `tuple` and `variant` in `then_with_stream`

msimberg:
Leftover from #315. It didn't affect any builds since CUDA with the P2300 reference implementation is still untested and unsupported.


msimberg:
bors merge

msimberg:
bors merge

PR #328, opened 2022-07-18
Remove redundant configuration options for CI

aurianer:
These 2 options are already set line 24 and line 25.


msimberg (approved):
Nice catch, thanks!

PR #351, opened 2022-08-08
Reduce chance of false positives in `contiguous_index_queue` test

msimberg:
Increase the size of the range to reduce chance of false positives. This is to avoid these kind of test failures: https://cdash.cscs.ch/test/29487306.


msimberg:
bors merge

PR #355, opened 2022-08-08
Fix use of `--pika:ignore-process-mask` command line option

msimberg:
Fixes the condition used for checking if a process mask should be used or not. Also adds a default `pika.ignore_process_mask = 0` option to the runtime configuration so that adding the option manually doesn't error out complaining about a nonexistent entry in the configuration.

Thanks @aurianer for noticing this.


msimberg:
bors merge

msimberg:
bors merge

PR #359, opened 2022-08-09
Attempt to fix `program_options` visibility issue

### LINKED ISSUE #350: Potential visibility issue in `program_options`
When using DLA-Future in cp2k with default pika configuration options (i.e. hidden symbols by default with GCC) cp2k does not find a `program_options` symbol. Disabling hidden visibility "fixes" the problem. The symbol that is not found is one of the `parse` functions here: https://github.com/pika-org/pika/blob/30582a69edb17ea18d5862f2c8bfe367fd93c0cc/libs/pika/program_options/include/pika/program_options/value_semantic.hpp#L94-L96. High priority because affects cp2k integration.

msimberg:
Fixes #350. It's slightly unclear to me why this is required as I assumed that a visibility attribute on the class itself would be enough, but it may be that the ruls for template specializations are different or something in that direction...

### DIFF
--- a/libs/pika/program_options/include/pika/program_options/value_semantic.hpp
+++ b/libs/pika/program_options/include/pika/program_options/value_semantic.hpp
@@ -91,7 +91,7 @@ namespace pika { namespace program_options {
       : public value_semantic
     {
     private:    // base overrides
-        void parse(std::any& value_store,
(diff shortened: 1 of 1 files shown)


msimberg:
bors try

msimberg:
bors merge

PR #360, opened 2022-08-09
Increase algorithms tests timeout in macOS workflow

msimberg:
Follow-up to #352. Also increase the timeout for the algorithms tests. `partial_sort_copy` (like `partial_sort_copy_range`) takes long to complete, most likely just due to insufficient time.


msimberg:
bors merge

Issue #367, opened 2022-08-10
Remove `PIKA_SANITY` macros


msimberg:
We're not currently using them. Unless we foresee a good use of them we can probably safely remove them and reintroduce them if needed in the future. Can be done as part of #16 for the `testing` module.

PR #368, opened 2022-08-10
Improve `report_errors` output

msimberg:
- Always print a message, even if no tests failed. This makes it easier to see that the test actually ran successfully. It also prints how many tests were run.
- Print a message and fail if `report_errors` is called but no tests were actually run.

Example outputs below. Success:
```
All tests passed. Ran 0 sanity checks and 150121 tests.
```
Failure:
```
Tests failed. 0/0 sanity checks and 1/13 tests failed.
```
No tests run:
```
No tests run. Did you forget to add tests?
```
Ideas to improve these messages are warmly welcome.

Related: #367. Not removing them yet, but I think we can quite safely remove the sanity checks (since they're unused).

Made a few tests that were triggering the "No tests run" case into compile-only tests. Added a few `PIKA_TEST(true)`s to others.

### DIFF
--- a/testing/src/testing.cpp
+++ b/testing/src/testing.cpp
@@ -108,7 +108,13 @@ namespace pika { namespace util {
             detail::global_fixture.get_failures(counter_sanity);
         auto test_failures = detail::global_fixture.get_failures(counter_test);
 
-        if (sanity_failures == 0 && test_failures == 0)
+        if (sanity_tests == 0 && test_tests == 0)
+        {
+            pika::util::ios_flags_saver ifs(stream);
+            stream << "No tests run. Did you forget to add tests?" << std::endl;
+            return 1;
+        }
+        else if (sanity_failures == 0 && test_failures == 0)
         {
             pika::util::ios_flags_saver ifs(stream);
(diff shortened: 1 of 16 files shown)


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

testing/src/testing.cpp
```diff
@@ -23,10 +23,28 @@
...
+            case counter_sanity:
+                ++sanity_tests_;
+                return;
+            case counter_test:
+                ++test_tests_;
+                return;
+            default:
+                break;
+            }
+            PIKA_ASSERT(false);
```
aurianer, line 44: can the `PIKA_ASSERT(false)` be in the `default:`?

aurianer (approved):
LGTM, thanks a lot!

testing/src/testing.cpp
msimberg: Yes, I think so.

aurianer:
bors merge

testing/src/testing.cpp
msimberg: I did not update this yet...

PR #369, opened 2022-08-10
Increase maximum ccache cache size on macOS workflow

msimberg:
In an attempt to make better use of the cache. The default limit is 500 MB and the cache is typically around 400 MB from one run. We may need to go back to the non-ccache, split setup for macOS if this doesn't work so well.


msimberg:
bors merge

msimberg:
CircleCI seems to be having minor issues now. Will retry the build later.

msimberg:
bors merge

msimberg:
bors merge

PR #375, opened 2022-08-12
Fix `config_defines.hpp` timestamp being updated on every CMake reconfiguration

msimberg:
The file was being potentially written for every module, and since some of the modules added options (that get written into the `config_defines.hpp` file) the file would be rewritten every CMake reconfiguration, but the contents would stay unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -974,8 +974,7 @@ pika_option(
   (default: OFF)"
   OFF
   ADVANCED
-  CATEGORY "Modules"
-  MODULE TOPOLOGY
+  CATEGORY "Debugging"
 )
 
 if(PIKA_WITH_ADDITIONAL_HWLOC_TESTING)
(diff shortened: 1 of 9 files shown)


msimberg:
bors try

CMakeLists.txt
```diff
@@ -523,6 +523,13 @@ pika_option(
   ADVANCED
 )
 
+pika_option(
+  PIKA_WITH_SWAP_CONTEXT_EMULATION BOOL
+  "Emulate SwapContext API for coroutines (Windows only, default: OFF)" OFF
+  CATEGORY "Thread Manager"
+  ADVANCED
```
aurianer, line 530: is omitting `MODULE COROUTINE` intentional?

aurianer (approved):
LGTM thanks!

CMakeLists.txt
aurianer: Sounds good to remove it!

msimberg: I removed the `MODULE` option completely from `pika_option` along with the `Modules` category. Does this look better now?

msimberg:
bors try

aurianer:
bors merge

Issue #353, opened 2022-08-08
Follow up on Codacy coverage setup


msimberg:
- [x] Check if enabled branches in workflow need to be changed
- [x] Check if it makes sense to include in bors status checks
- [x] Check if reports look correct (this should now be better with #381)

msimberg:
One thing that definitely needs to be fixed is to exclude tests/include only library files into the coverage report. Currently tests and examples are included in the report.

msimberg:
I'm closing this for now as the setup seems to be "good enough".

PR #381, opened 2022-08-17
Keep libpika source files for coverage reports, not tests or examples

msimberg:
Part of #353. From local testing this seems to include only the "right" files, i.e. no tests or examples. Let's see what codacy says once the reports are uploaded (this will only run once on `main`).


aurianer (approved):
LGTM thanks!

aurianer:
bors merge

PR #387, opened 2022-08-18
Report test errors on global test fixture destruction instead of requiring an explicit call to `report_errors`

msimberg:
This fixes a usability problem with the testing framework. It was previously required to call `report_errors` and to return the value from `report_errors` in `main`. This means that if `report_errors` was not called or called in the wrong place failing tests would actually report themselves as successful. I noticed at least one test (`barrier_cpp20_test`) that was missing `report_errors` completely.

This changes the test fixture to report errors in the destructor and if there are test failures to exit with failure. This way it's impossible to forget to check for failed tests.

This fixes one additional test problem, where `fail_future_2667_test` (which is an "expect compilation failure" test) failed to compile because of other reasons, in addition to the intended failure. Now it only fails because of the intended failure.


msimberg:
bors try

libs/pika/algorithms/tests/performance/benchmark_inplace_merge.cpp
```diff
@@ -236,5 +236,5 @@ int main(int argc, char* argv[])
     PIKA_TEST_EQ_MSG(pika::init(pika_main, argc, argv, init_args), 0,
         "pika main exited with non-zero status");
 
-    return pika::util::report_errors();
+    return 0;
```
aurianer, line 239: what happened in case we call `report_errors` twice?

aurianer (approved):
LGTM thanks!

msimberg:
> LGTM thanks! Some failure on CI though

Yep, this found even more tests that were not actually doing any `PIKA_TEST*` :/

libs/pika/algorithms/tests/performance/benchmark_inplace_merge.cpp
msimberg: The test summary will be printed twice but not much else should happen. Failures will still be reported and importantly the exit code will still be non-zero if tests fail. We have exactly one test where we still call `report_errors` explicitly and I'm not quite sure yet if we can do without it there. _If_ we get rid of all current explicit calls to `report_errors` then we could just hide `report_errors` inside the source file. I'll look into that separately.

msimberg:
All right, this turned out to be bigger than I expected, but I think it's converging towards something useful.

I've separated the performance testing functionality out of the testing library into a separate library, so that performance tests aren't expected to have `PIKA_TEST*` somewhere in them. Now, by default:
- unit and regression tests link to the testing library and are expected to have at least one test in them
- performance tests link to the performance testing library
- examples (as tests) don't link to any of the above

I've probably still missed a few spots, so let's see what CI says.

bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

aurianer (approved):
LGTM, thanks a lot for this!

aurianer:
bors merge

msimberg:
I think `barrier_cpp20` is another one for #191. It was previously not reporting errors so we never noticed it was failing. One failure is here: https://cdash.cscs.ch/test/32940023. I'll try merging once more, but I suspect that the test will need to be dealt with somehow before this can go in...

bors merge

msimberg:
bors merge

I've attempted to fix the `barrier_cpp20` test. It looks like the test itself was buggy (racy).

msimberg:
bors merge

msimberg:
bors merge

Issue #333, opened 2022-07-21
Don't require using hipcc as the host compiler for HIP


msimberg:
We should see if HIP's CMake setup is now usable without having `hipcc` compile everything.

msimberg:
Requires at least HIP 5.2.0 in spack because of this issue: https://github.com/spack/spack/issues/29527.

msimberg:
Revisit this once https://github.com/spack/spack/pull/31667 is merged and we can test with HIP 5.2.0.

msimberg:
HIP 5.2.0 is now available in spack (and installed on LUMI) so this is (maybe) possible to do now.

msimberg:
Fixed by #402.

PR #390, opened 2022-08-19
Change builtin detection to use `__has_builtin` instead of using configuration tests

msimberg:
The configuration tests for the various builtins are difficult to maintain. We hit another case when building CP2K where pika was built with `hipcc` and detected that it had `__make_integer_seq` (since it's clang) and when compiling CP2K with gcc it expected to have `__make_integer_seq` but failed because it has `__integer_pack` instead.

See https://godbolt.org/z/4ej7hnbTK for an example.

Since it's unreasonable to expect everyone to really use the same compiler (it's recommended but not a must) I've changed the detection to be at the use site of the builtins using `#if defined(__has_builtin) && __has_builtin(__integer_pack)`. This is in my opinion at least as simple where they are used and avoids detecting them wrongly if the consumer's compiler is different from the one used to compiler pika itself.

I've bumped the priority of #333 as well since it's an alternative "solution" to the same problem (however, only partial).


msimberg:
The variant implementation uses exactly this method as well, by the way.

msimberg:
bors try

msimberg:
bors try

aurianer (approved):
LGTM thanks!

aurianer:
bors merge

aurianer:
bors merge-

aurianer:
sorry my page was not up to date and didn't show the conflict on your PR

msimberg:
> sorry my page was not up to date and didn't show the conflict on your PR

No need to apologize, all good.

bors merge

msimberg:
bors cancel

msimberg:
bors merge

PR #391, opened 2022-08-19
Replace hipblas by rocblas

aurianer:
Necessary as first step for https://github.com/eth-cscs/DLA-Future/issues/577 to avoid redefinition of the macros.

Will wait for #67 to be merged.

### DIFF
--- a/libs/pika/async_cuda/src/cublas_exception.cpp
+++ b/libs/pika/async_cuda/src/cublas_exception.cpp
@@ -17,13 +17,34 @@ namespace pika::cuda::experimental {
         {
             switch (error)
             {
+#ifdef PIKA_HAVE_CUDA
             case CUBLAS_STATUS_SUCCESS:
                 return "CUBLAS_STATUS_SUCCESS";
             case CUBLAS_STATUS_INVALID_VALUE:
                 return "CUBLAS_STATUS_INVALID_VALUE";
             case CUBLAS_STATUS_INTERNAL_ERROR:
                 return "CUBLAS_STATUS_INTERNAL_ERROR";
-#ifdef PIKA_HAVE_HIP
+            case CUBLAS_STATUS_ALLOC_FAILED:
+                return "CUBLAS_STATUS_ALLOC_FAILED";
+            case CUBLAS_STATUS_ARCH_MISMATCH:
+                return "CUBLAS_STATUS_ARCH_MISMATCH";
+            case CUBLAS_STATUS_EXECUTION_FAILED:
+                return "CUBLAS_STATUS_EXECUTION_FAILED";
+            case CUBLAS_STATUS_LICENSE_ERROR:
+                return "CUBLAS_STATUS_LICENSE_ERROR";
+            case CUBLAS_STATUS_MAPPING_ERROR:
+                return "CUBLAS_STATUS_MAPPING_ERROR";
+            case CUBLAS_STATUS_NOT_INITIALIZED:
+                return "CUBLAS_STATUS_NOT_INITIALIZED";
+            case CUBLAS_STATUS_NOT_SUPPORTED:
+                return "CUBLAS_STATUS_NOT_SUPPORTED";
+#else
+            case rocblas_status_success:
+                return "rocblas_status_success";
+            case rocblas_status_invalid_value:
+                return "rocblas_status_invalid_value";
+            case rocblas_status_internal_error:
+                return "rocblas_status_internal_error";
             case rocblas_status_check_numerics_fail:
                 return "rocblas_status_check_numerics_fail";
             case rocblas_status_continue:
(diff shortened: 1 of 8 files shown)


aurianer:
bors try

aurianer:
bors try

libs/pika/async_cuda/tests/unit/then_with_stream.cu
```diff
@@ -113,11 +119,13 @@ struct dummy
         return x + 1;
     }
 
+#if !defined(PIKA_HAVE_HIP)
     int operator()(cusolverDnHandle_t, double x) const
     {
         ++cusolver_double_calls;
         return x + 1;
     }
+#endif
```
msimberg, line 128: With this change the expected call counts will need to be updated for HIP since the `cusolverDnHandle_t` overloads will never be called.

aurianer: Thanks, yes I went too fast

libs/pika/async_cuda/tests/unit/then_with_stream.cu
```diff
@@ -440,15 +448,22 @@ int pika_main()
...
+#if !defined(PIKA_HAVE_HIP)
         PIKA_TEST_EQ(dummy::cusolver_int_calls.load(), std::size_t(0));
+#endif
+        PIKA_TEST_EQ(dummy::host_int_calls.load(), std::size_t(2));
         PIKA_TEST_EQ(dummy::host_double_calls.load(), std::size_t(0));
         PIKA_TEST_EQ(dummy::stream_double_calls.load(), std::size_t(1));
         PIKA_TEST_EQ(dummy::cublas_double_calls.load(), std::size_t(0));
+#if !defined(PIKA_HAVE_HIP)
         PIKA_TEST_EQ(dummy::cusolver_double_calls.load(), std::size_t(1));
+#endif
```
msimberg, line 466: As already mentioned, let's wait for #67 to go in before this. It's not sufficient to just remove `cusolver_*_calls` checks. The counts from those will be distributed among the `host_*_calls`, `stream_*_calls`, `cublas_*_calls` variables.

msimberg (approved):
The `then_with_stream` test still needs updating, but this looks good to me otherwise.

libs/pika/async_cuda/tests/unit/then_with_stream.cu
aurianer: I updated the counts, should be good now

aurianer:
bors try

libs/pika/async_cuda/src/cublas_exception.cpp
```diff
@@ -19,36 +19,49 @@ namespace pika::cuda::experimental {
...
-                return "CUBLAS_STATUS_ARCH_MISMATCH";
-            case CUBLAS_STATUS_MAPPING_ERROR:
-                return "CUBLAS_STATUS_MAPPING_ERROR";
-            case CUBLAS_STATUS_EXECUTION_FAILED:
-                return "CUBLAS_STATUS_EXECUTION_FAILED";
             case CUBLAS_STATUS_INTERNAL_ERROR:
                 return "CUBLAS_STATUS_INTERNAL_ERROR";
-            case CUBLAS_STATUS_NOT_SUPPORTED:
-                return "CUBLAS_STATUS_NOT_SUPPORTED";
 #ifdef PIKA_HAVE_HIP
```
msimberg, line 26: Should we move the "common" error codes separately into the `PIKA_HAVE_HIP` and `PIKA_HAVE_CUDA` sections so that the returned strings are correct? I.e. so that we can have `rocblas_status_invalid_value` printed with HIP enabled instead of `CUBLAS_STATUS_INVALID_VALUE` etc.?

msimberg: Note, this comment does not invalidate my previous approval. Can be changed separately.

aurianer: yes definitely!

aurianer: I updated it in this PR

msimberg (approved):
Thanks @aurianer.

This is ready to be merged. However, as already discussed separately I'd like to time merging this close to merging the corresponding changes to DLA-Future. Let's wait for that to be close to ready before actually pressing the merge button.

msimberg:
I think with 0.8.0 just released we don't actually need to wait any longer to merge this. We have all the thus-far known required fixes for CP2K in 0.8.0 so CP2K can stop using pika main and stick to 0.8.0.

@aurianer do you have further changes you'd like to have in here before merging?

msimberg:
bors merge

msimberg:
bors merge

PR #393, opened 2022-08-22
Add a CircleCI workflow for arm64

### LINKED ISSUE #332: Add arm64 CI
Possibly on CircleCI (https://circleci.com/docs/using-arm) or on ault (through the same setup as for #67).

aurianer:
Attempt to fix #332.
Open usage with M1 (untested so possibly broken).


msimberg:
The build says "Resource class "arm.xlarge" is not a valid resource class. The default resource class will be used." Are you sure this is actually running on arm?

CMakeLists.txt
```diff
@@ -1500,16 +1500,13 @@ else()
     OR "${__target_arch}" STREQUAL "armv5"
     OR "${__target_arch}" STREQUAL "armv6"
     OR "${__target_arch}" STREQUAL "armv7"
+    OR "${__target_arch}" STREQUAL "armv64"
```
msimberg, line 1503:
```suggestion
    OR "${__target_arch}" STREQUAL "arm64"
```

aurianer: Thanks!

aurianer:
> The build says "Resource class "arm.xlarge" is not a valid resource class. The default resource class will be used." Are you sure this is actually running on arm?

I added a print of the architecture so let's see.. in case it's arm it might be that this resource is not available in the free plan

.circleci/config.yml
```diff
@@ -582,6 +582,55 @@ jobs:
           working_directory: /pika/build
           when: always
 
+  arm64_build:
+    <<: *docker_default
```
msimberg, line 586: You most likely don't want this. The image is not built for arm.

.circleci/config.yml
```diff
@@ -582,6 +582,63 @@ jobs:
...
+    machine:
+      image: ubuntu-2004:current
+    resource_class: arm.large
+    steps:
+      - checkout
+      - run:
+          name: Building Arm64
+          command: |
+            # Check architecture = arm
+            uname -m
```
msimberg, line 595: This can probably be removed now?

.circleci/config.yml
```diff
@@ -582,6 +582,63 @@ jobs:
...
+            # Install CMake
+            wget -q https://cmake.org/files/v3.22/cmake-3.22.6-linux-aarch64.tar.gz \
+              -O cmake.tar.gz && echo \
+              "79be85d3e76565faacd60695cee11d030f7e7dd8691476144fa25eb93dbd0397  cmake.tar.gz" \
+              | sha256sum --check --status && \
+              sudo tar --strip-components=1 -C /usr/local -xf cmake.tar.gz && \
+              rm cmake.tar.gz
+            mkdir build
+            git clone https://github.com/pika-org/pika.git
```
msimberg, line 604: This should be removed and use the `checkout` step instead.

.circleci/config.yml
```diff
@@ -582,6 +582,63 @@ jobs:
...
+                -DPIKA_WITH_STACKTRACES_STATIC_SYMBOLS=On \
+                -DPIKA_WITH_STACKTRACES_DEMANGLE_SYMBOLS=Off \
+                -DPIKA_WITH_TESTS_DEBUG_LOG=On \
+                -DPIKA_WITH_TESTS_DEBUG_LOG_DESTINATION=/build/debug-log-arm.txt \
+                -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=On
+            ninja -j4 -k 0 pika
+      - persist_to_workspace:
+          root: /home/circleci/project
+          paths:
+            - ./build
```
msimberg, line 640: I don't think this is needed since we're not splitting this into multiple jobs (right now at least...).

.circleci/config.yml
```diff
@@ -582,6 +582,63 @@ jobs:
...
+                -DPIKA_WITH_COMPILER_WARNINGS=On \
+                -DPIKA_WITH_COMPILER_WARNINGS_AS_ERRORS=On \
+                -DPIKA_WITH_DEPRECATION_WARNINGS=On \
+                -DPIKA_WITH_THREAD_LOCAL_STORAGE=On \
+                -DPIKA_WITH_STACKTRACES_STATIC_SYMBOLS=On \
+                -DPIKA_WITH_STACKTRACES_DEMANGLE_SYMBOLS=Off \
+                -DPIKA_WITH_TESTS_DEBUG_LOG=On \
+                -DPIKA_WITH_TESTS_DEBUG_LOG_DESTINATION=/build/debug-log-arm.txt \
+                -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=On
+            ninja -j4 -k 0 pika
```
msimberg, line 636: Could you please split this up into "Install dependencies", "Configure", "Build", and a new "Test" step? The "Test" step is optional to get working for this PR. I'd just like to see how good/bad the test runs actually are. If it works right away let's keep it in this PR. If there's _anything_ that looks off we delay it for a separate PR.

TL;DR. Splitting into multiple steps: required for this PR. Building and running tests: optional.

.circleci/config.yml
```diff
@@ -654,3 +711,4 @@ workflows:
             - tests.regressions
             - tests.performance
             - tests.headers
+      - arm64_build
```
msimberg, line 714: I don't know if this works, but could this be another "top-level workflow" on the same level as `build-and-test`? Perhaps with the name `build-and-test-arm64` (or just `build-arm64` if we don't get the testing part yet in this PR).

msimberg (changes requested):
Could you also add the arm64 step(s) to `bors.toml`?

Thanks for doing this, this is already very promising!

.circleci/config.yml
msimberg: Ok, no worries. We can leave it.

.circleci/config.yml
msimberg: What exactly do you mean by it doesn't work? The directory is not visible after cloning? If it doesn't work this will have to be changed to actually check out the branch that triggers the build.

.circleci/config.yml
msimberg: We can add it again if and when we do run the tests in separate steps.

.circleci/config.yml
```diff
@@ -654,3 +724,4 @@ workflows:
...
```
msimberg, line 729:
```suggestion
      - arm64_build:
          <<: *branch_filter
```

.circleci/config.yml
```diff
@@ -582,6 +587,71 @@ jobs:
...
+                -DPIKA_WITH_STACKTRACES_STATIC_SYMBOLS=On \
+                -DPIKA_WITH_STACKTRACES_DEMANGLE_SYMBOLS=Off \
+                -DPIKA_WITH_TESTS_DEBUG_LOG=On \
+                -DPIKA_WITH_TESTS_DEBUG_LOG_DESTINATION=/build/debug-log-arm.txt \
+                -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=On
+          working_directory: /home/circleci/project/build
+      - run:
+          name: Build pika Arm64
+          command: |
+            ninja -j4 -k 0 pika
```
msimberg, line 652: Let's try if this can build without running out of memory:
```suggestion
            ninja -k 0 pika
```

msimberg (changes requested):
Looks good. Now it's only the `bors.toml` change and `branch_filter` missing.

.circleci/config.yml
```diff
@@ -582,6 +587,67 @@ jobs:
...
+            # Install pika dependencies
+            sudo apt install clang ninja-build libboost-regex-dev \
+              libboost-context-dev libhwloc-dev mpi-default-dev -y
+            mkdir build
+          working_directory: /home/circleci/project
+      - run:
+          name: Configure pika Arm64
+          command: |
+            cmake \
+                /home/circleci/project/pika \
```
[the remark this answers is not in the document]
msimberg: This wasn't supposed to here anymore... Please ignore.

msimberg (approved):

PR #395, opened 2022-08-22
Re-enable release builds on Jenkins

### LINKED ISSUE #382: Re-enable release builds on Jenkins CI
I think with ccache etc. the cost of this would not be terribly high (but of course double of what we have right now). It would be good to check release builds as well...

msimberg:
Half-depends on #67 (i.e. should go in afterwards).

Fixes #382.

This may end up increasing the failure rate of #191 to something that is no longer sustainable. With this we also have to be a lot more mindful not to spam `bors try/merge` since the used node hours will at least double. Draft because of the above reasons.


msimberg:
bors try

msimberg:
There are for sure some interesting failures here that should be looked into... I'll use this PR as a bit of a testbed for those failures.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

libs/pika/algorithms/tests/unit/algorithms/CMakeLists.txt
```diff
@@ -147,10 +147,18 @@ set(tests
...
 
+# Cray's clang compiler produces segfaults in release mode on the unique test.
+# This seems to be a compiler bug. We disable the test for all non-Debug builds
+# with clang.
+if(NOT (NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug" AND "${CMAKE_COMPILER_ID}"
+                                                       STREQUAL "Clang")
+)
+  list(APPEND tests unique)
+endif()
+
```
msimberg, line 161: This problem is listed under `unique` on https://github.com/pika-org/pika/issues/191.

msimberg:
This should now be ready. I've had to disable one test in release mode with clang due to what seems to be a bug in Cray's clang.

bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
Perhaps I'll regret this, but let's at least give it a go.

bors merge

Issue #17, opened 2022-01-18
Separate CUDA/HIP/GPU support into another repository


msimberg:
The exact requirements of this are not 100% clear. This at least needs:
- The schedulers need a generic way to register polling callbacks, or
- The polling can be done in a separate thread?
- The runtime/a thread pool needs to know if it should wait for all events to finish

PR #401, opened 2022-08-24
Use `std::function` for startup/shutdown functions and run shutdown functions at shutdown

msimberg:
- Uses `std::function` instead of pika's `unique_function` for startup/shutdown functions
- Actually calls shutdown functions at shutdown
- Doesn't clear startup/shutdown functions registered before the runtime starts so that they are registered again if the runtime is restarted

This is preliminary work and will be followed up by better tests once I understand a bit better the requirements here. I'm experimenting with `pika::register_{startup,shutdown}_function` as a way to initialize/finalize a separate polling thread for CUDA polling (see #17 and #83).

### DIFF
--- a/libs/pika/runtime/include/pika/runtime/runtime.hpp
+++ b/libs/pika/runtime/include/pika/runtime/runtime.hpp
@@ -449,6 +449,7 @@ namespace pika {
         void wait_finalize();
 
         void call_startup_functions(bool pre_startup);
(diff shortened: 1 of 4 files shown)


msimberg:
bors try

aurianer (approved):
LGTM thanks!

aurianer:
bors merge

PR #409, opened 2022-08-26
Re-split macOS workflow job into multiple steps

msimberg:
`ccache` was not as effective as hoped because:
- The cache not shared across branches
- The workflow builds tests and examples, which means that changes to libpika typically lead to most tests and examples anyway being rebuilt.

This reverts the changes to the macOS workflow from #354, but keeps #370.

PR #410, opened 2022-08-26
Disable precompiled headers on coverage workflow

msimberg:
This seems to be inhibiting ccache to cache compilations. This workflow has in principle the same problem as in #409 (i.e. it builds all tests and examples and is prone to full rebuilds), but since the non-cached builds are already a lot faster than on macOS there's less to be gained by splitting the workflow into multiple steps.


msimberg:
bors merge

PR #414, opened 2022-08-26
Reduce chance of false positives in `limiting_executor` test

msimberg:
Description about the problem added to the test itself.

Note: the problem can easily be reproduced by adding a `std::chrono::sleep_for` after this line: https://github.com/pika-org/pika/blob/72834132ec83a3d09db8808b53233fc27b9e149d/libs/pika/executors/include/pika/executors/limiting_executor.hpp#L58.


msimberg:
bors try

aurianer (approved):
LGTM thanks!

aurianer:
bors try

msimberg:
bors merge

biddisco:
Note that the limiting executor should be deprecated and we should switch to an `async_scope` type approach ideally.

PR #416, opened 2022-08-26
Fix `shared_mutex2` test

msimberg:
The failure can be reproduced (before these changes) by adding `std::this_thread::sleep_for` here: https://github.com/pika-org/pika/blob/887eb9c56097bccdde92a90ca56c73a8e1334104/libs/pika/synchronization/tests/unit/shared_mutex/shared_mutex_locking_thread.hpp#L52. The test previously only did `pika::this_thread::yield()` to wait for the spawned threads to start running but this is not always sufficient. This adds synchronization using the already available condition variable, and adds additional sleeping to allow the other thread to run (though it shouldn't run if things are working correctly).

### DIFF
--- a/libs/pika/synchronization/tests/unit/shared_mutex/shared_mutex2.cpp
+++ b/libs/pika/synchronization/tests/unit/shared_mutex/shared_mutex2.cpp
@@ -57,7 +57,16 @@ void test_only_one_upgrade_lock_permitted()
                     simultaneous_running_count, max_simultaneous_running));
         }
 
-        pika::this_thread::yield();
+        // Wait for one of the threads to signal that it is unblocked
+        {
+            std::unique_lock<mutex_type> lk(unblocked_count_mutex);
+            unblocked_condition.wait(
+                lk, [&]() { return unblocked_count >= 1u; });
+        }
+
+        // Wait a moment to see if the second thread gets the lock concurrently
+        // (it should not)
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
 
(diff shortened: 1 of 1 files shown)


.github/workflows/macos_debug.yml
```diff
@@ -66,14 +65,8 @@ jobs:
...
-            --exclude-regex \
-          "tests.unit.modules.algorithms.default_construct|\
-          tests.unit.modules.algorithms.destroy|\
-          tests.unit.modules.algorithms.foreach_executors|\
-          tests.unit.modules.algorithms.max_element|\
-          tests.unit.modules.algorithms.replace_copy_if|\
-          tests.unit.modules.execution.standalone_thread_pool_executor|\
-          tests.unit.modules.resource_partitioner.used_pus"
+            --tests-regex shared_mutex2 \
+            --repeat-until-fail 10000
```
msimberg, line 69: This must be reverted before merging. I only added it to stress the test a bit more (since it typically fails on the macOS workflow).

msimberg: The build timed out because I made it do slightly too many iterations, but the test did not fail a single time: https://github.com/msimberg/pika/runs/8038639146?check_suite_focus=true.

msimberg: I've removed this temporary change.

msimberg:
bors try

aurianer (approved):
LGTM thanks!

aurianer:
bors merge

msimberg:
Another one for #191: https://github.com/pika-org/pika/runs/8070517198?check_suite_focus=true#step:6:203.

bors merge

PR #421, opened 2022-08-30
Add MPI streams

biddisco:
This work is not completely finished - I can still get deadlocks with the async storage test, but I would like to start the process of merging this, so I can test it more and use dlaf as a more advanced testing framework

```
Support throttling of mpi traffic using streams
    
Throttling using a single counter causes deadlocks as a
receive might be blocked by a send, or vice versa.
 ```
+ Some flyby cmake and cleanup fixes

### DIFF
--- a/cmake/pika_add_test.cmake
+++ b/cmake/pika_add_test.cmake
@@ -64,7 +64,7 @@ function(pika_add_test category name)
     set(args ${args} "--pika:bind=none")
   endif()
 
-  set(args "${${name}_UNPARSED_ARGUMENTS}" ${args})
+  set(args "${${name}_ARGS}" "${${name}_UNPARSED_ARGUMENTS}" ${args})
 
   set(_script_location ${PROJECT_BINARY_DIR})
 
--- a/cmake/pika_add_test.cmake
+++ b/cmake/pika_add_test.cmake
@@ -71,8 +71,10 @@ function(pika_add_test category name)
   set(cmd ${_exe})
 
   if(${name}_RUNWRAPPER)
+    set(_preflags_list_ ${MPIEXEC_PREFLAGS})
+    separate_arguments(_preflags_list_)
     list(PREPEND cmd "${MPIEXEC_EXECUTABLE}" "${MPIEXEC_NUMPROC_FLAG}"
-         "${${name}_LOCALITIES}"
+         "${${name}_LOCALITIES}" ${_preflags_list_}
     )
   endif()
 
--- a/libs/pika/async_mpi/CMakeLists.txt
+++ b/libs/pika/async_mpi/CMakeLists.txt
@@ -14,7 +14,7 @@ endif()
 # Default location is $PIKA_ROOT/libs/mpi/include
 set(async_mpi_headers
     pika/async_mpi/mpi_exception.hpp pika/async_mpi/mpi_polling.hpp
-    pika/async_mpi/transform_mpi.hpp
+    pika/async_mpi/mpi_throttling.hpp pika/async_mpi/transform_mpi.hpp
 )
 
 # Default location is $PIKA_ROOT/libs/mpi/src
--- a/libs/pika/async_mpi/CMakeLists.txt
+++ b/libs/pika/async_mpi/CMakeLists.txt
@@ -14,7 +14,7 @@ endif()
 # Default location is $PIKA_ROOT/libs/mpi/include
 set(async_mpi_headers
     pika/async_mpi/mpi_exception.hpp pika/async_mpi/mpi_polling.hpp
-    pika/async_mpi/mpi_throttling.hpp pika/async_mpi/transform_mpi.hpp
+    pika/async_mpi/transform_mpi.hpp
 )
 
 # Default location is $PIKA_ROOT/libs/mpi/src
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -21,34 +21,33 @@
 #include <cstddef>
 #include <cstdint>
 #include <iosfwd>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
 
 namespace pika::mpi::experimental {
 
-    enum stream_index : std::uint32_t
+    enum class stream_type : std::uint32_t
     {
-        default_stream = 0,
-        send_stream,
-        receive_stream,
-        collective_stream,
-        user_stream,
+        automatic = 0,
+        send,
+        receive,
+        collective,
+        user,
     };
 
     namespace detail {
         // -----------------------------------------------------------------
         // by convention the title is 7 chars (for alignment)
         using print_on = pika::debug::detail::enable_print<false>;
-        static constexpr print_on mpi_debug("MPI_FUT");
-        struct mpi_queue;
+        static print_on mpi_debug("MPI_FUT");
 
         using request_callback_function_type =
             pika::util::detail::unique_function<void(int)>;
 
         PIKA_EXPORT void add_request_callback(
-            request_callback_function_type&& f, MPI_Request req,
-            stream_index queue);
+            request_callback_function_type&&, MPI_Request, stream_type);
         PIKA_EXPORT void register_polling(pika::threads::thread_pool_base&);
         PIKA_EXPORT void unregister_polling(pika::threads::thread_pool_base&);
 
@@ -63,29 +62,7 @@ namespace pika::mpi::experimental {
         // -----------------------------------------------------------------
         /// Called by the mpi senders/executors to initiate throttling
         /// when necessary
-        PIKA_EXPORT void wait_for_throttling_snd(mpi_queue& queue);
-        PIKA_EXPORT void wait_for_throttling_snd(stream_index stream);
-
-        // -----------------------------------------------------------------
-        /// Set the number of messages above which throttling will be applied
-        /// when invoking an MPI function. If the number of messages in flight
-        /// exceeds the amount specified, then any thread attempting to invoke
-        /// and MPI function that generates an MPI_Request will be suspended.
-        /// This should be used with great caution as setting it too low can
-        /// cause deadlocks. The default value is size_t(-1) - i.e. unlimited
-        /// The value can be set using an environment variable as follows
-        /// PIKA_MPI_MSG_THROTTLE=64
-        /// but user code setting it will override any default or env value
-        /// This function returns the previous throttling threshold value
-        PIKA_EXPORT std::uint32_t set_max_requests_in_flight(
-            std::uint32_t, stream_index = stream_index(-1));
-
-        /// Query the current value of the throttling threshold
-        PIKA_EXPORT std::uint32_t get_max_requests_in_flight(stream_index);
-
-        // -----------------------------------------------------------------
-        /// returns the number of mpi requests currently outstanding
-        PIKA_EXPORT std::uint32_t get_num_requests_in_flight();
+        PIKA_EXPORT void wait_for_throttling(stream_type);
 
         // -----------------------------------------------------------------
         // set an error handler for communicators that will be called
@@ -97,6 +74,32 @@ namespace pika::mpi::experimental {
         // Checks for completed MPI_Requests and sets ready state in waiting receivers
         PIKA_EXPORT pika::threads::policies::detail::polling_status poll();
     }    // namespace detail
+
+    // -----------------------------------------------------------------
+    /// Set the number of messages above which throttling will be applied
+    /// when invoking an MPI function. If the number of messages in flight
+    /// exceeds the amount specified, then any thread attempting to invoke
+    /// and MPI function that generates an MPI_Request will be suspended.
+    /// This should be used with great caution as setting it too low can
+    /// cause deadlocks. The default value is size_t(-1) - i.e. unlimited
+    /// The value can be set using an environment variable as follows
+    /// PIKA_MPI_MSG_THROTTLE=64
+    /// but user code setting it will override any default or env value
+    /// This function returns the previous throttling threshold value
+    /// If the optional stream param is not specified, then all streams
+    /// are set with the same limit
+    PIKA_EXPORT std::uint32_t set_max_requests_in_flight(
+        std::uint32_t, std::optional<stream_type> = std::nullopt);
+
+    /// Query the current value of the throttling threshold for the stream
+    /// if the optional stream param is not specified, then the
+    /// automatic/default stream value is returned
+    PIKA_EXPORT std::uint32_t get_max_requests_in_flight(
+        std::optional<stream_type> = std::nullopt);
+
+    // -----------------------------------------------------------------
+    /// returns the number of mpi requests currently outstanding
+    PIKA_EXPORT std::uint32_t get_num_requests_in_flight();
 
     // initialize the pika::mpi background request handler
     // All ranks should call this function,
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -26,17 +26,29 @@
 #include <vector>
 
 namespace pika::mpi::experimental {
+
+    enum stream_index : std::uint32_t
+    {
+        default_stream = 0,
+        send_stream,
+        receive_stream,
+        collective_stream,
+        user_stream,
+    };
+
     namespace detail {
         // -----------------------------------------------------------------
         // by convention the title is 7 chars (for alignment)
         using print_on = pika::debug::detail::enable_print<false>;
         static constexpr print_on mpi_debug("MPI_FUT");
+        struct mpi_queue;
 
         using request_callback_function_type =
             pika::util::detail::unique_function<void(int)>;
 
         PIKA_EXPORT void add_request_callback(
-            request_callback_function_type&& f, MPI_Request req);
+            request_callback_function_type&& f, MPI_Request req,
+            stream_index queue);
         PIKA_EXPORT void register_polling(pika::threads::thread_pool_base&);
         PIKA_EXPORT void unregister_polling(pika::threads::thread_pool_base&);
 
@@ -51,7 +63,8 @@ namespace pika::mpi::experimental {
         // -----------------------------------------------------------------
         /// Called by the mpi senders/executors to initiate throttling
         /// when necessary
-        PIKA_EXPORT void wait_for_throttling();
+        PIKA_EXPORT void wait_for_throttling_snd(mpi_queue& queue);
+        PIKA_EXPORT void wait_for_throttling_snd(stream_index stream);
 
         // -----------------------------------------------------------------
         /// Set the number of messages above which throttling will be applied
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp [hunk 1]
```diff
@@ -26,17 +26,29 @@
...
 namespace pika::mpi::experimental {
+
+    enum stream_index : std::uint32_t
+    {
+        default_stream = 0,
+        send_stream,
+        receive_stream,
+        collective_stream,
+        user_stream,
+    };
```
msimberg, line 37: Maybe something like this?
```suggestion
    enum class stream_type : std::uint32_t // does the user need to know that it's used as an index? it's the type of stream that they're interested in, right?
    {
        default_ = 0, // possibly different name here to avoid the underscore...
        send,
        receive,
        collective,
        user,
    };
```

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -26,17 +26,29 @@
...
+
     namespace detail {
         // -----------------------------------------------------------------
         // by convention the title is 7 chars (for alignment)
         using print_on = pika::debug::detail::enable_print<false>;
         static constexpr print_on mpi_debug("MPI_FUT");
+        struct mpi_queue;
```
msimberg, line 44: Is it a stream or a queue? I don't mind which one, but it should be used consistently.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -26,17 +26,29 @@
...
 
         using request_callback_function_type =
             pika::util::detail::unique_function<void(int)>;
 
         PIKA_EXPORT void add_request_callback(
-            request_callback_function_type&& f, MPI_Request req);
+            request_callback_function_type&& f, MPI_Request req,
+            stream_index queue);
```
msimberg, line 51: Same here.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -61,17 +74,18 @@ namespace pika::mpi::experimental {
         /// This should be used with great caution as setting it too low can
         /// cause deadlocks. The default value is size_t(-1) - i.e. unlimited
         /// The value can be set using an environment variable as follows
-        /// PIKA_MPI_MSG_THROTTLE=512
+        /// PIKA_MPI_MSG_THROTTLE=64
         /// but user code setting it will override any default or env value
         /// This function returns the previous throttling threshold value
-        PIKA_EXPORT size_t set_max_requests_in_flight(size_t);
+        PIKA_EXPORT std::uint32_t set_max_requests_in_flight(
+            std::uint32_t, stream_index = stream_index(-1));
```
msimberg, line 81: What does the `-1` represent? Could we use an optional instead?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -61,17 +74,18 @@ namespace pika::mpi::experimental {
...
 
         /// Query the current value of the throttling threshold
-        PIKA_EXPORT size_t get_max_requests_in_flight();
+        PIKA_EXPORT std::uint32_t get_max_requests_in_flight(stream_index);
 
         // -----------------------------------------------------------------
         /// returns the number of mpi requests currently outstanding
-        PIKA_EXPORT size_t get_num_requests_in_flight();
+        PIKA_EXPORT std::uint32_t get_num_requests_in_flight();
```
msimberg, line 88: Why `uint32_t` instead of `size_t`? Not that we'll probably ever have more than `2^32` requests in flight at the same time... but still.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -383,22 +389,49 @@ namespace pika::mpi::experimental {
...
+        template <typename Sender, typename F,
+            PIKA_CONCEPT_REQUIRES_(
+                pika::execution::experimental::is_sender_v<Sender>)>
+        friend constexpr PIKA_FORCEINLINE auto tag_fallback_invoke(
+            transform_mpi_t, Sender&& sender, F&& f,
+            mpi::experimental::stream_index s)
+        {
+            return transform_mpi_detail::transform_mpi_sender<Sender, F>{
+                PIKA_FORWARD(Sender, sender), PIKA_FORWARD(F, f), s};
+        }
```
msimberg, line 414: Isn't one overload enough?
```suggestion
        template <typename Sender, typename F,
            PIKA_CONCEPT_REQUIRES_(
                pika::execution::experimental::is_sender_v<Sender>)>
        friend constexpr PIKA_FORCEINLINE auto tag_fallback_invoke(
            transform_mpi_t, Sender&& sender, F&& f,
            stream_index s = stream_index::default_)
        {
            return transform_mpi_detail::transform_mpi_sender<Sender, F>{
                PIKA_FORWARD(Sender, sender), PIKA_FORWARD(F, f), s};
        }
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -383,22 +389,49 @@ namespace pika::mpi::experimental {
...
         }
+
+        template <typename F>
+        friend constexpr PIKA_FORCEINLINE auto tag_fallback_invoke(
+            transform_mpi_t, F&& f, mpi::experimental::stream_index s)
+        {
+            return ::pika::execution::experimental::detail::partial_algorithm<
+                transform_mpi_t, F, mpi::experimental::stream_index>{
+                PIKA_FORWARD(F, f), s};
+        }
```
msimberg, line 434: Same here, isn't one overload enough?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -26,14 +26,18 @@
 #include <vector>
 
 namespace pika::mpi::experimental {
+
+    constexpr std::uint32_t max_mpi_streams = 4;
```
msimberg, line 30: Could this instead be the last member of the `stream_index` enum so that you don't have to keep this in sync with the enum?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -26,14 +26,18 @@
...
         // -----------------------------------------------------------------
         /// Holds an MPI_Request and a callback. The callback is intended to be
         /// called when the operation tied to the request handle completes.
         struct request_callback
         {
-            MPI_Request request;
-            request_callback_function_type callback_function;
+            MPI_Request request_;
+            request_callback_function_type callback_function_;
+            stream_index index_{0};
```
msimberg, line 40:
```suggestion
            stream_index index_{stream_index::default};
```
is maybe more obvious?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -53,7 +57,23 @@ namespace pika::mpi::experimental {
...
+        struct mpi_stream
+        {
+            mutex_type throttling_mtx_;
+            pika::lcos::local::condition_variable throttling_cond_;
+            std::atomic<std::uint32_t> in_flight_{0};
+            std::uint32_t limit_{get_throttling_default()};
+            std::uint32_t index;
+        };
+
+        using mpi_cbq_tuple =
```
msimberg, line 75:
```suggestion
        using mpi_callback_queue_tuple =
```
?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -221,17 +263,20 @@ namespace pika::mpi::experimental {
 
             // Eagerly check if request already completed. If it did, call the
             // callback immediately.
+#ifdef EAGER_CHECK
```
msimberg, line 266:
```suggestion
#ifdef PIKA_MPI_EAGER_CHECK
```
? Or if this is just for debugging please remove it.

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -235,7 +235,9 @@ void test_send_recv(std::uint32_t rank, std::uint32_t nranks, std::mt19937& gen,
             void* buffer_to_recv = &local_recv_storage[memory_offset_recv];
             auto rsnd = ex::just(buffer_to_recv, options.transfer_size_B,
                             MPI_UNSIGNED_CHAR, recv_rank, tag, MPI_COMM_WORLD) |
-                mpi::transform_mpi(MPI_Irecv) | ex::then([&](int result) {
+                mpi::transform_mpi(
+                    MPI_Irecv, /*mpi::mpi_stream::*/ mpi::receive_stream) |
```
msimberg, line 239: Is the commented-part still needed?

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -258,7 +260,9 @@ void test_send_recv(std::uint32_t rank, std::uint32_t nranks, std::mt19937& gen,
             void* buffer_to_send = &local_send_storage[memory_offset_send];
             auto ssnd = ex::just(buffer_to_send, options.transfer_size_B,
                             MPI_UNSIGNED_CHAR, send_rank, tag, MPI_COMM_WORLD) |
-                mpi::transform_mpi(MPI_Isend) | ex::then([&](int result) {
+                mpi::transform_mpi(
+                    MPI_Isend, mpi::/*mpi_stream::*/ send_stream) |
```
msimberg, line 264: Same here?

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -82,11 +82,13 @@ int pika_main(pika::program_options::variables_map& vm)
 
     // if comm size < 2 this test should fail
     // it needs to run on N>2 ranks to be useful
-    PIKA_TEST_MSG(size > 1, "This test requires N>1 mpi ranks");
+    //PIKA_TEST_MSG(size > 1, "This test requires N>1 mpi ranks");
```
msimberg, line 85: Should this still be commented out?

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -188,12 +184,12 @@ int pika_main(pika::program_options::variables_map& vm)
...
-        {
-            std::cout << "Rank " << rank << " flight " << tt << " counter "
-                      << counter << std::endl;
-        }
+        //        auto tt = mpi::detail::get_num_requests_in_flight();
+        //        if (tt != 0)
+        //        {
+        //            std::cout << "Rank " << rank << " flight " << tt << " counter "
+        //                      << counter << std::endl;
+        //        }
```
msimberg, line 192: Debugging?

libs/pika/resource_partitioner/examples/oversubscribing_resource_partitioner.cpp
```diff
@@ -86,18 +86,18 @@ int pika_main(/*pika::program_options::variables_map& vm*/)
         pika::execution::thread_priority::critical);
     pika::execution::parallel_executor normal_priority_executor;
 
-    pika::execution::parallel_executor mpi_executor;
+    pika::execution::parallel_executor pool_executor;
```
msimberg, line 89: These changes probably don't belong in this PR?

libs/pika/resource_partitioner/examples/simple_resource_partitioner.cpp
```diff
@@ -71,19 +71,19 @@ int pika_main(pika::program_options::variables_map&)
         pika::execution::thread_priority::critical);
     pika::execution::parallel_executor normal_priority_executor;
 
-    pika::execution::parallel_executor mpi_executor;
+    pika::execution::parallel_executor pool_executor;
```
msimberg, line 74: These changes also probably don't belong here?

cmake/pika_add_test.cmake
```diff
@@ -64,15 +64,17 @@ function(pika_add_test category name)
...
 
-  set(args "${${name}_UNPARSED_ARGUMENTS}" ${args})
+  set(args "${${name}_ARGS}" "${${name}_UNPARSED_ARGUMENTS}" ${args})
 
   set(_script_location ${PROJECT_BINARY_DIR})
 
   set(cmd ${_exe})
 
   if(${name}_RUNWRAPPER)
+    set(_preflags_list_ ${MPIEXEC_PREFLAGS})
```
msimberg, line 74: Is `MPIEXEC_PREFLAGS` a `find_package(mpi)` thing or should we rather name it `PIKA_TEST_MPIEXEC_PREFLAGS` or something like that?

msimberg (changes requested):
Looks good in general, and we don't have to be too careful with this since DLA-Future isn't using it yet. Mostly style comments.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: I'm happy with that change - I used "index" in the name because if the user decided they needed more streams, then
`user_stream + 1`, `user_stream + 2`, etc might be needed. But I will worry about that some other time and see iif this is a viable way of doing things first.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: it's a queue of stuff that we call a stream! I'll rename the vars appropriately.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: the `-1` is the set `all` queues/streams with the same value. An optional is a good idea. I'll do that. (and add a comment to document why)

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: because in your changes, you changed one of my `size_t` into a `uint32_t` so I thought for some reason you preferred it. Perhaps that was a victim of my conflict resolution. I originally had `size_t` for them all. (though frankly 64 bits is probably overkill for these vars)

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
biddisco: grrrr. I keep forgetting to undo those comments when I test locally with one rank. will fix

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
biddisco: yup. forgot about that. Frankly I don't mind commented out debug stuff in tests as it helps people later when they try to finx some new bug they discover

libs/pika/resource_partitioner/examples/oversubscribing_resource_partitioner.cpp
biddisco: no, but they are flyby fixes with "mpi" in them so I thought why-not.

cmake/pika_add_test.cmake
biddisco: its built into `find_package(mpi)`

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
msimberg: If that's the case, my bad! I don't have any memory of changing them to `uint32_t`s... Do you have a link to what I changed? In any case, `uint32_t` or `size_t` are both fine by me.

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
msimberg: Guard it with this instead of commenting it out: https://github.com/pika-org/pika/blob/cd738d191967f7f8656954372b93d8cc6e7997fd/libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp#L235?

libs/pika/resource_partitioner/examples/oversubscribing_resource_partitioner.cpp
msimberg: Ah, right. I missed that it indeed doesn't make sense to call it `mpi_executor`. Fine, let's keep it here (same for the other one).

biddisco:
Latest patch should address all comments. Moved a couple of functions to/from detail namespace since they might be considered part of public API. Will review these functions in an ongoing basis and possibly hide them as the module matures (throttling is not a finished product yet)

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -469,6 +514,48 @@ namespace pika::mpi::experimental {
         }
     }    // namespace detail
 
+    std::uint32_t set_max_requests_in_flight(
+        std::uint32_t N, std::optional<stream_type> s)
+    {
+        if (s == std::nullopt)
```
msimberg, line 520: Optional (heh...):
```suggestion
        if (!s)
```
or
```suggestion
        if (!s.has_value())
```
I'm fine with the comparison to `nullopt` as well.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -53,7 +61,23 @@ namespace pika::mpi::experimental {
...
+        {
+            mutex_type throttling_mtx_;
+            pika::lcos::local::condition_variable throttling_cond_;
+            std::atomic<std::uint32_t> in_flight_{0};
+            std::uint32_t limit_{get_throttling_default()};
+            std::uint32_t index;
+        };
+
+        using mpi_callback_queue_tuple =
+            std::tuple<request_callback_function_type, mpi_stream*>;
```
msimberg, line 80: Seeing this again this would probably be better off in a struct, but that can be separately.

msimberg:
bors try

msimberg (approved):
Thanks @biddisco for the updates. Two more optional comments that you're free to ignore.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -12,6 +12,7 @@
 #include <pika/config.hpp>
 #include <pika/assert.hpp>
 #include <pika/async_mpi/mpi_polling.hpp>
+#include <pika/async_mpi/mpi_throttling.hpp>
```
msimberg, line 15: This header doesn't exist by the way... All functionality is in `mpi_polling.hpp`, right?

libs/pika/async_mpi/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ endif()
 # Default location is $PIKA_ROOT/libs/mpi/include
 set(async_mpi_headers
     pika/async_mpi/mpi_exception.hpp pika/async_mpi/mpi_polling.hpp
-    pika/async_mpi/transform_mpi.hpp
+    pika/async_mpi/mpi_throttling.hpp pika/async_mpi/transform_mpi.hpp
```
msimberg, line 17: Header doesn't exist.

msimberg (changes requested):

libs/pika/async_mpi/CMakeLists.txt
biddisco: oops. I started creating a new file with the throttling code in, but removed it's contents during my conflict merge and forgot to remove the header includes. Will fix.

msimberg:
bors try

msimberg:
bors merge

msimberg (approved):

msimberg:
bors merge

PR #402, opened 2022-08-24
Use `enable_language(HIP)`

aurianer:

PR #425, opened 2022-09-01
Update minimum CMake version requirement in readme

msimberg:
Because of #402.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,7 +33,7 @@ pika requires:
   * `clang <https://clang.llvm.org>`_ 11 or greater
   * MSVC is may work but it is not tested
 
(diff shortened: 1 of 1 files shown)


aurianer (approved):
LGTM thanks a lot! Is the change of `executor_parameters_timer_hooks.cpp` supposed to be here?

msimberg:
> LGTM thanks a lot! Is the change of `executor_parameters_timer_hooks.cpp` supposed to be here?

No! Another casualty of my pushing to `main`... I based this on `main` before removing the stray commit. Thanks for noticing.

aurianer:
bors merge

PR #427, opened 2022-09-01
Use `nv_diag_suppress` when available instead of `diag_suppress`

msimberg:
I skipped this earlier, but not doing the conditional use of `(nv_)diag*` ends up with annoying warnings about deprecated pragmas.


msimberg:
bors try

aurianer (approved):
LGTM thanks!

PR #428, opened 2022-09-01
Clean up threadmanager

aurianer:
Part of #16
Note: I left `pika::threads::detail` namespace and not `pika::detail` directly as this namespace is used elsewhere and it would require add a lot of `threads::` everywhere in the module. I feel like if we want to do that, it should be done in the end and accross modules.


aurianer:
bors try

msimberg (approved):
Nice, thanks!

msimberg:
bors merge

PR #434, opened 2022-09-06
Don't run CUDA continuations eagerly if stream is ready

msimberg:
The probability of deep recursion (and in turn stack overflows) is too high compared to the almost zero performance benefit from it.

I'm still trying to understand why it's triggered so frequently in DLA-Future, but I think it's still fair to revert this for 0.8.0.

### DIFF
--- a/libs/pika/async_cuda/src/cuda_event_callback.cpp
+++ b/libs/pika/async_cuda/src/cuda_event_callback.cpp
@@ -149,15 +149,6 @@ namespace pika::cuda::experimental::detail {
         void add_to_event_callback_queue(
             event_callback_function_type&& f, cudaStream_t stream)
         {
-            // Eagerly check if the stream is empty. If it is the kernel has
-            // already completed and we can call the callback immediately.
-            auto status = cudaStreamQuery(stream);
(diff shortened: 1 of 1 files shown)


msimberg:
bors merge

biddisco:
Just fyi - we had the same problem in oomph/ghex when a receive posted a send, that posted a receive that posted a send - each one in a continuation of another. The stack overflowed. We added 'depth' counter and 'broke' the chain when a certain depth was reached. This is/was not a great solution, but the easiest way of fixing tests that would otherwise have to be rewritten.

msimberg:
> Just fyi - we had the same problem in oomph/ghex when a receive posted a send, that posted a receive that posted a send - each one in a continuation of another. The stack overflowed. We added 'depth' counter and 'broke' the chain when a certain depth was reached. This is/was not a great solution, but the easiest way of fixing tests that would otherwise have to be rewritten.

That's good to know, and I expect we may eventually hit a one in a thousand/million case of this happening with `transform_mpi` as well. With CUDA the performance difference was so marginal that it was an easy enough decision to just not do it. With MPI the tradeoff might be different.

PR #441, opened 2022-09-08
Require Codacy checks for bors

msimberg:
This is a test.


msimberg:
bors try

msimberg:
Closing.

It seems like Codacy only wants to put a status directly on the pull request, but not on branches like `main` or `trying` even though it tracks the quality for those branches. If we can speed up the coverage check we could run it automatically on every pull request instead of through bors, but since it currently takes about an hour per run I wouldn't want to run it all the time. Perhaps once we split the algorithms etc. out into a separate repository.

Issue #447, opened 2022-09-09
Rename runtime `state` enum


msimberg:
To `runtime_state`? And possibly at the same time use a separate enum for the scheduler state.

PR #445, opened 2022-09-09
Clean up threading base

aurianer:
Part of #16 
- Move functionalities that are not used by DLA-Future in detail namespace.
- Change `enum`s to `enum class`es.

PR #450, opened 2022-09-09
Rename `state` to `runtime_state`

aurianer:
Based on #445
Fix partially #447

### DIFF
--- a/libs/pika/runtime/include/pika/runtime/runtime.hpp
+++ b/libs/pika/runtime/include/pika/runtime/runtime.hpp
@@ -61,8 +61,8 @@ namespace pika {
         virtual notification_policy_type get_notification_policy(
             char const* prefix, os_thread_type type);
 
-        state get_state() const;
-        void set_state(state s);
+        runtime_state get_state() const;
+        void set_state(runtime_state s);
 
         /// The \a pika_main_function_type is the default function type usable
         /// as the main pika thread function.
@@ -407,7 +407,7 @@ namespace pika {
         // topology and affinity data
         threads::detail::topology& topology_;
 
-        std::atomic<state> state_;
+        std::atomic<runtime_state> state_;
 
         // support tying in external functions to be called for thread events
         notification_policy_type::on_startstop_type on_start_func_;
@@ -464,7 +464,7 @@ namespace pika {
     PIKA_EXPORT void set_error_handlers();
 
     namespace detail {
-        PIKA_EXPORT char const* get_runtime_state_name(pika::state st);
+        PIKA_EXPORT char const* get_runtime_state_name(pika::runtime_state st);
     }
 
     namespace util {
--- a/libs/pika/runtime/include/pika/runtime/runtime_fwd.hpp
+++ b/libs/pika/runtime/include/pika/runtime/runtime_fwd.hpp
@@ -82,7 +82,7 @@ namespace pika {
     ///
     /// This function returns whether the runtime system is currently being
     /// started or not, e.g. whether the current state of the runtime system is
-    /// \a pika::state::startup
+    /// \a pika:runtime_state::startup
     ///
     /// \note   This function needs to be executed on a pika-thread. It will
     ///         return false otherwise.
@@ -100,7 +100,7 @@ namespace pika {
     ///
     /// This function returns whether the runtime system is currently running
     /// or not, e.g.  whether the current state of the runtime system is
-    /// \a pika::state::running
+    /// \a pika:runtime_state::running
     ///
     /// \note   This function needs to be executed on a pika-thread. It will
     ///         return false otherwise.
@@ -111,7 +111,7 @@ namespace pika {
     ///
     /// This function returns whether the runtime system is currently stopped
     /// or not, e.g.  whether the current state of the runtime system is
-    /// \a pika::state::stopped
+    /// \a pika:runtime_state::stopped
     ///
     /// \note   This function needs to be executed on a pika-thread. It will
     ///         return false otherwise.
@@ -122,7 +122,7 @@ namespace pika {
     ///
     /// This function returns whether the runtime system is currently being
     /// shut down or not, e.g.  whether the current state of the runtime system
-    /// is \a pika::state::stopped or \a pika::state::shutdown
+    /// is \a pika:runtime_state::stopped or \a pika:runtime_state::shutdown
     ///
     /// \note   This function needs to be executed on a pika-thread. It will
     ///         return false otherwise.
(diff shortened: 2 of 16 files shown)


msimberg:
bors try-

msimberg:
bors try

libs/pika/algorithms/tests/performance/foreach_scaling.cpp
```diff
@@ -234,7 +234,7 @@ int pika_main(pika::program_options::variables_map& vm)
     chunk_size = vm["chunk_size"].as<int>();
     num_overlapping_loops = vm["overlapping_loops"].as<int>();
     disable_stealing = vm.count("disable_stealing");
-    fast_idle_mode = vm.count("fast_idle_mode");
+    scheduler_mode::fast_idle_mode = vm.count("fast_idle_mode");
```
msimberg, line 237: This is an unfortunately named `bool`. It should not be changed (this should fail to compile anyway).

msimberg: Same further down in the same file.

libs/pika/algorithms/tests/performance/foreach_scaling_helpers.hpp
```diff
@@ -35,7 +35,7 @@ int test_count = 100;
 int chunk_size = 0;
 int num_overlapping_loops = 0;
 bool disable_stealing = false;
-bool fast_idle_mode = false;
+bool scheduler_mode::fast_idle_mode = false;
```
msimberg, line 38: Same here, this should be reverted.

libs/pika/resource_partitioner/tests/unit/cross_pool_injection.cpp
```diff
@@ -67,17 +67,19 @@ int pika_main()
...
             // add these flags
-            pika::threads::scheduler_mode(pika::threads::enable_stealing |
-                pika::threads::enable_stealing_numa |
-                pika::threads::assign_work_thread_parent |
-                pika::threads::steal_after_local),
+            pika::threads::scheduler_mode(
+                pika::threads::scheduler_mode::enable_stealing |
+                pika::threads::scheduler_mode::enable_stealing_numa |
+                pika::threads::scheduler_mode::assign_work_thread_parent |
+                pika::threads::scheduler_mode::steal_after_local),
```
msimberg, line 74: I _think_ you should be able to remove the outer wrapping `pika::threads::scheduler_mode` since the `operator|` already returns a `scheduler_mode`. Same below.

libs/pika/resource_partitioner/tests/unit/scheduler_priority_check.cpp
```diff
@@ -60,18 +60,20 @@ int pika_main(variables_map& vm)
         std::cout << "Setting shared-priority mode flags" << std::endl;
         sched->add_remove_scheduler_mode(
             // add these flags
-            pika::threads::scheduler_mode(pika::threads::enable_stealing |
-                pika::threads::enable_stealing_numa |
-                pika::threads::assign_work_round_robin |
-                pika::threads::steal_high_priority_first),
+            pika::threads::scheduler_mode(
```
msimberg, line 63: Same comment here about the wrapping `pika::threads::scheduler_mode`.

libs/pika/threading_base/src/scheduler_mode.cpp
```diff
@@ -0,0 +1,43 @@
...
+    {
+        return static_cast<std::uint32_t>(sched) !=
+            static_cast<std::uint32_t>(i);
+    }
+
+    bool operator==(scheduler_mode sched, int i)
+    {
+        return static_cast<std::uint32_t>(sched) ==
+            static_cast<std::uint32_t>(i);
+    }
```
msimberg, line 36: Do we still need these?

tests/performance/local/future_overhead.cpp
```diff
@@ -164,14 +164,15 @@ void function_futures_limiting_executor(
     {
         sched->add_remove_scheduler_mode(
             // add these flags
-            pika::threads::scheduler_mode(pika::threads::enable_stealing |
-                pika::threads::enable_stealing_numa |
-                pika::threads::assign_work_round_robin |
-                pika::threads::steal_after_local),
+            pika::threads::scheduler_mode(
```
msimberg, line 167: Wrapping `scheduler_mode`.

tests/performance/local/future_overhead_report.cpp
```diff
@@ -78,14 +78,15 @@ void measure_function_futures_create_thread_hierarchical_placement(
     {
         sched->add_remove_scheduler_mode(
             pika::threads::scheduler_mode(
-                pika::threads::assign_work_thread_parent),
-            pika::threads::scheduler_mode(pika::threads::enable_stealing |
-                pika::threads::enable_stealing_numa |
-                pika::threads::assign_work_round_robin |
-                pika::threads::steal_after_local |
-                pika::threads::steal_high_priority_first));
+                pika::threads::scheduler_mode::assign_work_thread_parent),
```
msimberg, line 81: Same here.

msimberg (approved):
I realized only after starting to review that most of my comments actually already apply to #445. However, none of them are blockers for this PR and you're free to distribute them across the PRs (including new ones) as you wish.

libs/pika/resource_partitioner/tests/unit/cross_pool_injection.cpp
msimberg: In my opinion `scheduler_mode` is mentioned enough times in that expression that it doesn't add anything. It's like writing `int(int(3) + int(17) - int(5))`.

libs/pika/threading_base/src/scheduler_mode.cpp
msimberg: Where do we need them? More importantly, are they needed by users of `scheduler_mode` (my guess is no)?

libs/pika/resource_partitioner/tests/unit/cross_pool_injection.cpp
msimberg: Yeah, agreed on that point and to be honest I think `add_remove_scheduler_mode` is a mistake, and we should only have `add_scheduler_mode` and `remove_scheduler_mode`. Then it's obvious what is being added/removed and more importantly the order in which adding/removing happens is clearer. Not for this PR. I opened an issue: https://github.com/pika-org/pika/issues/455.

libs/pika/threading_base/src/scheduler_mode.cpp
msimberg: Thanks. Yeah, in that case please remove the `!=` and `==` operators and I think you should be able to compare to `scheduler_mode{}` instead.

msimberg: Enums are already comparable, so you don't need any overloads for that.

PR #436, opened 2022-09-06
Replace p2300 `fetch_content` by a `find_package` to allow using p2300 spack package

aurianer:

PR #451, opened 2022-09-13
Add `+p2300` variant to Piz Daint Jenkins CI configurations

msimberg:
I've updated the Piz Daint spack commit to include the `p2300` and `whip` packages. This PR doesn't actually attempt to use them, but just adds them to the environment for #436 and #423. Will do ault separately.

@aurianer note that I haven't constrained the P2300 commit in any way. If you rebase #436 and it turns out that there are now build problems, it may due to bugs in P2300 that are not there in the pinned commit in `pika_setup_p2300.cmake`. We can either try an older commit or ideally try to fix the problems, if there are any.


msimberg:
bors try

msimberg:
bors merge

PR #452, opened 2022-09-13
Fix CMake options for clang-14-cuda Jenkins config

msimberg:
Actually fixes #302. I never actually enabled CUDA on the clang 14 configuration. I expect this to fail at the moment. I have not tried the build myself.

PR #453, opened 2022-09-13
Add missing `+cuda` variant to specs for CUDA builds on Jenkins

msimberg:
The specs had `cuda` specified as a dependency but not `+cuda` specified as a variant, which meant that e.g. the `whip` dependency wasn't added to the `build-env`.


msimberg:
bors merge

msimberg:
This is not doing what I hoped for the `cce` configuration. It seems to break CUDA compilation with `CC` for some odd reason. Since we (almost...) have another clang+cuda configuration (https://github.com/pika-org/pika/pull/452) I may stop using `CC` for CUDA compilation, or try to use `CC` + `nvcc`... this needs a bit more investigation.

msimberg:
This is now based on #452 instead. I've simply disabled CUDA on the CCE build. I've also fixed some warnings about defaulted constructors being implicitly host-device.

msimberg:
bors try

msimberg:
bors try-

msimberg:
bors merge

msimberg:
bors cancel

msimberg:
bors merge

PR #457, opened 2022-09-16
Don't use hipcc as `CXX` in `linux_hip` GitHub actions workflow

msimberg:
`CXX` is no longer set to `hipcc` in version 3 of the CI image.


msimberg:
bors try

PR #458, opened 2022-09-19
Make signal handlers optional

### LINKED ISSUE #406: Reconsider setting termination handlers
Setting the termination handlers does have value because we can print e.g. stacktraces which are not printed by default on segfaults, pika thread-specific information etc.

On the other hand, when pika is embedded in an application behind multiple other libraries, combined with other runtimes etc. who is to say that pika's termination handler is _the_ termination handler to install?

Some in-between solutions would be to allow users to disable/enable pika's termination handler explicitly (with config options or explicit function calls).

msimberg:
Fixes #406.

Adds a configuration option `pika.install_signal_handlers` which defaults to off (`0`). Only forces the signal handlers to be installed if `--pika:attach-debugger=exception`. Also adds a missing default entry for `pika.diagnostics_on_terminate`.

### DIFF
--- a/libs/pika/command_line_handling/src/command_line_handling.cpp
+++ b/libs/pika/command_line_handling/src/command_line_handling.cpp
@@ -755,7 +755,15 @@ namespace pika::detail {
             else
             {
                 if (option == "startup")
+                {
                     debug::detail::attach_debugger();
+                }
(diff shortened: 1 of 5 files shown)


msimberg:
bors try

PR #460, opened 2022-09-22
For spack-user-config to be the default one in $SPACK_ROOT

aurianer:
We want to use the default `spack-user-config` as we want all the CI compilers to be available.

### DIFF
--- a/.jenkins/cscs/env-common.sh
+++ b/.jenkins/cscs/env-common.sh
@@ -7,8 +7,8 @@
 module load daint-gpu spack-config
 
 export SPACK_ROOT="/apps/daint/SSL/pika/spack"
-set +u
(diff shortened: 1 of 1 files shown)


.jenkins/cscs/env-common.sh
```diff
@@ -8,7 +8,7 @@ module load daint-gpu spack-config
 
 export SPACK_ROOT="/apps/daint/SSL/pika/spack"
 set +u
-test -z "$SPACK_USER_CONFIG_PATH" && export SPACK_USER_CONFIG_PATH="${SPACK_ROOT}/../spack-user-config"
+export SPACK_USER_CONFIG_PATH="${SPACK_ROOT}/../spack-user-config"
```
msimberg, line 11: Move this outside the `set +u` (just for style, it doesn't have a practical difference since `SPACK_ROOT` is definitely set). Feel free to merge after that.

PR #461, opened 2022-09-22
Clean up `thread_support` module

aurianer:
Part of #16.


libs/pika/include/include/pika/mutex.hpp
```diff
@@ -13,12 +13,12 @@
 #include <pika/thread_support/unlock_guard.hpp>
 
 namespace pika {
+    using pika::detail::unlock_guard;
```
msimberg, line 16: Let's remove this `using` since `unlock_guard` is now in `detail`.

msimberg:
bors merge

Issue #289, opened 2022-07-05
Allow `(unique_)any_sender` to have any set of completion signatures


msimberg:
Our implementation is currently limited to hold only one set of values sent from a wrapped sender. If the internal `any_receiver_impl::set_value` would instead accept a variant of tuples it would still not need to be templated but could support multiple sets of values sent from the predecessor. We could probably also have real deduction guides for the `(unique_)any_sender` constructors instead of `make(_unique)_any_sender`.

msimberg:
I think this makes most sense once we require C++20 and only use the P2300 reference implementation. Then we can simply put `completion_signatures` into the `any_sender` template parameters.

PR #464, opened 2022-09-26
Add `set_stopped_t()` to `(unique_)any_sender` completion signatures

msimberg:
Without #289 we need to pessimistically expect that all senders that get wrapped in a `(unique_)any_sender` may signal `set_stopped_t()`. This is one more tiny step towards using the reference implementation in DLA-Future.


msimberg:
bors try

msimberg:
bors try

Issue #164, opened 2022-03-31
Fix `algorithm_transform_mpi_test` with MPICH


aurianer:
Fix the error path of the `algorithm_transform_mpi.cpp` test with MPICH.
Currently working with openmpi

msimberg:
Fixed by #466.

PR #466, opened 2022-09-29
Run MPI tests with MPICH on Jenkins and fix error handling tests with MPICH

### LINKED ISSUE #418: Attempt to enable MPI tests on Jenkins
Assuming we can do nested `srun`s inside a batch job (which used to work) we should be able to change the MPI tests to use `srun` in place of `mpirun`. It would be good to have slightly more comprehensive MPI tests.

msimberg:
Attempt to fix #418 and #164.

For #418 we'll see what slurm says about launching jobs within jobs.

For #164 the problem seems to have been that OpenMPI and MPICH simply seem to have different definitions of what they consider to be errors. OpenMPI considers `-1` as a root rank to be a problem. MPICH doesn't. They both seem to consider `MPI_DATATYPE_NULL` as a data type to be an error, however, so I've updated the error-handling tests to use that instead to trigger errors. OpenMPI and MPICH also print different error messages, so they have to be handled separately.

### DIFF
--- a/.jenkins/cscs/env-cce.sh
+++ b/.jenkins/cscs/env-cce.sh
@@ -11,7 +11,7 @@ hwloc_version="2.0.3"
 spack_compiler="cce@${cce_version}"
 spack_arch="cray-cnl7-broadwell"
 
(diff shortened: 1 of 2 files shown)


msimberg:
bors try

msimberg:
bors merge

msimberg:
bors merge

PR #492, opened 2022-10-28
More reference type handling fixes to `when_all`, `then_with_stream`, and `transform_mpi`

msimberg:
This fixes three more reference handling problems that were not included in #284:
1. `when_all` needs to `decay`/`remove_cvref` the types that it sends to successors. Previously it would directly use the value types of predecessors for its own value types. Since it needs to store the values internally before sending them onward to successors (my move) it needs to decay the types.
2. `then_with_stream` and `transform_mpi` did not decay value types before storing them internally.

This adds tests for the above cases as well. This must go into 0.10.0.


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
The mac M1 runner is still not running, but all the other configurations are passing so I'm going to merge this manually.

PR #494, opened 2022-11-01
Add support for Tracy 0.9

msimberg:
Tracy 0.9 moved the Tracy.hpp header to a subdirectory tracy. This adds support for 0.9 and keeps supporting old versions by checking if the `tracy/Tracy.hpp` is available and using it if it is. Otherwise it falls back to the old Tracy.hpp include.

This is a last minute PR for the 0.10.0 release.


msimberg:
bors try

PR #500, opened 2022-11-03
Remove `add_remove_scheduler_mode`

### LINKED ISSUE #455: Remove `add_remove_scheduler_mode`
The convenience from it is questionable, and the clarity gained from using `add_scheduler_mode` and `remove_scheduler_mode` separately I think makes up for any potential convenience gained from having a combined function.

msimberg:
Fixes #455.

The main difference of the combined call compared to two separate calls is that there will be two atomic updates. However, I don't think these are in any way meant to be used in performance critical parts and there are other factors that make relying on this being atomic not a very good assumption (`add_remove_scheduler_mode` was not really atomic either).

### DIFF
--- a/libs/pika/resource_partitioner/tests/unit/cross_pool_injection.cpp
+++ b/libs/pika/resource_partitioner/tests/unit/cross_pool_injection.cpp
@@ -66,18 +66,12 @@ int pika_main()
         sched->get_description())
     {
         using ::pika::threads::scheduler_mode;
-        sched->add_remove_scheduler_mode(
-            // add these flags
-            scheduler_mode::enable_stealing |
-                scheduler_mode::enable_stealing_numa |
-                scheduler_mode::assign_work_thread_parent |
-                scheduler_mode::steal_after_local,
-            // remove these flags
+        sched->add_scheduler_mode(scheduler_mode::assign_work_thread_parent |
+            scheduler_mode::steal_after_local);
+        sched->remove_scheduler_mode(scheduler_mode::enable_stealing |
+            scheduler_mode::enable_stealing_numa |
             scheduler_mode::assign_work_round_robin |
-                scheduler_mode::steal_high_priority_first);
-        sched->update_scheduler_mode(scheduler_mode::enable_stealing, false);
-        sched->update_scheduler_mode(
-            scheduler_mode::enable_stealing_numa, false);
+            scheduler_mode::steal_high_priority_first);
     }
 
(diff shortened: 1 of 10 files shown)


libs/pika/resource_partitioner/tests/unit/cross_pool_injection.cpp
```diff
@@ -66,15 +66,14 @@ int pika_main()
...
-            scheduler_mode::assign_work_round_robin |
-                scheduler_mode::steal_high_priority_first);
+        sched->add_scheduler_mode(scheduler_mode::enable_stealing |
+            scheduler_mode::enable_stealing_numa |
+            scheduler_mode::assign_work_thread_parent |
+            scheduler_mode::steal_after_local);
+        sched->remove_scheduler_mode(scheduler_mode::assign_work_round_robin |
+            scheduler_mode::steal_high_priority_first);
+        // TODO: Should enable_stealing_numa and enable_stealing just be removed
+        // in the call above?
```
msimberg, line 76: @biddisco could I ask you to comment on what you intended to do here. There are two calls below to `update_scheduler_mode` which individually disable `enable_stealing_numa` and `enable_stealing`. Presumably they should just be moved from `add_scheduler_mode` to `remove_scheduler_mode` and the `update_scheduler_mode` calls removed from below?

biddisco: Note : on my branch that "todo" comment does not exist. Odd.
The individual flags here do not really matter - the purpose of the test is to randomly inject tasks from one pool to another and make sure nothing crashes. The flags for stealing across numa are usually moot because we create many small pools and so they don't cross numa boundaries generally, hence the lack of need for enable stealing across numa. disabling stealing yes/no is mostly to further stress the scheduler - if stealing is allowed, lockups are less probable, but if stealing is disabled, a worker thread getting stuck somehow with stuff still in the queue is more likely to result in a deadlock/test timeout. 
Functionally speaking, the flags can be set to anything really and the test ought to pass.

msimberg: I've disabled the stealing and removed the calls to `update_scheduler_mode`.

msimberg:
bors try

msimberg:
bors merge

msimberg:
Not quite sure what happened on Piz Daint in those last jobs, so retrying.

PR #501, opened 2022-11-03
Make mimalloc the default allocator

### LINKED ISSUE #231: Make mimalloc the default allocator?
Previous tests showed that mimalloc often performed the best. We could change the default in pika and/or the spack package.

Further benchmarks in DLA-Future are needed before we conclude that it's really the best option.

At least mimalloc has a package in the Ubuntu repositories and a spack package, so the change is feasible from that perspective.

msimberg:
Fixes #231.


msimberg:
bors merge

PR #504, opened 2022-11-08
Update minimum HIP version in README.rst

msimberg:
While lower versions than 5.2.0 may work, in practice we require an as up-to-date version as possible.


msimberg:
bors merge

Issue #19, opened 2022-01-18
Separate algorithms into a separate repository


msimberg:
The main open question here is if the `algorithms` project should rely on the `pika` runtime, the other way around, or neither? The default execution policies assume that a global thread pool exists which would normally be set up by the runtime.

PR #507, opened 2022-11-10
Add `inshpect` checks

### LINKED ISSUE #6: Move inspect tool out from repository and allow using configuration file
This will be more useful when/if parts of the repository (e.g. algorithms) are in separate repositories. With a single repository this is not very urgent.

msimberg:
Almost fixes #6.

Adds a minimal version of `inspect` (https://github.com/msimberg/inshpect) to run alongside the current `inspect`. It does not have all the features of `inspect`, but has some other benefits. I'm mostly doing this so that we don't need to duplicate the `inspect` tool for #19 (and similar).

Please see the README of `inshpect` for more details but it keeps the following checks from `inspect` because they are the most useful:
- `pragma_once`
- `license`
- `spdx_license`
- `copyright`
- `deprecated_macros`
- `deprecated_include`
- `deprecated_names`
- `include`

I've left out the following checks because I think they are better enforced by formatting tools:
- `crlf`
- `end`
- `link`
- `tab`
- `ascii`
- `whitespace`
- `length`

The following I've left out simply because I see little value in them:
- `path_name`
- `apple`
- `windows`
- `assert`
- `minmax`
- `unnamed`

We can add more checks to `inshpect` if we see that the checks that I've left out actually end up being regular problems, but so far I think almost all failures reported by `inspect` are in the first category and occasionally too long lines (always in comments, because `clang-format` doesn't reformat those; I think we can live with too long lines in comments).

Importantly, `inshpect` gets the patterns to check for from a config file, so all we need to keep in the repo is one  (or two) config files.

- The main config file `.inshpect.toml` is for all C++ files and enables all checks. It's currently restricted to `libs` because the `inspect` tool ironically doesn't pass most checks. I want to run both tools side-by-side for a while to see that they roughly catch the same issues. Once we're relatively confident in `inshpect` we can remove `inspect`.
- The secondary config file `.inspect-noncpp.toml` is for everything else. It disables the C++-specific checks (includes, deprecated names, etc.) but still checks licenses etc. on other files.

`inshpect` already caught some issues that `inspect` apparently didn't catch. I don't know what causes this difference.

`inshpect` doesn't know how to ignore failures in C-style `/* ... */` comments, so I've changed some to `//`-style. We just have some Boost-legacy code that uses the old-style comments.

Finally, `inshpect` does run slightly slower than `inspect`. The `inshpect` check takes roughly 6 minutes on CircleCI and `inspect` takes roughly 4 minutes, but I think that's still acceptable. There may be room to speed up `inshpect` (likewise `inspect`).

@aurianer I would greatly appreciate any ideas you have for this and potentially refactoring/improving `inshpect` itself.

To do:
- [x] Add `inshpect` and dependencies to CI images
- [x] Add back `noinspect` for old `inspect` tool


msimberg:
bors try

msimberg:
bors try

.circleci/config.yml
```diff
@@ -300,6 +300,32 @@ jobs:
...
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: /pika
+      - run:
+          name: Running inshpect
+          command: |
+              export INSHPECT_NUMTHREADS=2
+              export INSHPECT_VERBOSE=1
+              # TODO: Currently restricted to libs, remove libs when regular inspect tool is removed
```
msimberg, line 313: Just to clarify, this TODO will stay for now. I will keep #6 open after this is merged.

msimberg:
bors try

msimberg:
bors merge

PR #509, opened 2022-11-17
Clean up `thread_pool_util` module

aurianer:
Is there anything more to do here?


msimberg:
> Is there anything more to do here?

In principle no, but this is functionality that I'm not convinced is necessary. Full runtime suspension yes, but individual pool suspension not necessarily. However, since I'm not 100% sure this can be removed yet, and if it's removed and the scheduler is refactored adding it back will not be entirely trivial, I'd like to ask you to just move the functionality to `detail` for the moment (under `pika::threads` or directly under `pika` are both fine).

msimberg (approved):

msimberg:
bors merge

Issue #448, opened 2022-09-09
Avoid nesting `detail` namespaces into `experimental`


msimberg:
It's unnecessary, and in the worst case also contributes to debug bloat.

PR #510, opened 2022-11-18
Fix compilation with HIP and stdexec

### LINKED ISSUE #336: Test P2300 reference implementation with hipcc
Once #47 is done, we can attempt to add `PIKA_WITH_P2300_REFERENCE_IMPLEMENTATION` to that configuration.

Like with CPU-only builds of the P2300 reference implementation the same restrictions most likely apply with HIP as well:
- clang + libstdc++ requires clang 14 and libstdc++ 11 for proper coroutines support.
- clang + libc++ requires LLVM 11 or newer, possibly 12?
- GCC + libstdc++ seems to require GCC 11, but fails to compile on many things (which is why we don't currently have `PIKA_WITH_P2300_REFERENCE_IMPLEMENTATION` enabled with GCC in CI).

That means we probably want a HIP build with a HIP version based on llvm 14 or newer and gcc 11 or newer for CI.

aurianer:
- Use gcc 12.2.0 (as needed for stdexec)
- Therefore update to hip@5.3.0 (needed to work with gcc@12)
- Move `result_type_signature_helper` and `single_result` into a `helpers.hpp` file in the execution module.

Fixes #336.


msimberg:
What's the status of this? Does it look like GCC 11 is fine for stdexec?

libs/pika/execution/include/pika/execution/algorithms/detail/helpers.hpp
```diff
@@ -7,12 +7,33 @@
 #pragma once
 
 #include <pika/config.hpp>
+#include <pika/execution_base/receiver.hpp>
 #include <pika/type_support/pack.hpp>
 
 #include <type_traits>
 #include <variant>
 
+namespace pika::detail {
```
msimberg, line 16: Let's maybe keep this in the `pika::execution::experimental::detail` namespace (as for `single_result`)? I think the struct name needs to be a bit more explicit if it's going to be in the top-level `pika::detail` at some point. https://github.com/pika-org/pika/issues/448 is relevant, but no need to mix that into this PR.

msimberg:
HIP 5.3.0 is now available in spack.

.jenkins/cscs-ault/env-hipcc.sh
```diff
@@ -5,18 +5,20 @@
...
 spack_compiler="gcc@${gcc_version}"
 spack_arch="linux-centos8-zen"
 
-spack_spec="pika@main+rocm amdgpu_target='gfx900,gfx906' arch=${spack_arch} %${spack_compiler} malloc=system cxxstd=${cxx_std} ^boost@${boost_version} ^hwloc@${hwloc_version} ^hip@${hip_version} ^whip amdgpu_target='gfx900,gfx906'"
+spack_spec="pika@main+rocm+p2300 amdgpu_target='gfx900:xnack-,gfx906:xnack-' arch=${spack_arch} %${spack_compiler} malloc=system cxxstd=${cxx_std} ^boost@${boost_version} ^hwloc@${hwloc_version} ^hip@${hip_version} ^whip amdgpu_target='gfx900:xnack-,gfx906:xnack-'"
 
 configure_extra_options+=" -DCMAKE_BUILD_RPATH=$(spack location -i ${spack_compiler})/lib64"
-configure_extra_options+=" -DCMAKE_HIP_ARCHITECTURES=gfx900;gfx906"
+configure_extra_options+=" -DCMAKE_HIP_ARCHITECTURES=gfx900:xnack-;gfx906:xnack-"
+configure_extra_options+=" -DCMAKE_CXX_FLAGS=-DSTDEXEC_DISABLE_STD_DEPRECATIONS"
```
msimberg, line 20: Do we still need this?

msimberg: But what are those warnings? Can we fix the warnings instead?

.jenkins/cscs-ault/env-hipcc.sh
```diff
@@ -5,18 +5,21 @@
...
 spack_arch="linux-centos8-zen"
+stdexec_version="c0fe16f451137b00ceb85d912ec3a288990ce766"
 
-spack_spec="pika@main+rocm amdgpu_target='gfx900,gfx906' arch=${spack_arch} %${spack_compiler} malloc=system cxxstd=${cxx_std} ^boost@${boost_version} ^hwloc@${hwloc_version} ^hip@${hip_version} ^whip amdgpu_target='gfx900,gfx906'"
+# The xnack- architectures are not supported by rocblas@5.3.3
+spack_spec="pika@main+rocm+p2300 amdgpu_target='gfx900,gfx906' arch=${spack_arch} %${spack_compiler} malloc=system cxxstd=${cxx_std} ^boost@${boost_version} ^hwloc@${hwloc_version} ^hip@${hip_version} ^whip amdgpu_target='gfx900,gfx906' ^stdexec@${stdexec_version}"
 
 configure_extra_options+=" -DCMAKE_BUILD_RPATH=$(spack location -i ${spack_compiler})/lib64"
-configure_extra_options+=" -DCMAKE_HIP_ARCHITECTURES=gfx900;gfx906"
+configure_extra_options+=" -DCMAKE_HIP_ARCHITECTURES=gfx900:xnack-;gfx906:xnack-"
```
msimberg, line 21: Should we keep these the same as in the spack spec?

libs/pika/async_cuda/tests/performance/synchronize.cu
```diff
@@ -155,6 +155,10 @@ int pika_main(pika::program_options::variables_map& vm)
                 ex::transfer(ex::thread_pool_scheduler{}) | ex::then([] {}) |
                 ex::transfer(sched) | cu::then_with_stream(f) |
                 ex::transfer(ex::thread_pool_scheduler{}) | ex::then([] {}) |
+            // The compilation of this loop unrolling with llvm-amdgpu's
+            // Clang 5.3.3 is hanging.
+#if !(defined(PIKA_HAVE_P2300_REFERENCE_IMPLEMENTATION) ||                     \
+    defined(PIKA_HAVE_HIP))
```
msimberg, line 161: Please ifdef out this test or change the batch size for the full test. Right now this is misleading since it doesn't spawn as many kernels as the other tests.

.jenkins/cscs-ault/entry.sh
```diff
@@ -38,7 +38,7 @@ sbatch \
     --nodes="1" \
     --partition="${configuration_slurm_partition}" \
     --account="djenkssl" \
-    --time="00:30:00" \
+    --time="02:00:00" \
```
msimberg, line 41: Do we need such long time here? As far as I see this is still only building examples and the CUDA/HIP tests, correct?

msimberg (changes requested):
Could you please rebase on master?

msimberg (approved):

msimberg:
bors merge

PR #522, opened 2022-11-28
Don't link to `stdc++fs` anymore

msimberg:
It is no longer needed by our minimum supported compiler versions.


msimberg:
bors merge

msimberg:
bors merge

msimberg:
bors merge

msimberg:
Ugh, I was wrong (I probably knew at some point, but forgot). This is still needed for Cray's environment which uses libstdc++ 8 (starting at version 9 linking to `stdc++fs` separately is no longer required). Closing without merging but we may be able to come back to this soon enough again if we upgrade the Cray version used in CI.

PR #529, opened 2022-11-30
Attempt to use ccache on macOS CI workflows

msimberg:
I suspect there was a reason for not using ccache on macOS, but I'm hoping I'm misremembering, so retrying.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -11,10 +11,7 @@ status = [
   "github/linux/hip/fast",
   "github/linux/sanitizers/fast",
   "github/linux/tracy/fast",
-  "github/macos/debug/core",
-  "github/macos/debug/tests/examples_regressions",
-  "github/macos/debug/tests/performance",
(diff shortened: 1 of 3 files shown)


msimberg:
bors try

msimberg:
This disabled ccache previously: https://github.com/pika-org/pika/pull/409. I'm hoping that maybe now that algorithms are split out the cache doesn't end up quite as big as before, so maybe it'll work a bit better this time.

msimberg:
bors try-

msimberg:
bors try

msimberg:
bors try

msimberg:
The cache is indeed smaller now and might actually work ok. The arm64 runner seems to be down at the moment so we'll have to wait with merging this, but it looks promising at least.

msimberg:
bors merge

msimberg:
bors try-

msimberg:
bors merge

PR #545, opened 2022-12-14
Replace pika's `is_invocable` with the `std` equivalent

### LINKED ISSUE #516: Attempt to replace `is_invocable` with `std` equivalent
This _should_ have no effect during runtime, but may have an effect on compile times. If the effect is small enough it's probably still worth it to use `std::is_invocable`.

msimberg:
Fixes #516.


msimberg:
bors merge

PR #548, opened 2023-01-04
`async_rw_mutex` enhancements

msimberg:
Changes a few essential and less essential things in `async_rw_mutex` for use in DLA-Future.
- The most important fix is in https://github.com/pika-org/pika/commit/ead0444d4ec1f41f40b2474e67436bb1e09d8ca7. It fixes deadlocks which happen if one tries to do a read-only access after a read-write access, and the read-only access is eagerly waited for with e.g. `sync_wait` in the same scope as the `async_rw_mutex` itself. This was caused by the `async_rw_mutex` holding on to shared states too long. I'm not 100% happy with the current fix (especially the `yield_while`), but it will do for now.
- This also makes read-only access senders connectable with l-value references.
- Removes the conversion operators from the wrapper types. They're gone because they cause problems when used in DLA-Future (specifically, they seem to require complete types which is a problem in DLA-Future where `async_rw_mutex`es of `Tile`s can be stored in `Tile`s).
- Minor changes are use of `no_unique_address` and more const/constexpr in various places.

### DIFF
--- a/libs/pika/synchronization/include/pika/synchronization/async_rw_mutex.hpp
+++ b/libs/pika/synchronization/include/pika/synchronization/async_rw_mutex.hpp
@@ -718,6 +718,22 @@ namespace pika::execution::experimental {
                 return operation_state<R>{PIKA_FORWARD(R, r),
                     PIKA_MOVE(s.prev_state), PIKA_MOVE(s.state)};
             }
+
+            template <typename R>
+            friend auto tag_invoke(pika::execution::experimental::connect_t,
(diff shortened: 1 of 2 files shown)


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

msimberg:
bors merge

msimberg:
bors cancel

msimberg:
bors merge

PR #571, opened 2023-01-26
Disable Jenkins HIP configuration in bors configuration

msimberg:
The Jenkins ault agent is currently down and blocking other PRs from merging.

PR #581, opened 2023-01-31
Add release notes for 0.12.0

msimberg:


msimberg:
@aurianer note that I'm not including #510 in these release notes only because of https://github.com/pika-org/pika/pull/571. Let's include it in the next release notes, assuming we have the ault CI back by then.

PR #583, opened 2023-02-02
Move nested `done_type` helper structs into common definition

msimberg:
A lot of sender adaptors use a `struct done_type {}` in variants to mark that `set_stopped` has been called. This:
- renames them to `struct stopped_type {}`, and
- moves them out of the sender definitions into a single, common definition (nesting them inside the senders increases debug bloat)

I've also renamed the helper visitors `done_error_value_visitor` to `stopped_error_value_visitor` to match the `done` to `stopped` change.


msimberg:
bors try

msimberg:
bors merge

PR #587, opened 2023-02-08
Add `const` to l-value reference connect overloads

msimberg:
This makes no practical difference, but may prevent accidental moves from the reference in the future.


msimberg:
bors try

msimberg:
bors try

msimberg:
bors merge

Issue #11, opened 2022-01-18
Remove filesystem compatibility layer


msimberg:
All versions of macOS that should be supported already have `<filesystem>`. This also removes one more Boost dependency (Boost `error_code`).

PR #589, opened 2023-02-08
Remove leftover `std::filesystem` feature test file

msimberg:
The CMake feature test was removed in #11 but the test file was apparently still hanging around. This removes the file, and the cmake-format entry for the CMake feature test function.


msimberg:
bors try

msimberg:
bors merge

PR #592, opened 2023-02-08
Minor fix in `cuda_sender` test

msimberg:
The `cuda_sched_copy` variable was unused, `cuda_sched` was used instead. This simply uses the copy instead of the original scheduler. This shouldn't fix any bug or change the behaviour, but at least uses the variable.


msimberg:
bors try

msimberg:
bors merge

Issue #557, opened 2023-01-10
Attempt to use `condition_variable` to signal completion instead of callback in CUDA and MPI integrations


msimberg:
As pointed out by @biddisco an alternative way to signal completion in the CUDA and MPI integrations would be to have the receivers wait on a condition variable and have the polling loops signal that condition variable, instead of the current approach of running a continuation in the polling loop. This has the benefit of naturally preserving the scheduler context for continuations, and the downside of requiring that condition variable is waited for on a pika thread (not a problem in current use cases, but may limit its use in more generic contexts, i.e. on other schedulers).

PR #593, opened 2023-02-09
Improve MPI polling

### LINKED ISSUE #187: Optimize MPI polling implementation
Based on benchmarks in DLA-Futures it looks like using MPI polling instead of a `yield_while` on a request is slightly slower. It's unclear if this is something inherent in the polling implementation, or something else. One possibility is that with polling too many requests are submitted to MPI and its performance degrades. We could experiment with limiting the number of active requests allowed to see if this improves performance.

biddisco:
Fixes #187 and partially addresses #557

This PR reworks extensively the mpi_polling, the most important feature is that continuations are not triggered under lock, and that they can be explicitly transferred to a new task/pool.

Throttling is possible on a per stream basis, initially disabled until dlaf testing requires it. 
The number of completions to handle per poll iteration may be controlled.
env vars that control behaviour are
`PIKA_MPI_COMPLETION_MODE` sets the mode to use for handling completions 
`PIKA_MPI_POLLING_SIZE` sets the number of completions to handle per poll iteration
`PIKA_MPI_MSG_THROTTLE` sets the number of messages (per stream) to allow in-flight (disabled by default)

### DIFF
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -199,8 +199,6 @@ namespace pika::mpi::experimental::detail {
                 using type = util::detail::transform_t<Tuple, std::decay>;
             };
 
-            // note that result_type = pika::detail::variant<std::tuple<MPI_Request>>;
-
             template <typename Receiver_, typename F_, typename Sender_>
             operation_state(Receiver_&& receiver, F_&& f, Sender_&& sender, stream_type s)
               : receiver(PIKA_FORWARD(Receiver_, receiver))
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -120,6 +120,10 @@ namespace pika::mpi::experimental::detail {
                             using namespace pika::debug::detail;
                             using ts_element_type = std::tuple<std::decay_t<Ts>...>;
                             using invoke_result_type = mpi_request_invoke_result_t<F, Ts...>;
+                            using namespace pika::debug::detail;
+                            PIKA_DETAIL_DP(mpi_tran<5>,
+                                debug(str<>("dispatch_mpi_recv"), "set_value_t", "stream",
+                                    detail::stream_name(r.op_state.stream_)));
 
                             // move/copy the received params into our local opstate
                             r.op_state.ts.template emplace<ts_element_type>(
@@ -127,7 +131,7 @@ namespace pika::mpi::experimental::detail {
                             // and get a reference to the tuple of local params
                             auto& t = std::get<ts_element_type>(r.op_state.ts);
                             // init a request
-                            MPI_Request request{MPI_REQUEST_NULL};
+                            MPI_Request request;
                             int status = MPI_SUCCESS;
                             // invoke the function, with the contents of the param tuple as args
                             pud::invoke_fused(
@@ -136,25 +140,27 @@ namespace pika::mpi::experimental::detail {
                                     if constexpr (std::is_void_v<invoke_result_type>)
                                     {
                                         PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
-                                        PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
-                                            "MPI_REQUEST_NULL returned from mpi "
-                                            "invocation");
                                     }
                                     else
                                     {
                                         status =
                                             PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
-                                        PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
-                                            "MPI_REQUEST_NULL returned from mpi "
-                                            "invocation");
                                     }
+                                    PIKA_DETAIL_DP(mpi_tran<7>,
+                                        debug(str<>("dispatch_mpi_recv"), "invoke mpi",
+                                            detail::stream_name(r.op_state.stream_), request));
+
+                                    PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
+                                        "MPI_REQUEST_NULL returned from mpi invocation");
                                 },
                                 t);
-
                             // function called, Ts... can now be released (if refs hold lifetime)
                             r.op_state.ts = {};
                             if (poll_request(request))
                             {
+                                PIKA_DETAIL_DP(mpi_tran<7>,
+                                    debug(str<>("dispatch_mpi_recv"), "eager poll ok",
+                                        detail::stream_name(r.op_state.stream_), request));
                                 // calls set_value(request), or set_error(mpi_exception(status))
                                 set_value_error_helper(
                                     status, PIKA_MOVE(r.op_state.receiver), MPI_REQUEST_NULL);
@@ -204,8 +210,8 @@ namespace pika::mpi::experimental::detail {
 #endif
             ts_type ts;
 
+            // we don't bother storing the result, but the correct type is ...
             using result_type = pika::detail::variant<std::tuple<MPI_Request>>;
-            result_type result;
 
             template <typename Receiver_, typename F_, typename Sender_>
             operation_state(Receiver_&& receiver, F_&& f, Sender_&& sender, stream_type s)
@@ -214,9 +220,9 @@ namespace pika::mpi::experimental::detail {
               , stream_{s}
               , op_state(exp::connect(PIKA_FORWARD(Sender_, sender), dispatch_mpi_receiver{*this}))
             {
-                PIKA_DETAIL_DP(mpi_tran,
-                    debug(
-                        debug::detail::str<>("operation_state"), "stream", detail::stream_name(s)));
+                using namespace pika::debug::detail;
+                PIKA_DETAIL_DP(
+                    mpi_tran<5>, debug(str<>("operation_state"), "stream", detail::stream_name(s)));
             }
 
             friend constexpr auto tag_invoke(exp::start_t, operation_state& os) noexcept
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -131,35 +131,25 @@ namespace pika::mpi::experimental::detail {
                                 debug(str<>("dispatch_mpi_recv"), "set_value_t", "stream",
                                     detail::stream_name(r.op_state.stream_)));
 
-                            // move/copy the received params into our local opstate
-                            r.op_state.ts.template emplace<ts_element_type>(
-                                PIKA_FORWARD(Ts, ts)...);
-                            // and get a reference to the tuple of local params
-                            auto& t = std::get<ts_element_type>(r.op_state.ts);
                             // init a request
                             MPI_Request request;
                             int status = MPI_SUCCESS;
-                            // invoke the function, with the contents of the param tuple as args
-                            pud::invoke_fused(
-                                [&](auto&... ts) mutable {
-                                    // execute the mpi function call, passing in the request object
-                                    if constexpr (std::is_void_v<invoke_result_type>)
-                                    {
-                                        PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
-                                    }
-                                    else
-                                    {
-                                        status =
-                                            PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
-                                    }
-                                    PIKA_DETAIL_DP(mpi_tran<7>,
-                                        debug(str<>("dispatch_mpi_recv"), "invoke mpi",
-                                            detail::stream_name(r.op_state.stream_), ptr(request)));
+                            // execute the mpi function call, passing in the request object
+                            if constexpr (std::is_void_v<invoke_result_type>)
+                            {
+                                PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
+                            }
+                            else
+                            {
+                                status = PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
+                            }
+                            PIKA_DETAIL_DP(mpi_tran<7>,
+                                debug(str<>("dispatch_mpi_recv"), "invoke mpi",
+                                    detail::stream_name(r.op_state.stream_), ptr(request)));
 
-                                    PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
-                                        "MPI_REQUEST_NULL returned from mpi invocation");
-                                },
-                                t);
+                            PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
+                                "MPI_REQUEST_NULL returned from mpi invocation");
+
                             if (status != MPI_SUCCESS)
                             {
                                 PIKA_DETAIL_DP(mpi_tran<5>,
@@ -169,8 +159,7 @@ namespace pika::mpi::experimental::detail {
                                     std::make_exception_ptr(mpi_exception(status)));
                                 return;
                             }
-                            // function called, Ts... can now be released (if refs hold lifetime)
-                            r.op_state.ts = {};
+
                             if (poll_request(request))
                             {
                                 PIKA_DETAIL_DP(mpi_tran<7>,
@@ -207,23 +196,6 @@ namespace pika::mpi::experimental::detail {
             {
                 using type = pud::transform_t<Tuple, std::decay>;
             };
-
-#if defined(PIKA_HAVE_STDEXEC)
-            using ts_type = pika::util::detail::prepend_t<
-                pika::util::detail::transform_t<
-                    execution::experimental::value_types_of_t<std::decay_t<Sender>,
-                        execution::experimental::empty_env, std::tuple, pika::detail::variant>,
-                    value_types_helper>,
-                pika::detail::monostate>;
-#else
-            using ts_type = pika::util::detail::prepend_t<
-                pika::util::detail::transform_t<
-                    typename execution::experimental::sender_traits<std::decay_t<Sender>>::
-                        template value_types<std::tuple, pika::detail::variant>,
-                    value_types_helper>,
-                pika::detail::monostate>;
-#endif
-            ts_type ts;
 
             // we don't bother storing the result, but the correct type is ...
             using result_type = pika::detail::variant<std::tuple<MPI_Request>>;
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -1,0 +1,230 @@
+//  Copyright (c) 2023 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+/// \file parallel/algorithms/transform_xxx.hpp
+
+#pragma once
+
+#include <pika/config.hpp>
+#include <pika/assert.hpp>
+#include <pika/async_mpi/mpi_helpers.hpp>
+#include <pika/async_mpi/mpi_polling.hpp>
+#include <pika/concepts/concepts.hpp>
+#include <pika/datastructures/variant.hpp>
+#include <pika/debugging/demangle_helper.hpp>
+#include <pika/debugging/print.hpp>
+#include <pika/execution/algorithms/detail/helpers.hpp>
+#include <pika/execution/algorithms/detail/partial_algorithm.hpp>
+#include <pika/execution/algorithms/transfer.hpp>
+#include <pika/execution_base/any_sender.hpp>
+#include <pika/execution_base/receiver.hpp>
+#include <pika/execution_base/sender.hpp>
+#include <pika/executors/inline_scheduler.hpp>
+#include <pika/executors/limiting_scheduler.hpp>
+#include <pika/executors/thread_pool_scheduler.hpp>
+#include <pika/functional/detail/tag_fallback_invoke.hpp>
+#include <pika/functional/invoke.hpp>
+#include <pika/functional/invoke_fused.hpp>
+#include <pika/mpi_base/mpi.hpp>
+
+#include <exception>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+
+namespace pika::mpi::experimental::transform_mpi_detail {
+
+    // -----------------------------------------------------------------
+    // route calls through an impl layer for ADL resolution
+    template <typename Sender, typename F>
+    struct dispatch_mpi_impl
+    {
+        struct dispatch_mpi_type;
+    };
+
+    template <typename Sender, typename F>
+    using dispatch_mpi = typename dispatch_mpi_impl<Sender, F>::dispatch_mpi_type;
+
+    // -----------------------------------------------------------------
+    // transform MPI adapter - sender type
+    template <typename Sender, typename F>
+    struct dispatch_mpi_impl<Sender, F>::dispatch_mpi_type
+    {
+        using is_sender = void;
+
+        std::decay_t<Sender> sender;
+        std::decay_t<F> f;
+        stream_type stream;
+
+        // -----------------------------------------------------------------
+        // completion signatures
+        template <template <typename...> class Tuple, template <typename...> class Variant>
+        using value_types = Variant<Tuple<MPI_Request>>;
+
+        template <template <typename...> class Variant>
+        using error_types = pud::unique_t<
+            pud::prepend_t<typename exp::sender_traits<Sender>::template error_types<Variant>,
+                std::exception_ptr>>;
+
+        static constexpr bool sends_done = false;
+
+        // -----------------------------------------------------------------
+        // operation state for a internal receiver
+        template <typename Receiver>
+        struct operation_state
+        {
+            std::decay_t<Receiver> receiver;
+            std::decay_t<F> f;
+            stream_type stream_;
+
+            // -----------------------------------------------------------------
+            // The mpi_receiver receives inputs from the previous sender,
+            // invokes the mpi call, and sets a callback on the polling handler
+            struct dispatch_mpi_receiver
+            {
+                using is_receiver = void;
+
+                operation_state& op_state;
+
+                template <typename Error>
+                friend constexpr void
+                tag_invoke(exp::set_error_t, dispatch_mpi_receiver&& r, Error&& error) noexcept
+                {
+                    exp::set_error(PIKA_MOVE(r.op_state.receiver), PIKA_FORWARD(Error, error));
+                }
+
+                friend constexpr void tag_invoke(
+                    exp::set_stopped_t, dispatch_mpi_receiver&& r) noexcept
+                {
+                    exp::set_stopped(PIKA_MOVE(r.op_state.receiver));
+                }
+
+                // receive the MPI function invocable + arguments and add a request,
+                // then invoke the mpi function with the added request
+                // if the invocation gives an error, set_error
+                // otherwise return the request by passing it to set_value
+                template <typename... Ts,
+                    typename = std::enable_if_t<is_mpi_request_invocable_v<F, Ts...>>>
+                friend constexpr void
+                tag_invoke(exp::set_value_t, dispatch_mpi_receiver&& r, Ts&&... ts) noexcept
+                {
+                    pika::detail::try_catch_exception_ptr(
+                        [&]() mutable {
+                            using namespace pika::debug::detail;
+                            using ts_element_type = std::tuple<std::decay_t<Ts>...>;
+                            using invoke_result_type = mpi_request_invoke_result_t<F, Ts...>;
+
+                            // move/copy the received params into our local opstate
+                            r.op_state.ts.template emplace<ts_element_type>(
+                                PIKA_FORWARD(Ts, ts)...);
+                            // and get a reference to the tuple of local params
+                            auto& t = std::get<ts_element_type>(r.op_state.ts);
+                            // init a request
+                            MPI_Request request{MPI_REQUEST_NULL};
+                            int status = MPI_SUCCESS;
+                            // invoke the function, with the contents of the param tuple as args
+                            pud::invoke_fused(
+                                [&](auto&... ts) mutable {
+                                    // execute the mpi function call, passing in the request object
+                                    if constexpr (std::is_void_v<invoke_result_type>)
+                                    {
+                                        PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
+                                        PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
+                                            "MPI_REQUEST_NULL returned from mpi "
+                                            "invocation");
+                                    }
+                                    else
+                                    {
+                                        status =
+                                            PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
+                                        PIKA_ASSERT_MSG(request != MPI_REQUEST_NULL,
+                                            "MPI_REQUEST_NULL returned from mpi "
+                                            "invocation");
+                                    }
+                                },
+                                t);
+                            // function called, Ts... can now be released (if refs hold lifetime)
+                            r.op_state.ts = {};
+                            // calls set_value(request), or set_error(mpi_exception(status))
+                            set_value_error_helper(status, PIKA_MOVE(r.op_state.receiver), request);
+                        },
+                        [&](std::exception_ptr ep) {
+                            exp::set_error(PIKA_MOVE(r.op_state.receiver), PIKA_MOVE(ep));
+                        });
+                }
+
+                friend constexpr exp::empty_env tag_invoke(
+                    exp::get_env_t, dispatch_mpi_receiver const&) noexcept
+                {
+                    return {};
+                }
+            };
+
+            using operation_state_type =
+                exp::connect_result_t<std::decay_t<Sender>, dispatch_mpi_receiver>;
+            operation_state_type op_state;
+
+            template <typename Tuple>
+            struct value_types_helper
+            {
+                using type = pud::transform_t<Tuple, std::decay>;
+            };
+
+#if defined(PIKA_HAVE_STDEXEC)
+            using ts_type = pika::util::detail::prepend_t<
+                pika::util::detail::transform_t<
+                    execution::experimental::value_types_of_t<std::decay_t<Sender>,
+                        execution::experimental::empty_env, std::tuple, pika::detail::variant>,
+                    value_types_helper>,
+                pika::detail::monostate>;
+#else
+            using ts_type = pika::util::detail::prepend_t<
+                pika::util::detail::transform_t<
+                    typename execution::experimental::sender_traits<std::decay_t<Sender>>::
+                        template value_types<std::tuple, pika::detail::variant>,
+                    value_types_helper>,
+                pika::detail::monostate>;
+#endif
+            ts_type ts;
+
+            using result_type = pika::detail::variant<std::tuple<MPI_Request>>;
+            result_type result;
+
+            template <typename Receiver_, typename F_, typename Sender_>
+            operation_state(Receiver_&& receiver, F_&& f, Sender_&& sender, stream_type s)
+              : receiver(PIKA_FORWARD(Receiver_, receiver))
+              , f(PIKA_FORWARD(F_, f))
+              , stream_{s}
+              , op_state(exp::connect(PIKA_FORWARD(Sender_, sender), dispatch_mpi_receiver{*this}))
+            {
+                PIKA_DETAIL_DP(mpi_tran,
+                    debug(
+                        debug::detail::str<>("operation_state"), "stream", detail::stream_name(s)));
+            }
+
+            friend constexpr auto tag_invoke(exp::start_t, operation_state& os) noexcept
+            {
+                return exp::start(os.op_state);
+            }
+        };
+
+        template <typename Receiver>
+        friend constexpr auto
+        tag_invoke(exp::connect_t, dispatch_mpi_type const& s, Receiver&& receiver)
+        {
+            return operation_state<Receiver>(
+                PIKA_FORWARD(Receiver, receiver), s.f, s.sender, s.stream);
+        }
+
+        template <typename Receiver>
+        friend constexpr auto tag_invoke(exp::connect_t, dispatch_mpi_type&& s, Receiver&& receiver)
+        {
+            return operation_state<Receiver>(
+                PIKA_FORWARD(Receiver, receiver), PIKA_MOVE(s.f), PIKA_MOVE(s.sender), s.stream);
+        }
+    };
+
+}    // namespace pika::mpi::experimental::transform_mpi_detail
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -35,11 +35,12 @@
 
 namespace pika::mpi::experimental::detail {
 
-    namespace pud = pika::util::detail;
-    namespace exp = execution::experimental;
+    namespace {
+        namespace ex = execution::experimental;
+    }
 
     // -----------------------------------------------------------------
-    // route calls through an impl layer for ADL resolution
+    // route calls through an impl layer for ADL isolation
     template <typename Sender, typename F>
     struct dispatch_mpi_sender_impl
     {
@@ -73,16 +74,15 @@ namespace pika::mpi::experimental::detail {
         using value_types = Variant<Tuple<MPI_Request>>;
 
         template <template <typename...> class Variant>
-        using error_types = pud::unique_t<
-            pud::prepend_t<typename exp::sender_traits<Sender>::template error_types<Variant>,
-                std::exception_ptr>>;
+        using error_types = util::detail::unique_t<util::detail::prepend_t<
+            typename ex::sender_traits<Sender>::template error_types<Variant>, std::exception_ptr>>;
 
 #endif
 
         static constexpr bool sends_done = false;
 
         // -----------------------------------------------------------------
-        // operation state for a internal receiver
+        // operation state for an internal receiver
         template <typename Receiver>
         struct operation_state
         {
@@ -101,15 +101,15 @@ namespace pika::mpi::experimental::detail {
 
                 template <typename Error>
                 friend constexpr void
-                tag_invoke(exp::set_error_t, dispatch_mpi_receiver&& r, Error&& error) noexcept
+                tag_invoke(ex::set_error_t, dispatch_mpi_receiver r, Error&& error) noexcept
                 {
-                    exp::set_error(PIKA_MOVE(r.op_state.receiver), PIKA_FORWARD(Error, error));
+                    ex::set_error(PIKA_MOVE(r.op_state.receiver), PIKA_FORWARD(Error, error));
                 }
 
                 friend constexpr void tag_invoke(
-                    exp::set_stopped_t, dispatch_mpi_receiver&& r) noexcept
+                    ex::set_stopped_t, dispatch_mpi_receiver r) noexcept
                 {
-                    exp::set_stopped(PIKA_MOVE(r.op_state.receiver));
+                    ex::set_stopped(PIKA_MOVE(r.op_state.receiver));
                 }
 
                 // receive the MPI function invocable + arguments and add a request,
@@ -119,13 +119,14 @@ namespace pika::mpi::experimental::detail {
                 template <typename... Ts,
                     typename = std::enable_if_t<is_mpi_request_invocable_v<F, Ts...>>>
                 friend constexpr void
-                tag_invoke(exp::set_value_t, dispatch_mpi_receiver&& r, Ts&&... ts) noexcept
+                tag_invoke(ex::set_value_t, dispatch_mpi_receiver r, Ts&&... ts) noexcept
                 {
                     pika::detail::try_catch_exception_ptr(
                         [&]() mutable {
                             using namespace pika::debug::detail;
                             using ts_element_type = std::tuple<std::decay_t<Ts>...>;
                             using invoke_result_type = mpi_request_invoke_result_t<F, Ts...>;
+
                             using namespace pika::debug::detail;
                             PIKA_DETAIL_DP(mpi_tran<5>,
                                 debug(str<>("dispatch_mpi_recv"), "set_value_t", "stream",
@@ -141,6 +142,7 @@ namespace pika::mpi::experimental::detail {
                             }
                             else
                             {
+                                static_assert(std::is_same_v<invoke_result_type, int>);
                                 status = PIKA_INVOKE(PIKA_MOVE(r.op_state.f), ts..., &request);
                             }
                             PIKA_DETAIL_DP(mpi_tran<7>,
@@ -155,7 +157,7 @@ namespace pika::mpi::experimental::detail {
                                 PIKA_DETAIL_DP(mpi_tran<5>,
                                     debug(str<>("set_error"), "status != MPI_SUCCESS",
                                         detail::error_message(status)));
-                                exp::set_error(PIKA_MOVE(r.op_state.receiver),
+                                ex::set_error(PIKA_MOVE(r.op_state.receiver),
                                     std::make_exception_ptr(mpi_exception(status)));
                                 return;
                             }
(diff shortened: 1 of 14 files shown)


biddisco:
Note that this PR includes a tweaked (100 column) `.clang-format` file in the async_mpi lib folder, and the pre-commit hooks are updated to cd into the right dir before clang-format checking - but the CI probably does not manage this and should be investigated. If not easy to do, then the clang-format changes can be reverted and the files reformatted.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -43,61 +52,94 @@ namespace pika::mpi::experimental {
...
         template <typename F, typename... Ts>
-        using mpi_request_invoke_result_t =
-            std::decay_t<pika::util::detail::invoke_result_t<F,
-                std::add_lvalue_reference_t<std::decay_t<Ts>>...,
-                MPI_Request*>>;
+        using mpi_request_invoke_result_t = std::decay_t<pika::util::detail::invoke_result_t<F,
+            std::add_lvalue_reference_t<std::decay_t<Ts>>..., MPI_Request*>>;
 
+        // -----------------------------------------------------------------
+        // route calls through an impl layer for ADL resolution
```
msimberg, line 131:
```suggestion
        // route calls through an impl layer for ADL isolation
```
?

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -43,61 +52,94 @@ namespace pika::mpi::experimental {
...
         {
             struct transform_mpi_sender_type;
         };
 
         template <typename Sender, typename F>
-        using transform_mpi_sender = typename transform_mpi_sender_impl<Sender,
-            F>::transform_mpi_sender_type;
+        using transform_mpi_sender = typename transform_mpi_sender_impl<Sender, F>::transform_mpi_sender_type;
 
+        // -----------------------------------------------------------------
```
msimberg, line 141: I know this is your preferred style and I prefer comments without these lines, and that this is purely subjective, but can we do without these lines? There's already whitespace to separate things, and the comment also acts as a separator, no? We don't use this style elsewhere.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -109,198 +151,259 @@ namespace pika::mpi::experimental {
...
+                typename pika::execution::experimental::sender_traits<Sender>::template error_types<Variant>,
+                std::exception_ptr>>;
 
             static constexpr bool sends_done = false;
 #endif
 
+            struct priority_set_restore
+            {
+                pika::execution::thread_priority old_priority_;
+                priority_set_restore(pika::execution::thread_priority new_p)
```
msimberg, line 198:
```suggestion
                priority_set_restore(pika::execution::thread_priority new_priority)
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -109,198 +151,259 @@ namespace pika::mpi::experimental {
...
+            // operation state for a given receiver
             template <typename Receiver>
             struct operation_state
             {
                 std::decay_t<Receiver> receiver;
                 std::decay_t<F> f;
                 stream_type stream;
-
+                pika::spinlock mutex_;
+                pika::condition_variable cond_var_;
```
msimberg, line 219: I know this is not enforced across the codebase, but let's try to keep it consistent inside a single class at least:
```suggestion
                pika::spinlock mutex;
                pika::condition_variable cond_var;
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -336,100 +439,306 @@ namespace pika::mpi::experimental {
...
+                    transform_mpi_sender<Sender, F>{PIKA_FORWARD(Sender, sender), PIKA_FORWARD(F, f), s},
+                    pika::execution::experimental::get_completion_scheduler<
+                        pika::execution::experimental::set_value_t>(sender));
+            }
+            else
+            {
+                namespace ex = pika::execution::experimental;
+                namespace mpi = pika::mpi::experimental;
+                auto mode = mpi::get_completion_mode();
+                if (mode == 0)
```
msimberg, line 524: Could we name the modes (with an enum) instead?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -213,106 +352,107 @@ namespace pika::mpi::experimental {
...
                 "MPI event polling has not been enabled on any pool. Make sure "
                 "that MPI event polling is enabled on at least one thread "
                 "pool.");
 
-            // Eagerly check if request already completed. If it did, call the
-            // callback immediately.
-            int flag = 0;
-            int result = MPI_Test(&request, &flag, MPI_STATUS_IGNORE);
-            if (flag)
+            // if already complete, kip callback
```
msimberg, line 402:
```suggestion
            // if already complete, skip callback
```

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -4,19 +4,20 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-set(tests algorithm_transform_mpi mpi_ring_async_sender_receiver
-          mpi_async_storage
+set(tests # algorithm_transform_mpi
+    mpi_ring_async_sender_receiver
+    # mpi_async_storage
```
msimberg, line 9: Uncomment others before merging.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -4,19 +4,20 @@
...
 set(mpi_ring_async_sender_receiver_PARAMETERS
-    ARGS "--in-flight-limit=32"
+    ARGS "--in-flight-limit=32" "--rounds=10" "--iterations=100" "--pika:ignore-process-mask"
     THREADS 4 LOCALITIES 2 RUNWRAPPER mpi
 )
 
 set(mpi_async_storage_PARAMETERS
     ARGS
-        "--in-flight-limit=256" "--localMB=256" "--transferKB=1024" "--seconds=1"
+        "--in-flight-limit=256" "--localMB=256" "--transferKB=1024" "--seconds=1" "-Ipika.stacks.use_guard_pages=0" "--pika:ignore-process-mask"
```
msimberg, line 20: Are the added flags just for testing?

libs/pika/thread_pools/include/pika/thread_pools/scheduling_loop.hpp
```diff
@@ -662,8 +662,8 @@ namespace pika::threads::detail {
                         num_thread, running, thrd, enable_stealing)))
             {
                 tfunc_time_wrapper tfunc_time_collector(idle_rate);
-                PIKA_ASSERT(get_thread_id_data(thrd)->get_scheduler_base() ==
-                    &scheduler);
+                //                PIKA_ASSERT(get_thread_id_data(thrd)->get_scheduler_base() ==
+                //                    &scheduler);
```
msimberg, line 666: This should probably be uncommented again, or we need to update _something_ (change the assert, add asserts elsewhere, etc.).

msimberg (commented):
I only did a quick superficial pass over this, let's discuss in more detail in the meeting.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -26,32 +26,33 @@
...
     enum class stream_type : std::uint32_t
     {
         automatic = 0,
-        send,
-        receive,
-        collective,
-        user,
+        send_1,
+        send_2,
+        receive_1,
```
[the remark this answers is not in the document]
biddisco: I am using them in testing. Since we are currently using the namespace `pika::mpi::experimental` I feel like it ought to be ok to leave some experimental features in until the API has been battle tested, and final decisions made as to what is good and what isn't.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -26,32 +26,33 @@
...
+        receive_2,
+        collective_1,
+        collective_2,
+        user_1,
+        user_2,
+        max_stream
```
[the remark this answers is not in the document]
biddisco: if you mean, `max_stream` its just a sentinel to use in array sizes of how many streams to manage

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
biddisco: yes, an enum could be used, but then the ENV var setting would probably need to be changed to use string or something. Not sure it makes anything better yet. Same comment about `experimental` features applies here.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: The other tests are currently broken.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: `rounds` and `iterations` are part of the test, stack guard pages help prevent segfaults and the ignore process mask is because the default `mpiexec` causes warnings since threads are not setup unless we add a bunch of other openmpi/other flags on machines like my laptop

libs/pika/thread_pools/include/pika/thread_pools/scheduling_loop.hpp
biddisco: I had to uncomment this to get the bypass scheduler to work, but I moved all the bypass scheduler stuff into another PR that is waiting, so I'll uncomment this for now

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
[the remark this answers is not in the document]
msimberg: Yeah, sorry, that's what I mean. I.e. `max_stream` is not a stream type that the user would choose for their calls. I'm just wondering if we have a better name for it... but this does not block anything and can be changed separately, so I'll mark this as resolved.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
msimberg: Fair enough, let's keep it like this for now.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg: `rounds` and `iterations` are fine.

`pika.stacks.use_guard_pages`: do you know why this particular test needs this? Is this something we should consider disabling by default anyway? Or enabling only in debug mode or something like that?

`--pika:ignore-process-mask`: it'd be nice if we could control this with a CMake variable instead (either a flag, or an "extra flags" option).

msimberg:
bors try

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg: By the way, I take your comment as an explanation for why they're uncommented, but I assume they will be updated in this PR?

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -39,12 +39,11 @@
 
 // a debug level of zero disables messages with a priority>0
 // a debug level of N shows messages with priority<N
-constexpr int debug_level = 0;
+constexpr int debug_level = 9;
```
msimberg, line 42: This means debugging is enabled, right? Should this stay or is it just temporary?

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -62,6 +61,18 @@ namespace mpi = pika::mpi::experimental;
...
+    template <typename Sender>
+    auto async(Sender&& sender)
+    {
+        namespace ex = pika::execution::experimental;
+        auto sched = ex::thread_pool_scheduler{&pika::resource::get_thread_pool("default")};
+        auto snd = ex::schedule(sched) |
+            ex::then([sender = std::move(sender)]() mutable { ex::start_detached(std::move(sender)); });
+        ex::start_detached(std::move(snd));
+    }
+}    // namespace pika::execution
```
msimberg, line 74: I think it'd be clearer if you just replace the `just` with a `transfer_just` where you're using this right now. At least I don't like using the `pika::execution::async` name for this since it doesn't behave the same as `pika::async`.

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -6,12 +6,19 @@
 
 #include <pika/assert.hpp>
 #include <pika/execution.hpp>
+#if __has_include(<pika/executors/thread_pool_scheduler_queue_bypass.hpp>)
+#include <pika/executors/thread_pool_scheduler_queue_bypass.hpp>
+#endif
```
msimberg, line 11: This doesn't belong here, right?

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -40,169 +58,425 @@ namespace mpi = pika::mpi::experimental;
...
+}
 
-void msg_recv(int rank, int size, int /*to*/, int from, int token, unsigned tag)
+// ------------------------------------------------------------
+// when messages arrive/complete in random order, it might be the
+// case that message N + x completes long before message N, so
+// we don't reuse tags immediately and instead offset them by
+// a multiple to ensure we don't have two with the same id in flight
+// at the same time
+inline std::uint32_t tag_no(std::uint64_t rank, std::uint64_t iteration, std::uint32_t ranks)
```
msimberg, line 110: Maybe?
```suggestion
inline std::uint32_t tag(_num(ber))(std::uint64_t rank, std::uint64_t iteration, std::uint32_t ranks)
```
`tag_no` sounds pretty negative to me... :wink:

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -40,169 +58,425 @@ namespace mpi = pika::mpi::experimental;
...
+    template <typename Sender>
+    auto async(Sender&& sender)
+    {
+        namespace ex = pika::execution::experimental;
+        auto sched = ex::thread_pool_scheduler{&pika::resource::get_thread_pool("default")};
+        auto snd = ex::schedule(sched) |
+            ex::then([sender = std::move(sender)]() mutable { ex::start_detached(std::move(sender)); });
+        ex::start_detached(std::move(snd));
+    }
+}    // namespace pika::execution
```
msimberg, line 190: Same comment here as in the other test. If you really want a helper like this, call it maybe `on`, give it a scheduler as the first parameter, and just put it in the global namespace for this test. Also:
```suggestion
namespace pika::execution {
    template <typename Sender>
    auto async(Sender&& sender)
    {
        namespace ex = pika::execution::experimental;
        auto sched = ex::thread_pool_scheduler{&pika::resource::get_thread_pool("default")};
        auto snd = ex::schedule(sched) |
            ex::then([sender = std::forward<Sender>(sender)]() mutable { ex::start_detached(std::move(sender)); });
        ex::start_detached(std::move(snd));
    }
}    // namespace pika::execution
```

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -40,169 +58,425 @@ namespace mpi = pika::mpi::experimental;
...
+
+struct receiver
```
msimberg, line 192: Is it confusing to call this `receiver` in the context of sender/receiver?

libs/pika/threading_base/src/print.cpp
```diff
@@ -11,6 +11,12 @@
 #include <cstdint>
 #include <thread>
 
+#if defined(__linux) || defined(linux) || defined(__linux__)
+#include <linux/unistd.h>
+#include <sys/mman.h>
+#define DEBUGGING_PRINT_LINUX
```
msimberg, line 17:
```suggestion
#define PIKA_DEBUGGING_PRINT_LINUX
```

tools/pre-commit
```diff
@@ -31,9 +31,13 @@ normal=$(tput sgr0)
 
 cxxfiles=()
 for file in `git diff --cached --name-only --diff-filter=ACMRT | grep -E "\.(cpp|hpp)$"`; do
+    # to allow for per-directory clang format files, we cd into the dir first
+    DIR=$(dirname "$file")
+    pushd $DIR > /dev/null
     if ! cmp -s <(git show :${file}) <(git show :${file}|$CLANG_FORMAT_VERSION -style=file); then
         cxxfiles+=("${file}")
     fi
+    popd > /dev/null
```
msimberg, line 40: We don't need these changes anymore, do we?

msimberg (changes requested):

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -30,9 +33,15 @@
 
 namespace pika::mpi::experimental {
     namespace transform_mpi_detail {
```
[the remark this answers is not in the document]
msimberg: See https://github.com/pika-org/pika/issues/600.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -109,198 +151,259 @@ namespace pika::mpi::experimental {
...
+                                                pika::execution::thread_priority::high);
+                                            std::unique_lock l{r.op_state.mutex_};
+                                            r.op_state.cond_var_.wait(l, [&]() { return r.op_state.resume; });
+                                        }
+                                    }
+                                    r.op_state.ts = {};
+                                    r.op_state.status = MPI_SUCCESS;
+                                    if constexpr (!std::is_void_v<invoke_result_type>)
+                                    {
+                                        // @todo - use a helper like other modes
```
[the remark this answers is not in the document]
biddisco: no. it's a very small constexpr if and not worth a struct

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -336,100 +439,306 @@ namespace pika::mpi::experimental {
...
+                                    mpi::get_pool_name())},
+                            pika::execution::thread_stacksize::nostack));
+                    auto snd1 =
+                        transform_mpi_sender<decltype(snd0), F>{
+                            std::move(snd0), PIKA_FORWARD(F, f), s} |
+                        ex::transfer(ex::thread_pool_scheduler_queue_bypass{
+                            &pika::resource::get_thread_pool("default")});
+                    return ex::make_unique_any_sender(std::move(snd1));
+                }
+*/
```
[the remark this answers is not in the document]
biddisco: I've left the bypass scheduler stuff commented, but reactivated the others. once the bypass stuff is in or out, I'll redo this section.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -336,100 +439,306 @@ namespace pika::mpi::experimental {
...
+
+            if constexpr (pika::execution::experimental::detail::has_completion_scheduler_v<
+                              pika::execution::experimental::set_value_t, std::decay_t<Sender>>)
+            {
+                return pika::execution::experimental::transfer(
+                    transform_mpi_sender<Sender, F>{PIKA_FORWARD(Sender, sender), PIKA_FORWARD(F, f), s},
+                    pika::execution::experimental::get_completion_scheduler<
+                        pika::execution::experimental::set_value_t>(sender));
+            }
+            else
```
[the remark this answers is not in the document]
biddisco: indeed. That's just an unfinished hangover from the original implementation. I will add in the right transfer back to the completion scheduler in another commit.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -124,20 +185,51 @@ namespace pika::mpi::experimental {
         /// a single instance of all the mpi variables initialized once at startup
         static mpi_data mpi_data_;
 
+        // default transfer mode for mpi continuations
+        static std::size_t task_completion_mode_ = get_completion_mode_default();
+
+        static std::string polling_pool_name = "pika:polling";
```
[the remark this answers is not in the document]
biddisco: the "pika" in the name is so that when users dump out thread bindings, they see that this was a pool created by pika, and not one of their own pools (if they created some), but to be fair, it's fine without the pika and I'll get rid of it.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -124,20 +185,51 @@ namespace pika::mpi::experimental {
...
+
+        inline mpi_stream& get_stream_ref(stream_type stream)
```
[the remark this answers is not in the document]
biddisco: No. The contents are modified. it's just an accessor for convenience.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -124,20 +185,51 @@ namespace pika::mpi::experimental {
...
+
+        // stream operator to display debug mpi_request
+        PIKA_EXPORT std::ostream& operator<<(std::ostream& os, MPI_Request const& req)
+        {
+            using namespace pika::debug::detail;
+            // clang-format off
+            os <<  "req " << hex<8>(req);
             // clang-format on
             return os;
         }
```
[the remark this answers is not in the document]
biddisco: I must upgrade the whole debug/print framework to use fmt. That will be a new PR entirely I think.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -167,42 +260,89 @@ namespace pika::mpi::experimental {
...
+            // we do not throttle if we are not on a pika thread
+            using namespace pika::threads::detail;
+            if (get_self_id() != invalid_thread_id)
             {
                 std::unique_lock lk(stream.throttling_mtx_);
-                [[maybe_unused]] auto scp =
-                    mpi_debug.scope("throttling", "wait");
+                [[maybe_unused]] auto scp = mpi_debug<1>.scope("throttling", "wait", "locked", stream);
                 stream.throttling_cond_.wait(lk);
             }
```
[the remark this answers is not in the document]
biddisco: I've temporarily set `constexpr bool enable_throttling = false`, so the throttling is right now disable. But I will test using a counting semaphore and add a commit to the PR later

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -167,42 +260,89 @@ namespace pika::mpi::experimental {
...
+                {
+                    val = def;
+                    std::cerr << str<>("get_env_value") << " invalid " << s << " " << dec<4>(def)
+                              << std::endl;
+                    PIKA_DP(mpi_debug<3>, error(str<>("get_env_value"), "invalid", s, dec<4>(val)));
+                }
+                PIKA_DP(mpi_debug<3>, debug(str<>("get_env_value"), s, dec<4>(val)));
             }
+            return val;
+        }
```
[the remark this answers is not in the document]
biddisco: ok

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
biddisco: 0 means messages are squashed. 9 means all messages from 0-9 are shown. Mostly I set important messages to low numbers, and less important to high numbers, then set debug level to around 4 or 5 - unless I need to see everything. A value of 0 turn stuff off, unless of course you instantiate a debug object of level 0, but I only use 0 for critical messages that are not really debug only. (c.f. warning, debug, trace, type levels but numerical)

libs/pika/async_mpi/src/mpi_polling.cpp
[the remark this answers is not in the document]
msimberg: https://github.com/pika-org/pika/issues/610

libs/pika/async_mpi/src/mpi_polling.cpp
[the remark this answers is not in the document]
msimberg: Where is that set? Could you please push if it's unpushed?

https://github.com/pika-org/pika/issues/611

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -167,42 +260,89 @@ namespace pika::mpi::experimental {
...
         // -----------------------------------------------------------------
         void wait_for_throttling(stream_type stream)
         {
-            wait_for_throttling_impl(
-                mpi_data_.default_queues_[static_cast<uint32_t>(stream)]);
+            // if throttling is disabled, then do nothing
+            if constexpr (detail::throttling_enabled)
+            {
+                wait_for_throttling_impl(get_stream_ref(stream));
+            }
```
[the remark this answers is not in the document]
msimberg: Is there any reason for not changing this?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -167,42 +260,89 @@ namespace pika::mpi::experimental {
...
+                mpi_data_.default_queues_[i].limit_ = val;
+                mpi_data_.default_queues_[i].index_ = i;
+            }
+            // check env settings for individual streams
             for (size_t i = 0; i < mpi_data_.default_queues_.size(); ++i)
             {
-                mpi_data_.default_queues_[i].index = i;
+                std::string str = "PIKA_MPI_MSG_THROTTLE_" + std::string(stream_name(stream_type(i)));
+                std::transform(
+                    str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::toupper(c); });
```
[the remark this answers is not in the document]
msimberg: Is this change unpushed?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -213,106 +353,107 @@ namespace pika::mpi::experimental {
...
                 "pool.");
 
-            // Eagerly check if request already completed. If it did, call the
-            // callback immediately.
-            int flag = 0;
-            int result = MPI_Test(&request, &flag, MPI_STATUS_IGNORE);
-            if (flag)
+            // if already complete, skip callback
+#ifndef DISALLOW_EAGER_POLLING_CHECK
+            if (eager_check && detail::poll_request(request))
```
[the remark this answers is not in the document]
msimberg: Is this unpushed?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -213,106 +353,107 @@ namespace pika::mpi::experimental {
...
 
-            MPI_Comm_create_errhandler(
-                detail::pika_MPI_Handler, &detail::pika_mpi_errhandler);
-            MPI_Comm_set_errhandler(
-                MPI_COMM_WORLD, detail::pika_mpi_errhandler);
+            MPI_Comm_create_errhandler(detail::pika_MPI_Handler, &detail::pika_mpi_errhandler);
+            MPI_Comm_set_errhandler(MPI_COMM_WORLD, detail::pika_mpi_errhandler);
         }
 
+        bool poll_request(MPI_Request& req)
```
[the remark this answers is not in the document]
msimberg: Please mark this as resolved if you're not planning on changing the name.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -345,132 +485,177 @@ namespace pika::mpi::experimental {
...
+                    detail::request_callback req_callback;
+                    while (detail::mpi_data_.request_callback_queue_.try_dequeue(req_callback))
+                    {
+                        --detail::mpi_data_.request_queue_size_;
+                        add_to_request_callback_vector(PIKA_MOVE(req_callback));
+                    }
+
+                    const std::size_t max_test_vector_size = 512;
+                    std::size_t vsize =
+                        std::min(detail::mpi_data_.request_vector_.size(), max_test_vector_size);
```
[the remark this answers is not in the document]
msimberg: Unpushed?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -345,132 +485,177 @@ namespace pika::mpi::experimental {
...
+                    if (detail::mpi_data_.max_polling_requests > 1)
+                    {
+                        std::size_t req_size = std::min(vsize, detail::mpi_data_.max_polling_requests);
+                        detail::mpi_data_.indices_vector_.resize(req_size);
+                        detail::mpi_data_.status_vector_.resize(req_size);
+
+                        int result = MPI_Testsome(req_size, detail::mpi_data_.request_vector_.data(),
+                            &outcount, detail::mpi_data_.indices_vector_.data(),
+                            detail::mpi_data_.status_vector_.data());
+                        /*use MPI_STATUSES_IGNORE ?*/
```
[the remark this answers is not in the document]
msimberg: Please remove or open an issue if you're not planning to change this in this PR.

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
msimberg: Please at least rename this to almost anything but `async`. Maybe `start_detached_on`?

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg: https://github.com/pika-org/pika/issues/612

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg: - https://github.com/pika-org/pika/issues/613
- https://github.com/pika-org/pika/issues/614

libs/pika/threading_base/include/pika/threading_base/thread_data.hpp
```diff
@@ -433,6 +433,24 @@ namespace pika::threads::detail {
...
+        // convenience object to temporarily change priority
+        struct scoped_thread_priority
+        {
+            pika::execution::thread_priority old_priority_;
+            scoped_thread_priority(pika::execution::thread_priority new_p)
+              : old_priority_{
+                    threads::detail::get_self_id_data()->get_priority()}
+            {
+                threads::detail::get_self_id_data()->set_priority(new_p);
+            }
```
msimberg, line 445:
```suggestion
            explicit scoped_thread_priority(pika::execution::thread_priority new_priority)
              : old_priority_{
                    threads::detail::get_self_id_data()->get_priority()}
            {
                threads::detail::get_self_id_data()->set_priority(new_priority);
            }
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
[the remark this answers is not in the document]
msimberg: This is still unresolved. Do you have a commit that fixes this?

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -0,0 +1,208 @@
...
+    // return a scheduler on the default pool with added priority if requested
+    inline auto default_pool_scheduler(execution::thread_priority p)
+    {
+        if (p == execution::thread_priority::normal)
+        {
+            return exp::thread_pool_scheduler{&resource::get_thread_pool("default")};
+        }
+        return exp::with_priority(
+            exp::thread_pool_scheduler{&resource::get_thread_pool("default")}, p);
+    }
```
msimberg, line 100: This can stay here since it's in `detail`, but we could consider moving this to somewhere a bit more prominent (most likely resource partitioner/runtime or similar). No changes requested.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -13,43 +13,68 @@
...
 #include <utility>
 #include <vector>
 
 namespace pika::mpi::experimental {
 
+    //    enum progress_mode
+    //    {
+    //        can_block,
+    //        cannot_block
+    //    };
```
msimberg, line 38:
```suggestion
```

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -13,43 +13,68 @@
...
+
+    template <typename E>
+    constexpr typename std::underlying_type<E>::type to_underlying(E e) noexcept
+    {
+        return static_cast<typename std::underlying_type<E>::type>(e);
```
msimberg, line 43:
```suggestion
    constexpr std::underlying_type_t<E> to_underlying(E e) noexcept
    {
        return static_cast<std::underlying_type_t<E>>(e);
```

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -72,38 +92,169 @@ namespace pika::mpi::experimental {
...
+        };
+
+        // 2 bits define continuation mode
+        inline handler_mode get_handler_mode(int flags)
+        {
+            return static_cast<handler_mode>(flags & to_underlying(handler_mode::method_mask));
+        }
+
+        // 1 bit defines pool or no pool
+        inline bool use_HP_com(int mode)
```
msimberg, line 149: Do you think we could expand `com` to `completion` here and elsewhere:
```suggestion
        inline bool use_HP_completion(int mode)
```
?
`com` is too close to `comm` in my opinion and we're not running out of characters.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -29,352 +33,109 @@
...
-        struct transform_mpi_sender_impl<Sender, F>::transform_mpi_sender_type
+    inline constexpr struct transform_mpi_t final
+      : pika::functional::detail::tag_fallback<transform_mpi_t>
+    {
+    private:
+        template <typename Sender, typename F,
+            PIKA_CONCEPT_REQUIRES_(exp::is_sender_v<std::decay_t<Sender>>)>
+        friend constexpr PIKA_FORCEINLINE exp::unique_any_sender<>
+        tag_fallback_invoke(transform_mpi_t, Sender&& sender, F&& f,
+            /*progress_mode p, */ stream_type s = stream_type::automatic)
```
msimberg, line 48: What is the `progress_mode` meant to do (eventually) and does it need to stay around?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -8,39 +8,94 @@
...
+#include <pika/synchronization/mutex.hpp>
+//
+// If we can use ranges/v3 then these can be used for zip vector compacting
+/*
+#include <range/v3/algorithm/copy.hpp>
+#include <range/v3/algorithm/remove_if.hpp>
+#include <range/v3/view/transform.hpp>
+#include <range/v3/view/zip.hpp>
+*/
+//
```
msimberg, line 28:
```suggestion
```
You can open an issue about this.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -8,39 +8,94 @@
...
+        // -----------------------------------------------------------------
+        /// To enable independent throttling of sends/receives/other
+        /// we maintain several "queues" which have their own condition
+        /// variables for suspension
+        struct mpi_stream
+        {
+            std::int32_t limit_;
+            const char* name_;
+            std::shared_ptr<semaphore_type> semaphore_;
+            semaphore_type* another_ptr_;
```
msimberg, line 84:
```suggestion
```

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -118,22 +193,81 @@ namespace pika::mpi::experimental {
...
         };
 
+        struct initializer
+        {
+            initializer()
+            {
+                init_throttling_default();
+            }
+        };
+
```
msimberg, line 203: What does this do? If there's no destructor you could just call `init_throttling_default` directly? But this also looks unused.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -118,22 +193,81 @@ namespace pika::mpi::experimental {
...
         static mpi_data mpi_data_;
 
+        void init_stream(stream_type s, std::int32_t limit)
+        {
+            mpi_stream& stream = mpi_data_.default_queues_[to_underlying(s)];
+            auto* sp = new semaphore_type(limit);
+            stream.semaphore_ = std::shared_ptr<semaphore_type>(sp, [](semaphore_type* p) {
+                // std::cout << "deleting shared pointer\n\n\n";
+                delete p;
+            });
```
msimberg, line 214:
```suggestion
            stream.semaphore_ = std::make_shared<semaphore_type>(limit);
```
?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -143,59 +277,56 @@ namespace pika::mpi::experimental {
...
+                std::transform(str.begin(), str.end(), str.begin(),
+                    [](unsigned char c) { return std::toupper(c); });
+                val = pika::detail::get_env_var_as<std::uint32_t>(str.c_str(), def);
+                if (val != def)
+                {
+                    detail::init_stream(stream_type(i), val);
+                }
             }
+            // return default val for automatic (unspecified) stream
+            //            return get_stream_ref(stream_type::automatic).limit_;
```
msimberg, line 310:
```suggestion
```
?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -226,87 +350,93 @@ namespace pika::mpi::experimental {
...
-            size_t const size = detail::mpi_data_.request_vector_.size();
+            using detail::mpi_data_;
+            /*
+            auto z = ranges::views::zip(mpi_data_.requests_, mpi_data_.callbacks_);
+            auto e = ranges::remove_if(z, [](auto&& r) { return r.first == MPI_REQUEST_NULL; });
+            mpi_data_.requests_.erase(
+                mpi_data_.requests_.begin() + (e - z.begin()), mpi_data_.requests_.end());
+            mpi_data_.callbacks_.erase(
+                mpi_data_.callbacks_.begin() + (e - z.begin()), mpi_data_.callbacks_.end());
+            */
```
msimberg, line 432:
```suggestion
```

libs/pika/coroutines/src/thread_enums.cpp
```diff
@@ -12,6 +12,7 @@
 #include <pika/coroutines/thread_enums.hpp>
 
 #include <cstddef>
+#include <iostream>
```
msimberg, line 15:
```suggestion
#include <ostream>
```

msimberg:
bors try

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -0,0 +1,298 @@
...
+            // invokes the mpi call, and sets a callback on the polling handler
+            struct dispatch_mpi_receiver
+            {
+                using is_receiver = void;
+
+                operation_state& op_state;
+
+                template <typename Error>
+                friend constexpr void
+                tag_invoke(exp::set_error_t, dispatch_mpi_receiver&& r, Error&& error) noexcept
```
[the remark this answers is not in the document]
biddisco: I've changed it to take by value in 6 places - dispatch*3 + trigger*3 - error/value/stopped

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -0,0 +1,298 @@
...
+                                        "MPI_REQUEST_NULL returned from mpi invocation");
+                                },
+                                t);
+                            if (status != MPI_SUCCESS)
+                            {
+                                PIKA_DETAIL_DP(mpi_tran<5>,
+                                    debug(str<>("set_error"), "status != MPI_SUCCESS",
+                                        detail::error_message(status)));
+                                exp::set_error(PIKA_MOVE(r.op_state.receiver),
+                                    std::make_exception_ptr(mpi_exception(status)));
```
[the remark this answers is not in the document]
biddisco: since there is already a check for `is_void`, it is enough to put the static assert `<int>` in the `else` part.
For now I'd prefer to leave the exceptions, they have proven to be useful so far finding bugs/errors as the mpi error message is displayed and if we only return an int status, then the user has to do the checking. I'll happily redo this for the next patch if necessary.

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -0,0 +1,298 @@
...
+                                return;
+                            }
+                            // function called, Ts... can now be released (if refs hold lifetime)
+                            r.op_state.ts = {};
```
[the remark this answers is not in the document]
biddisco: Removed all this arg copying (dlaf now has a `let_value` wrapper to keep tiles/etc alive). I will consider adding a `let_value` wrapper to pika in a future patch.

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -0,0 +1,208 @@
...
+#include <pika/async_mpi/mpi_polling.hpp>
+#include <pika/concepts/concepts.hpp>
+#include <pika/datastructures/variant.hpp>
+#include <pika/debugging/demangle_helper.hpp>
+#include <pika/debugging/print.hpp>
+#include <pika/execution/algorithms/detail/helpers.hpp>
+#include <pika/execution/algorithms/detail/partial_algorithm.hpp>
+#include <pika/execution/algorithms/drop_value.hpp>
+#include <pika/execution/algorithms/just.hpp>
+#include <pika/execution/algorithms/schedule_from.hpp>
```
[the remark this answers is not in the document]
biddisco: Note - `libs/pika/execution/include/pika/execution/algorithms/transfer.hpp` uses schedule_from, but does not include it. #705

libs/pika/async_mpi/src/mpi_polling.cpp
biddisco: grumble grumble. that's useful code that I will want when ranges are available.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
biddisco: removed for now

libs/pika/async_mpi/src/mpi_polling.cpp
biddisco: left over from experiments with throttling. Would like to leave it for now and remove it or use it properly when I re-enable throttling

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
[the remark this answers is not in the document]
msimberg: Thanks, good catch.

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -0,0 +1,268 @@
...
+            operation_state_type op_state;
+
+            template <typename Tuple>
+            struct value_types_helper
+            {
+                using type = pud::transform_t<Tuple, std::decay>;
+            };
+
+            // we don't bother storing the result, but the correct type is ...
+            using result_type = pika::detail::variant<std::tuple<MPI_Request>>;
```
msimberg, line 202: If we don't bother storing the result, why leave it here?
```suggestion
```

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -0,0 +1,263 @@
...
+                                // call set_value/set_error depending on mpi return status
+                                set_value_error_helper(
+                                    r.op_state.status, PIKA_MOVE(r.op_state.receiver));
+                                break;
+                            }
+                            default:
+                            {
+                                // I bet the "review" police don't like this exception
+                                throw mpi_exception(MPI_ERR_UNKNOWN, "eek!");
+                                break;
```
msimberg, line 186: :rotating_light:
Reaching this point is not an MPI problem, it's a logic error and we should terminate here.
```suggestion
                                PIKA_UNREACHABLE;
```
:rotating_light:

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -0,0 +1,263 @@
...
+        static constexpr bool sends_done = false;
+
+        // -----------------------------------------------------------------
+        // operation state for an internal receiver
+        template <typename Receiver>
+        struct operation_state
+        {
+            std::decay_t<Receiver> receiver;
+            int mode_flags_;
+            int status;
```
msimberg, line 85: Is `status` always initialized? Codacy doesn't seem to think so but it may be wrong.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -467,62 +627,175 @@ namespace pika::mpi::experimental {
...
+        detail::task_completion_flags_ = flags;
+
+        // if pool is now disabled, just exit
+        if ((flags & 1) == 0)
+            return false;
+
+        // Disable idle backoff on the MPI pool
+        using pika::threads::scheduler_mode;
+        auto smode = scheduler_mode::default_mode;
+        smode = scheduler_mode(smode & ~scheduler_mode::enable_idle_backoff);
```
msimberg, line 750: This should work (see https://github.com/pika-org/pika/blob/b38f3b6c8fc5ebece3fd730c3b806022645cd583/libs/pika/threading_base/include/pika/threading_base/scheduler_mode.hpp#L72-L76):
```suggestion
        smode = smode & ~scheduler_mode::enable_idle_backoff;
```

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -467,62 +627,175 @@ namespace pika::mpi::experimental {
...
+            if (detail::mpi_data_.size_ == 1)
+            {
+                flags &= ~1;
+            }
+        }
+        using namespace pika::debug::detail;
+        PIKA_DETAIL_DP(detail::mpi_debug<6>,
+            debug(str<>("completion mode"), bin<8>(flags), detail::mode_string(flags)));
+        // override the variable used to control completion mode and pool flags
+        setenv("PIKA_MPI_COMPLETION_MODE", std::to_string(flags).c_str(), true);
```
msimberg, line 739: Why are we modifying an environment variable here?

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg: Guard pages are now disabled by default so you can remove that option.

libs/pika/async_mpi/src/mpi_polling.cpp
msimberg: Just FYI, we'd have to wait for C++23 to get `zip`. We can of course consider using standalone ranges-v3 until then but we should see how many use cases we have for it before adding it as a dependency.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -467,62 +627,175 @@ namespace pika::mpi::experimental {
...
+        int is_initialized;
+        MPI_Initialized(&is_initialized);
+        if (is_initialized)
+        {
+            MPI_Comm_rank(MPI_COMM_WORLD, &detail::mpi_data_.rank_);
+            MPI_Comm_size(MPI_COMM_WORLD, &detail::mpi_data_.size_);
+        }
+        else
+        {
+            throw std::runtime_error("MPI must be initialized prior to pool creation");
```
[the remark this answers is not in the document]
biddisco: we need to know how many ranks there are. if only 1, then skip pool creation.

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -0,0 +1,263 @@
...
+            return operation_state<Receiver>(
+                PIKA_FORWARD(Receiver, receiver), PIKA_MOVE(s.sender), s.completion_mode_flags_);
+        }
+    };
+
+}    // namespace pika::mpi::experimental::detail
+
+namespace pika::mpi::experimental {
+
+    namespace exp = pika::execution::experimental;
```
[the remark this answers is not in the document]
biddisco: I have put the using clause inside an anonymous namespace to hide it from unintented include usage. Applied that to all places where I've added global namespace aliases

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
biddisco: no it isn't, but it is only used by one of the modes (suspend/resume), the others call set_error when necessary without needing to store it. (default initialization would not hurt, but ...)

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
biddisco: because it takes me ges to find the right type when I change anything and so having it there is a reminder to my future self when I decide to modify the result for some reason. I will remove it

libs/pika/async_mpi/src/mpi_polling.cpp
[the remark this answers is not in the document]
msimberg: Thanks, that makes sense. Could you update the exception?

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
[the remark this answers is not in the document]
msimberg: Did you forget to push this change? Also, doesn't that just affect visibility of namespace-level objects, not visibility of a `namespace a = b`?

libs/pika/async_mpi/src/mpi_polling.cpp
msimberg: This is still there, did you forget to push?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -13,43 +13,62 @@
...
 #include <utility>
 #include <vector>
 
 namespace pika::mpi::experimental {
 
+    template <typename E>
+    constexpr std::underlying_type_t<E> to_underlying(E e) noexcept
+    {
+        return static_cast<std::underlying_type_t<E>>(e);
+    }
```
msimberg, line 38: This should most probably not be in the public namespace, but under `detail`.

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -0,0 +1,266 @@
...
+                // otherwise return the request by passing it to set_value
+                template <typename... Ts,
+                    typename = std::enable_if_t<is_mpi_request_invocable_v<F, Ts...>>>
+                friend constexpr void
+                tag_invoke(ex::set_value_t, dispatch_mpi_receiver r, Ts&&... ts) noexcept
+                {
+                    pika::detail::try_catch_exception_ptr(
+                        [&]() mutable {
+                            using namespace pika::debug::detail;
+                            using ts_element_type = std::tuple<std::decay_t<Ts>...>;
```
msimberg, line 127: I think this is unused. Can it be removed?

libs/pika/threading_base/include/pika/threading_base/thread_data.hpp
msimberg: The `explicit` would still be good to have here (I don't care much about the parameter name). This could also be made immovable.

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -30,9 +33,15 @@
 
 namespace pika::mpi::experimental {
     namespace transform_mpi_detail {
+        // -----------------------------------------------------------------
+        // by convention the title is 7 chars (for alignment)
+        using print_on = pika::debug::detail::enable_print<false>;
+        static print_on mpi_tran("MPITRAN");
```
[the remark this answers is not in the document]
msimberg: Note, `constexpr` can stay. Now it's just `inline`.

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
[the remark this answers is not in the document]
biddisco:
```
namespace pika::mpi::experimental::detail {
    namespace {
    }
...
```
It only makes the `ex::` namespace visible when you are already including this file and in namespace `pika::mpi::experimental::detail` so not really leaking out.

msimberg (approved):
Approved modulo CI runs.

https://github.com/pika-org/pika/issues/612 still needs to be resolved outside of this PR.

msimberg:
bors try

msimberg:
bors merge

PR #601, opened 2023-02-17
Add formatting script that rejoins line-broken strings and then reapplies clang-format

biddisco:
Rejoin broken strings and then reapply clang-format to improve string layout

### DIFF
--- a/tools/string-join.awk
+++ b/tools/string-join.awk
@@ -1,8 +1,14 @@
-# if line ends with a " store line in buffer
-# if line starts with " and buffer is non empty, join to buffer and remove "[[:space:]]"
-# if neither start or end, print buffer if non empty, print line 
-# if we see "clang-format off" turn off processing
-# if we see "clang-forma on" turn back on processing
+# Copyright (c) 2019 The STE||AR-Group
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+# join strings across lines, not foolproof but handles 99% of c++ cases
+# can fail for empty strings like
+# std::string x = "stuff " + "    "
+#                 " more string carried over from above"
+#
 
 function clear_buffer() {
     if (buffer) print buffer
@@ -10,14 +16,9 @@ function clear_buffer() {
     in_string=0
 }
 
-# don't allow \n" to be replaced, to keep user enforced breaks
+# remove first instance of "   "
 function join_strings(buf) {
-    #  can't get negative lookbehind to work in awk, so instead
-    gsub("\\\\n\"", "REPLACE_HACK_24601", buf)
-    gsub("\\\\\"", "REPLACE_HACK_24602", buf)
-    gsub("\"[[:space:]]*\"", "", buf)
-    gsub("REPLACE_HACK_24601", "\\n\"", buf)
-    gsub("REPLACE_HACK_24602", "\\\"", buf)
+    sub("\"[[:space:]]*\"", "", buf)
     buffer=buf
 }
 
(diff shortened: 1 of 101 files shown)


tools/string-join.awk
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) 2019 The STE||AR-Group
```
msimberg, line 1:
```suggestion
# Copyright (c) 2023 ETH Zurich
```
Maybe also on the formatting script.

tools/string-join.awk
```diff
@@ -0,0 +1,81 @@
...
+    buffer=buf
+}
+
+BEGIN {
+    enabled=1
+    in_string=0 
+}
+
+# if we see "clang-format off" turn off processing
+# if we see "clang-forma on" turn back on processing
```
msimberg, line 31:
```suggestion
# if we see "clang-format on" turn back on processing
```

tools/string-join.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# iterate over cpp/etc files and join strings that are broken
+# then, rerun clang-format to rebreak the strings into good sizes
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+awkscript=$SCRIPT_DIR/string-join.awk
+tmpfile="$(mktemp /tmp/XXXXXXXXX.cpp)" || exit 1
+
+for file in $(git ls-files | grep -E "\.(cpp|hpp|cu)(\.in)?$"); do
+    awk -f $awkscript ${file} >${tmpfile}
```
msimberg, line 17:
```suggestion
    awk -f "$awkscript" "${file}" >"${tmpfile}"
```
and the same for the other codacy suggestions below (it's all about double-quoting variables).

msimberg (changes requested):
Thanks @biddisco. This is really nice.

Could you please update this to use the new formatting script: https://github.com/pika-org/pika-algorithms/blob/3e4ea7cd7f5eec8eb9c462beae0076c550aa0151/.circleci/config.yml#L104?

biddisco:
> Thanks @biddisco. This is really nice.
> 
> Could you please update this to use the new formatting script: https://github.com/pika-org/pika-algorithms/blob/3e4ea7cd7f5eec8eb9c462beae0076c550aa0151/.circleci/config.yml#L104?

I don't quite understand what you're asking - are you saying that the default clang-format check should also run the string rejoin script too? - or are you asking for a change to the too/script to use something else?

msimberg:
> I don't quite understand what you're asking - are you saying that the default clang-format check should also run the string rejoin script too? - or are you asking for a change to the tool/script to use something else?

The first. The only way we're going to _consistently_ remember to use the string-joining script is by enforcing it in CI. Otherwise it's only going to applied every now and then whenever someone remembers.

biddisco:
> > The first. The only way we're going to consistently remember to use the string-joining script is by enforcing it in CI. Otherwise it's only going to applied every now and then whenever someone remembers.
> 
> For what I understood the string joining was a one time thing. With the fact that we now allow the line length to be 100, the strings will most likely not be broken into multiple lines no? @biddisco correct me if I'm wrong

Yes, the string join could be run every time, but it is really only needed after a change in something like clang-format rules. It might smarten up some code that has strings when added but I don't think it needs to be run always.

I'd be fine with merging 'as is' (if you'e happy) and leaving it out of CI for the time being - there is one outstanding corner case with strings that could cause a fail (and then every commit would trigger it in CI) that needs to be fixed - then it could be added painlessly to CI.

msimberg:
Ok, thanks @aurianer and @biddisco for the comments. Let's leave it out for now then. I would prefer it to be in CI eventually, but if you're not completely happy with it yet let's leave that for another day.

msimberg (approved):

msimberg:
Merging manually as CI is not running at the moment.

PR #607, opened 2023-02-22
Clean up `hardware` module

msimberg:
Part of #16.

This moves the `timestamp` functionality to the `timing` module, and removes the unused `bit_manipulation.hpp` header. This leaves nothint in the `hardware` module, so it is removed.


msimberg:
bors merge

PR #606, opened 2023-02-22
Add support for NVHPC

msimberg:
This is still in need of some cleanup.

This adds support for using `nvc++` as the host and device compiler through `CMAKE_CXX_COMPILER`. This works the same way as `hipcc` used to work, i.e. `CMAKE_CUDA_COMPILER` is not set to `nvc++` (CMake does not yet support setting `nvc++` as a device compiler), and instead we change all `.cu` files to have `LANGUAGE` `CXX`, but with some CUDA-specific flags added. The device support in `nvc++` is experimental but seems to mostly be enough (though a bit buggy in some places) for our current test use cases.

This also introduces some changes to compile with `nvcc` 12, but this PR will not add CI for that. There are still some upstream fixes that need to go in before we can use that (e.g. into `fmt`).

I hope to have this in by 0.13.0 but no promises.

General reviews are welcome already, despite this being marked as a draft.

PR #616, opened 2023-03-06
Clean up macros and fix bug in `shared_priority_queue_scheduler`

biddisco:
A small fix to the scheduler where a wrong param is set, plus some debug cleanup that is needed for other cuda/mpi polling branches

### DIFF
--- a/libs/pika/schedulers/include/pika/schedulers/shared_priority_queue_scheduler.hpp
+++ b/libs/pika/schedulers/include/pika/schedulers/shared_priority_queue_scheduler.hpp
@@ -60,7 +60,6 @@ namespace pika {
     template <int Level>
     static debug::detail::print_threshold<Level, debug_level> spq_deb("SPQUEUE");
     static debug::detail::print_threshold<1, 1> spq_arr("SPQUEUE");
-#define DEBUG(printer, Expr) PIKA_DP_ONLY(printer, Expr)
 }    // namespace pika
 
 // ------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


biddisco:
bors try

libs/pika/schedulers/include/pika/schedulers/shared_priority_queue_scheduler.hpp
```diff
@@ -783,8 +782,7 @@ namespace pika::threads {
             execution::thread_schedule_hint schedulehint, bool allow_fallback,
             execution::thread_priority priority = execution::thread_priority::normal) override
         {
-            schedule_work(thrd, schedulehint, allow_fallback, false,
-                priority = execution::thread_priority::normal);
```
msimberg, line 787: No action required, just FYI: `nvc++` actually complained about this assignment, so this is also fixed on #606.

libs/pika/threading_base/src/print.cpp
```diff
@@ -11,6 +11,12 @@
 #include <cstdint>
 #include <thread>
 
+#if defined(__linux) || defined(linux) || defined(__linux__)
+# include <linux/unistd.h>
+# include <sys/mman.h>
+# define DEBUGGING_PRINT_LINUX
```
msimberg, line 17:
```suggestion
# define PIKA_DEBUGGING_PRINT_LINUX
```

msimberg:
bors merge

PR #622, opened 2023-03-10
Add GitHub merge queue branches to CI configurations

msimberg:
I'm testing if this could replace bors. I'm not removing the bors branches or app yet, and we can quickly turn this off in the settings if it doesn't work as well as bors, but in theory it works exactly the same (except for the a missing `bors try` equivalent).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,8 @@ branch_filter: &branch_filter
           # Bors branches
           - trying
           - staging
+          # GitHub merge queue branches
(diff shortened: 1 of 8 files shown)


msimberg:
I'm keeping the branches in the configs but turning this off at least for now... The missing bors try is a bit annoying. The other limitation is that this currently requires at least two runs of CI for every PR (unless the branch is on top of main): one green set of CI runs to even be able to add the PR to the merge queue, and then another one to actually merge in the merge queue.

PR #632, opened 2023-03-27
Clean up `include` module

msimberg:
Part of #16.

Remove includes that contain detail-only functionality from e.g. `pika/functional.hpp`. Replace includes of `pika/functional.hpp` with more specific includes when required for detail functionality.


msimberg:
bors try

PR #638, opened 2023-03-28
Fix option description for `PIKA_WITH_THREAD_DEBUG_INFO`

msimberg:
The option is not implicitly enabled in debug builds.


msimberg:
bors merge

PR #639, opened 2023-03-28
Update CI reference for future overhead test

### LINKED ISSUE #564: Investigate possible performance regression in the future overhead test

msimberg:
Hopefully fixes #564.

The reference we have currently is from September 2022. I went back to a release prior to the reference and I see a "performance regression" even there which tells us that there's been a regression on Piz Daint itself, so I'm simply updating the reference. On the positive side, squinting at a few of the last runs it may be that the variance is slightly lower now which is good, but I don't have any proof for this.

I've also added an option to `srun` to use the lowest supported CPU frequency to avoid frequency scaling and turbo boost altogether. This does mean the benchmark runs a lot slower, and may not be 100% realistic, but it does seem to be a bit more reliable. Let's try this out for a while and see how the reports work.


msimberg:
bors cancel

msimberg:
bors merge

msimberg:
bors try-

msimberg:
bors merge

PR #650, opened 2023-04-11
Use `find`'s `-depth` option when finding directories to delete in CI

msimberg:
This means using depth-first search, which avoids having `find` try to access directories that have just been deleted. This also removes the redirection of errors to `/dev/null` since that shouldn't be needed anymore.


msimberg:
bors merge

PR #660, opened 2023-04-24
Disable logging by default

msimberg:
Part 1 of N PRs to try to improve performance in DLA-Future. This is one of the changes in https://github.com/pika-org/pika/compare/main...disable-cruft-scheduler where we see up to 25% improvement in DLA-Future miniapps. This branch needs to be benchmarked to see if it makes a difference on its own.


msimberg:
This doesn't seem to have any meaningful impact (as one would hope) so I'm going to close it for now. Maybe when other parts have been optimized enough in the future this might have a measurable impact, but for now it doesn't hurt to leave it on.

PR #663, opened 2023-04-24
Disable guard pages by default

msimberg:
Fourth of N PRs to try to improve performance in DLA-Future. This is one of the changes in https://github.com/pika-org/pika/compare/main...disable-cruft-scheduler where we see up to 25% improvement in DLA-Future miniapps. This branch needs to be benchmarked to see if it makes a difference on its own.

https://github.com/pika-org/pika/tree/disable-guard-pages-default-compile-time needs to be benchmarked together with this branch to ensure that its enough to disable guard pages at runtime instead of compile time.


msimberg:
There seems to be no difference in the future overhead test from this change, but it has a big effect on DLA-Future, so we'll go ahead with this. Disabling them at runtime seems to be sufficient so we'll stick to that.

PR #667, opened 2023-04-26
Replace limiting executor with limiting scheduler

biddisco:
This is a more sender/receiver conforming version
of the task limiting mechanism.

A test is provided that spawns tasks and does not allow more than N to exist at a given time on the chosen scheduler. Multiple limiting schedulers may be used at the same time and they manage their own limiting independently.


libs/pika/executors/include/pika/executors/limiting_scheduler.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+    // -----------------------------------------------------------------
+    // by convention the title is 7 chars (for alignment)
+    // a debug level of N shows messages with level 1..N
+    using namespace pika::debug::detail;
+    template <int Level>
+    static print_threshold<Level, 0> lsc_debug("SCLIMIT");
+
+    template <typename OnScheduler>
+    struct limiting_scheduler
```
msimberg, line 41: Would this be ok as well:
```suggestion
    template <typename Scheduler>
    struct limiting_scheduler
```
?

libs/pika/executors/include/pika/executors/limiting_scheduler.hpp
```diff
@@ -0,0 +1,188 @@
...
+                sem->release();
+            };
+            internal_scheduler_.execute(PIKA_MOVE(semaphore_lambda), nullptr);
+        }
+
+        template <typename F>
+        friend void tag_invoke(execute_t, limiting_scheduler const& sched, F&& f)
+        {
+            pika::execution::experimental::execute(sched, PIKA_FORWARD(F, f));
+        }
```
msimberg, line 87: The `execute`s are backwards. The member function should use the generic `execution::execute` and the `tag_invoke` overload can use the member `execute`.

_However_, I wouldn't bother with the `execute`s at all. `connect` and `start` are the basis operations and you get `execute` for free from those. `execute` is customized on `thread_pool_scheduler` only to avoid unnecessary instantiations and indirection.

```suggestion
        template <typename F>
        void execute(F&& f) const
        {
            using namespace pika::debug::detail;
            PIKA_DETAIL_DP(lsc_debug<5>, debug(str<>("acquire")));
            semaphore_->acquire();
            auto semaphore_lambda = [f = std::forward<F>(f), sem = this->semaphore_]() mutable {
                PIKA_INVOKE(f, );
                PIKA_DETAIL_DP(lsc_debug<5>, debug(str<>("release")));
                sem->release();
            };
            pika::execution::experimental::execute(internal_scheduler_, PIKA_MOVE(semaphore_lambda));
        }

        template <typename F>
        friend void tag_invoke(execute_t, limiting_scheduler const& sched, F&& f)
        {
            execute(sched, PIKA_FORWARD(F, f));
        }
```

msimberg (changes requested):
Looks generally ok.

There are a couple of big gaps that I don't know how or if we can patch up. As it is right now this scheduler will disable all customizations that may be there for the internal scheduler. I.e. something like
```
schedule(limiting_scheduler(13, thread_pool_scheduler())) |
  then(f) |
  bulk(n, g);
```
will not be using the customization of of `bulk` for `thread_pool_scheduler`. We could forward customizations, but the problem is then that from that point onwards the completion scheduler will be the internal scheduler instead. Also: should the bulk work be limited counting each task as well, or does the whole bulk region just count as one big task? Probably the latter.

Another issue that might come up is something like the following:
```
schedule(limiting_scheduler(...)) |
    then(f) |
    transfer(inline_scheduler) |jjj
    then(g);
```
In this case `g` will also be "limited" because the counting semaphore is released only after returning from the whole (inline) chain of calls. `inline_scheduler` need not be exactly that, `cuda_scheduler` also fits the description.

All of this makes me think that something `async_scope`- or `on`-like is the better generic way forward. That doesn't mean this can't be useful in its current form though.

libs/pika/executors/include/pika/executors/limiting_scheduler.hpp
biddisco: I can't rename the OnScheduler to Scheduler, cos, that name is used for the Op state and internal sender template params. But I'll call it WrappedScheduler for the time being.

libs/pika/executors/include/pika/executors/limiting_scheduler.hpp
biddisco: ok. I'll switch the executes for now and then look at using connect/start for the next version.

biddisco:
Thanks for this review. I was slightly surprised it worked at all as it was mostly cut'n'pasted from elsewhere.

 ```
In this case g will also be "limited" because the counting semaphore is released only after returning 
from the whole (inline) chain of calls. inline_scheduler need not be exactly that, 
cuda_scheduler also fits the description.
```
yes. This is a problem  - once you've out a limiting scheduler in your pipeline - you're basically saying - throttle all of this. It's hard to imagine how you could limit the number of (say) mpi messages to N, but allow more than N continuations to run - but in general yes - we want the `release` of the semaphore to happen when the `function` is done and not when the whole chain is done. To break the invocations we'd need an explicit transfer off the limiting scheduler, or something alone these lines as you've written
```
All of this makes me think that something async_scope- or on-like is the better generic way forward. 
That doesn't mean this can't be useful in its current form though.
```

I am considering this only as a better version of my throttling so that I can fix the mpi ring test to throttle properly and remove some of the the code from the mpi PR and use this instead, but I agree that a better version should be made for V2 and this is only a provisional implementation.
```
There are a couple of big gaps that I don't know how or if we can patch up. 
As it is right now this scheduler will disable all customizations that may be there for the internal scheduler. 
```
^ this too.

**Review by msimberg — APPROVED** [review]
This doesn't have any effect of what we're doing right now in DLA-Future so let's go ahead with the current version. Thanks!

**Comment by msimberg** [comment]
bors try

File: libs/pika/executors/include/pika/executors/limiting_scheduler.hpp [hunk 3]
```diff
@@ -0,0 +1,190 @@
...
+                sem->release();
+            };
+            pika::execution::experimental::execute(
+                internal_scheduler_, PIKA_MOVE(semaphore_lambda));
+        }
+
+        template <typename F>
+        friend void tag_invoke(execute_t, limiting_scheduler const& sched, F&& f)
+        {
+            execute(sched, PIKA_FORWARD(F, f));
```
msimberg, line 87: Oh right, this needs to be
```suggestion
            sched.execute(PIKA_FORWARD(F, f));
```
Sorry for the previous suggestion...

msimberg:
bors try

libs/pika/executors/tests/unit/limiting_scheduler.cpp
```diff
@@ -0,0 +1,259 @@
...
+        pika::this_thread::yield();
+        work_fn();
+    }
+
+    // task is completing, decrement active task count
+    --active_count;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+void test_limit_inline()
```
msimberg, line 92: Could you not template this on the scheduler? Or are there important differences in the two tests that you can't combine into one function?

libs/pika/executors/include/pika/executors/inline_scheduler.hpp
```diff
@@ -0,0 +1,38 @@
...
+        {
+            PIKA_INVOKE(f, );
+        }
+
+        template <typename F>
+        friend void tag_invoke(execute_t, inline_scheduler const& sched, F&& f)
+        {
+            sched.execute(PIKA_FORWARD(F, f));
+        }
+    };
```
msimberg, line 35: Since this is not a scheduler (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2300r6.html#spec-execution.schedulers) I'd prefer it if we only pulled in the stdexec one conditionally (you wouldn't need the header at all and would use it from `exec/inline_scheduler.hpp` instead; the include of the `stdexec_forward.hpp` isn't doing what you you wish for it to do either since it only pulls in `namespace stdexec`, not `namespace exec`). Would that be sufficient for your testing? You couldn't use it with DLA-Future (yet) but maybe it's still enough for now?

msimberg (changes requested):

libs/pika/executors/tests/unit/limiting_scheduler.cpp
biddisco: Only real difference is that the inline one tests that the tasks have the same ID, but the other one has different IDs (TEST_NEQ/TEST_EQ)

msimberg:
What's the plan for this PR?

PR #670, opened 2023-05-03
Always steal tasks if scheduler mode allows for it

msimberg:
Below is an example of the performance bump seen in DLA-Future. Most algorithms don't have such a clear performance improvement, but this change also doesn't hurt the other algorithms. The dark blue line corresponds to the changes in this PR. The other lines contain other scheduler-related changes that don't seem to make a difference in DLA-Future. One of the other lines is pika main.
bt_red2band_strong_ppn_10240_10240_256_128

PR #672, opened 2023-05-04
Unify `pika::spinlock` and `pika::concurrency::detail::spinlock` into one implementation

### LINKED ISSUE #517: Attempt to use non-yielding spinlocks in short critical sections
Or, even better, replace locks with lock-free implementations. This might be hard. However, many locks protect protect sections that are almost always only written by one writer and read by one reader, and the sections are very short. In these cases it's _probably_ not worth it to yield the user-level thread in the unfortunate cases that a lock can't be taken quickly enough.

msimberg:
Fixes #517, i.e. uses a non-yielding everywhere where a spinlock was used previously. The implementations were almost identical, but used a slightly different yielding strategy. The yielding one (`pika::spinlock`) used `yield_while` allowing yielding of the user-level thread. The non-yielding one (`pika::concurrency::detail::spinlock`) was sleeping the OS-thread after one iteration. This now uses the `pika::spinlock` strategy everywhere, except that yielding is disallowed when spinning. The name `pika::spinlock` is removed and only the `detail` one remains.

Note that we used to have _three_ spinlock implementations. Now we still have two. The remaining one is a very basic one with near zero dependencies (no lock registration, no ITT support) that is still used in things like the configuration maps and resource partitioner.

This doesn't change performance in either direction in DLA-Future, but it's a prerequisite for having a semi-blocking barrier (https://github.com/eth-cscs/DLA-Future/issues/833).

We will need to make corresponding changes in pika-algorithms since it uses `pika::spinlock` in a few places.


msimberg:
I think the performance regression actually comes from https://github.com/pika-org/pika/pull/670, not this PR. I'm going to revert that one instead for the moment.

msimberg:
I rebased this on main after #670 was reverted and the performance test is back to normal. I would go ahead with this.

msimberg:
bors try

msimberg:
Rebased after #679 was merged.

bors try

msimberg:
bors merge

PR #674, opened 2023-05-05
Revert "Always steal tasks if scheduler mode allows for it" #670

msimberg:
Reverts #670. It seems it introduced a performance regression, at least in the future overhead benchmark. It did show a small performance improvement in DLA-Future, so I'm guessing the effect of the change is very much workload dependent. We might instead have to consider tweaking the threshold when we enable stealing.


msimberg:
bors merge

msimberg:
bors merge

PR #677, opened 2023-05-08
Eagerly reset shared state in `async_rw_mutex`

msimberg:
Eagerly reset the shared state held by the operation state to allow continuations to run as early as possible. This fixes a deadlock that occurred in the added test case.


msimberg:
I've split out the spinlock change into a separate PR: #679.

msimberg:
bors try

msimberg:
This does actually need #678 to go in first because of some changes in stdexec. The `async_rw_mutex` test passes now with nvc++ though.

msimberg:
I've removed the stdexec changes from this PR and simply disabled the new test for now with nvhpc+stdexec. It needs to be reenabled again in #678.

bors merge

PR #691, opened 2023-05-22
Allow compilation with fmt 10 with CUDA/HIP

### LINKED ISSUE #686: Fix compilation with fmt 10
It seems like compiling pika with CUDA/HIP enabled now breaks with fmt 10 (e.g. https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4700071344751697/7514005670787789/-/jobs/4274622344#L6). This seems to be coming from fmt not wanting to format `cudaError_t`/`hipError_t` without a formatter. We should investigate why this changed and then fix it in pika (cast to `int` since that's all the error codes are?).

msimberg:
Fixes compilation by explicitly casting the enums to `int`s for printing, since they are no longer implicitly converted for printing by fmt 10.

Also adds fmt 10 to the CUDA 12 CI configuration.

Fixes #686.

### DIFF
--- a/.jenkins/cscs/env-gcc-cuda-12.sh
+++ b/.jenkins/cscs/env-gcc-cuda-12.sh
@@ -9,10 +9,11 @@ gcc_version="12.1.0"
 boost_version="1.82.0"
 hwloc_version="2.9.1"
 cuda_version="12.0.0"
+fmt_version="10.0.0"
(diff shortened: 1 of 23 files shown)


msimberg:
bors try

msimberg:
bors try

msimberg:
bors merge

PR #702, opened 2023-06-20
Take values in `any_receiver` `set_value` by universal reference

aurianer:
Fix a bug triggered in DLAF or with this minimal test in pika:
```
                                                                                                    
class SweepWorker {                                                                                   
public:                                                                                               
  SweepWorker() {}                                                                                    
  SweepWorker(SweepWorker&&) {};                                                             
};                                                                                                    
                                                                                                      
int main()                                                                                            
{                                                                                                     
    ex::unique_any_sender<SweepWorker> send_ = ex::just(std::move(SweepWorker()));                  
    return 0;                                                                                       
} 
```
Adding noexcept to the copy constructor in DLAF or having this fix in solves the problem


msimberg (changes requested):
Thanks!

Could please add the test from the description (i.e. a test with a throwing move constructor) and a test of wrapping a sender which sends `T&` or `T const&` (you can use `split` or one of the test helper senders) into a `(unique_)any_sender<T>` (no ref). I _think_ that case may be broken by this change but I'm not completely sure (can't pass a `T&` to `T&&`).

There's a third case: wrapping a sender sending `T&` or `T const&` in a `(unique_)any_sender<T&>`. You don't need to consider this case in this PR because we pretended not to support that case so far.

libs/pika/execution_base/include/pika/execution_base/any_sender.hpp
```diff
@@ -447,7 +447,7 @@ namespace pika::execution::experimental::detail {
             new (p) any_receiver_impl(PIKA_MOVE(receiver));
         }
 
-        void set_value(Ts... ts) && override
+        void set_value(Ts&&... ts) && override
         {
             pika::execution::experimental::set_value(PIKA_MOVE(receiver), PIKA_MOVE(ts)...);
```
msimberg, line 452: I think this should be:
```suggestion
            pika::execution::experimental::set_value(PIKA_MOVE(receiver), PIKA_FORWARD(Ts, ts)...);
```
_if_ we support `any_sender` with reference types. However, for now, we could instead simply static assert that it's only used with non-reference types since I think we don't have an immediate use case for storing reference types. It may be needed in the future though so you could open an issue for that.

msimberg (commented):
What's the status of this PR?

libs/pika/execution_base/include/pika/execution_base/any_sender.hpp
```diff
@@ -504,15 +504,17 @@ namespace pika::execution::experimental::detail {
...
+            pika::execution::experimental::set_value_t, any_receiver&& r, Ts_&&... ts) noexcept
+            -> decltype(PIKA_MOVE(PIKA_MOVE(r.storage.get())).set_value(PIKA_FORWARD(Ts_, ts)...))
         {
             // We first move the storage to a temporary variable so that
             // this any_receiver is empty after this set_value. Doing
             // PIKA_MOVE(storage.get()).set_value(...) would leave us with a
             // non-empty any_receiver holding a moved-from receiver.
             auto moved_storage = PIKA_MOVE(r.storage);
-            PIKA_MOVE(moved_storage.get()).set_value(PIKA_MOVE(ts)...);
+            PIKA_MOVE(moved_storage.get()).set_value(PIKA_FORWARD(Ts_, ts)...);
```
msimberg, line 517: Now that we've moved the possibility of constructors throwing into this function we should add a try-catch block around this, calling `set_error` in the catch.

msimberg (changes requested):
Thanks, the functionality looks good now. It needs few small tweaks and tests.

libs/pika/execution_base/tests/unit/any_sender.cpp
```diff
@@ -699,6 +699,63 @@ void test_when_all()
...
+    tag_invoke(pika::execution::experimental::connect_t, const_reference_sender&& s, R&& r)
+    {
+        return operation_state<R>{std::move(s.x), std::forward<R>(r)};
+    }
+};
+
+void test_reference_and_const_reference()
+{
+    int x = 42;
+    [[maybe_unused]] ex::unique_any_sender<int> send_const_ref = const_reference_sender<int>{x};
```
msimberg, line 755: Don't forget to make the tests cover both `unique_any_sender` and `any_sender`.

libs/pika/execution_base/tests/unit/any_sender.cpp
```diff
@@ -11,8 +11,11 @@
 #include <pika/modules/errors.hpp>
 #include <pika/testing.hpp>
 
+#include "algorithm_test_utils.hpp"
```
msimberg, line 14: Could you please move it to a subdirectory?
```suggestion
#include <pika/execution_base/tests/algorithm_test_utils.hpp>
```
Don't forget to update the tests in `execution` to use this header as well.

libs/pika/execution_base/tests/unit/any_sender.cpp
```diff
@@ -696,6 +626,59 @@ void test_when_all()
...
+    throwing_constructor(const throwing_constructor&) = default;
+};
+
+void test_throwing_constructor()
+{
+    {
+        // With unique_any_sender
+        throwing = false;
+        std::atomic<bool> set_error_called{false};
+        ex::unique_any_sender<throwing_constructor> send_unique_any_ =
```
msimberg, line 649: Could you please update the variable names here (and further down) to be consistent with the other variables in this file? Especially no trailing `_` please.

libs/pika/execution_base/tests/include/pika/execution_base/tests/algorithm_test_utils.hpp
```diff
@@ -372,6 +372,13 @@ struct const_reference_sender
     {
         return operation_state<R>{std::move(s.x), std::forward<R>(r)};
     }
+
+    template <typename R>
+    friend auto
+    tag_invoke(pika::execution::experimental::connect_t, const_reference_sender const& s, R&& r)
+    {
+        return operation_state<R>{std::move(s.x), std::forward<R>(r)};
```
msimberg, line 380:
```suggestion
        return operation_state<R>{s.x, std::forward<R>(r)};
```

msimberg (commented):
bors try

libs/pika/execution_base/tests/unit/any_sender.cpp
```diff
@@ -696,6 +626,93 @@ void test_when_all()
...
+        ex::start(os);
+        PIKA_TEST(set_error_called);
+    }
+
+    {
+        // Throwing move with any_sender
+        throwing = false;
+        std::atomic<bool> set_error_called{false};
+        ex::any_sender<throwing_constructor> as1 = ex::just(std::move(throwing_constructor()));
+        auto r = error_callback_receiver<decltype(check_exception_ptr)>{
```
msimberg, line 669: Valid warnings triggered by clang:
```suggestion
            ex::just(throwing_constructor());
        auto r = error_callback_receiver<decltype(check_exception_ptr)>{
            check_exception_ptr, set_error_called};
        auto os = ex::connect(std::move(as1), std::move(r));
        throwing = true;
        ex::start(os);
        PIKA_TEST(set_error_called);
    }

    {
        // Throwing move with any_sender
        throwing = false;
        std::atomic<bool> set_error_called{false};
        ex::any_sender<throwing_constructor> as1 = ex::just(throwing_constructor());
        auto r = error_callback_receiver<decltype(check_exception_ptr)>{
```

msimberg (approved):
Thanks @aurianer, let's get this merged!

msimberg:
bors merge

msimberg:
bors merge

Issue #703, opened 2023-06-21
Remove `int` return value from certain runtime functions


msimberg:
Some runtime control functions such as `pika::suspend`, `resume`, `wait` return an `int` but the int contains no useful information (the functions will either throw or set `ec` on failures). I think we can make them return `void`. Functions like `pika::init` and `pika::stop` still need to return `int` (the return value from `pika_main`).

PR #704, opened 2023-06-21
Add `pika::wait`

### LINKED ISSUE #696: Add `pika::wait` function to wait for runtime to be idle
Essentially:
```
namespace pika {
    void wait() {
        get_thread_manager().wait();
    }
}
```
Note, this is not `runtime::wait()` which is internal and waits for `finalize` to be called. `pika::wait` should be used to replace calls to `get_thread_manager().wait()` in DLA-Future.

msimberg:
Fixes #696.

Calls `thread_manager::wait`. `pika::wait` can also be called on a pika thread, in which case it'll ignore the calling thread and wait for all other work to finished. The behaviour is undefined/will likely deadlock if `pika::wait` is called concurrently from multiple pika threads.

Note, I made `wait` return an `int` to be consistent with the other related functions, but I opened #703 to possibly change it to `void`.


msimberg:
bors try

msimberg:
bors try

msimberg:
This seems to have exposed some problems in `thread_manager::wait`. I will try to get back to this in a few weeks.

msimberg:
This is on hold because I had forgotten about the probabilistic nature of checking if the runtime is idle. We don't have a global indicator to check that all of the thread pools and MPI/CUDA polling is idle, which means that it's inherently racy to individually check those for idleness and infer idleness of the runtime from those.

We may need to try to introduce a global counter but we have to be careful not to kill performance with a global atomic counter.

msimberg:
I've removed the test from this PR and left only the `pika::wait` wrapper for `thread_manager::wait` so that we can start replacing uses of `thread_manager::wait` in DLA-Future with a public API function. The test is in #704 but still needs a deterministic check for the runtime being empty.

msimberg:
Merging despite failures, which seem to mostly be timeouts on header tests.

PR #707, opened 2023-06-27
Kill jobs on ault with djenkssl user

aurianer:
Some jobs are pending with reason "launch failed requeue held", this is an attempt to kill those.


msimberg:
This PR is not meant to be merged, correct? Is it still needed?

PR #709, opened 2023-07-11
Add test for `pika::wait`

msimberg:
`pika::wait` will currently probabilistically wait for the runtime to be empty. This adds a test which should never fail. This is a draft until we implement a way to wait for the runtime to be empty deterministically. See https://github.com/pika-org/pika/pull/704#issuecomment-1602212921. The test from #704 was extracted to a separate PR so that we can merge `pika::wait` on its own.


msimberg:
bors try

msimberg:
Not ready for merging yet because it needs further testing in DLA-Future and an update to the performance test reference. However, I'd appreciate an early look from @aurianer and @biddisco.

The global counter obviously has a performance impact especially with many tasks (like in the future overhead benchmark), but preliminary results seem to show now measurable performance difference in DLA-Future and in that case I'd rather take the correctness of this method over a hypothetical performance improvement of the current method on main. We can always try to optimize this later to find a method that's both accurate (i.e. `wait` never returns before the runtime is empty) and performant (i.e. that doesn't require a single shared atomic).

msimberg:
I'm going to go ahead with this. The performance test reference showed a small regression but DLA-Future shows the same or slightly better performance.

Please keep an eye out for weird hangs etc. during shutdown!

PR #712, opened 2023-07-11
Attach debugger after printing failure information

msimberg:
When an assertion is handled we optionally allow attaching a debugger before terminating. Currently on `main` the debugger can be attached before printing the error message. This PR moves `may_attach_debugger` to after the error message has been printed so that one can see the reason for failure before deciding to attach a debugger.


msimberg:
bors try

msimberg:
bors try

PR #713, opened 2023-07-11
Reformat short blocks etc. to a single line

msimberg:
This is a proposal to change the `.clang-format` configuration to allow reformatting short blocks, functions, lambdas etc. to a single line. I much prefer this style because it's often a lot more compact and readable. However, I want to see what you think. Not urgent, not a must. Let me know what you think.


msimberg:
> I'm mostly fine with the changes though I think it reduces the readability of the if statements, but it's not a super strong opinion so I'm flexible

Yep, good point! The initial config was the most extreme `AllowShort*OnASingleLine` configuration. For comparison, here's the result with `AllowShortIfStatementsOnASingleLine: WithoutElse`, which means that only lonely `if (x) y;` without `else` are allowed to be reformatted to a single line. I'm not stuck to this either and would happily use `AllowShortIfStatementsOnASingleLine: Never` as well.

Just to be clear, the case where I find these settings most useful is short functions and lambdas. Other cases are less useful.

biddisco (approved):
I'm ok with these changes

msimberg:
I'll update this and get it ready for merging after the 0.17.0 release.

msimberg:
bors try

msimberg:
bors try

msimberg:
Test failures are unrelated to the changes here.

msimberg:
I'm going to merge this manually now. Timeouts are filesystem related again.

PR #716, opened 2023-07-17
Fix freeing stack when guard pages are disabled

### LINKED ISSUE #715: Starting and stopping the runtime repeatedly seems to leak resources/memory on some systems
On some systems (but not all) a test program like
```
#include <pika/init.hpp>

int main() {
    while (true) {
        pika::start(nullptr, 0, nullptr);
        pika::finalize();
        pika::stop();
    }
}
```
will eventually run out of memory (typically failing while trying to allocate thread stacks). Changing the `pika.thread_queue.init_threads_count` configuration option (easily changed with the environment variable `PIKA_THREAD_QUEUE_INIT_THREADS_COUNT` environment variable) changes the behaviour. The option controls how many thread stacks are pre-initialized at startup. With a high value the failure happens quickly and with a low value the failure is significantly delayed. This would indicate a leak of thread stacks. However, from the looks of it stacks are correctly released at shutdown. Additionally, on some systems memory does not seem to leak. This could indicate a difference in behaviour of `mmap`/`munmap`. When `PIKA_WITH_THREAD_STACK_MMAP=OFF` there seem to be no leaks supporting the `mmap`/`munmap` theory. What exactly could be causing this is not clear.

msimberg:
When guard pages are disabled, we were still allocating an extra page as if we were using guard pages but then deallocating as if we're not using guard pages. This lead to memory leaks.

Fixes #715.

### DIFF
--- a/libs/pika/coroutines/include/pika/coroutines/detail/posix_utility.hpp
+++ b/libs/pika/coroutines/include/pika/coroutines/detail/posix_utility.hpp
@@ -48,7 +48,9 @@
 # include <cerrno>
 # include <cstddef>
 # include <cstdlib>
+# include <cstring>
 # include <stdexcept>
+# include <string>
 
 # if defined(_POSIX_MAPPED_FILES) && _POSIX_MAPPED_FILES > 0
 #  include <errno.h>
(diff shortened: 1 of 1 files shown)


msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

RMeli (approved):
https://github.com/eth-cscs/DLA-Future/pull/886#issuecomment-1639855081

msimberg:
bors merge

Issue #49, opened 2022-01-21
Add documentation for pika


aurianer:
We could initially refer to hpx's documentation and provide only the things that differs from it. But it might diverge too much from it in the future to keep it like this.

msimberg:
To try to have something actionable here, I think we would need the following in order of importance:

1. list of public API headers
2. examples of typical use cases
3. list of public API functions/classes/variables (doxygen, and manually curated?)
4. examples of using specific APIs
5. detailed documentation per function/class/variable

msimberg:
I gave hdoc (https://hdoc.io/) a try as an alternative to doxygen and the result is here: https://docs.hdoc.io/msimberg/pika/index.html.
- I like the search
- It would need a way to exclude `detail` namespaces
- Some of the docstrings don't render correctly
In principle if the last two points were fixable I'd consider this but as far as I understand there's at least no way to fix the middle point at the moment (the last may be fixable).

PR #723, opened 2023-07-28
Add documentation with deployment to GitHub pages

msimberg:
This adds v0.0.1-gamma (i.e. the most very basic) of a documentation page using sphinx.

It more or less contains the contents of the current `README.rst` with a little added section about pika's relation to stdexec/std::execution.

This is deployed using a couple of GitHub actions that use deployments (https://github.com/pika-org/pika/deployments) which as far as I can tell keeps it away from the git history and environments (https://github.com/pika-org/pika/settings/environments) for branch protections.

In this PR I will still:
- [x] remove duplicated content from `README.rst`
- [x] link to documentation in `README.rst`
- [x] Add a basic API page which lists the public headers that are currently in `README.rst`
- [x] Re-enable branch filters so that the the documentation is only deployed from `main` (but still built on PRs)
- [x] Fix any suggestions you have for the content that is included now

I will not:
- Add an API reference outside of the public headers
- Add examples
- Add other technical content (like explanations of how the CUDA/HIP/MPI integrations work, or how the thread pool works)

~The logo... leaves room for improvement. We may be better off without anything for now.~ The "logo" is just a home icon to go back to the index page.

This branch deploys already now to https://pika-org.github.io/pika/ so you can have a look at the result.

@aurianer and @biddisco I'd appreciate any and all feedback you have on the current content in this PR. Please comment on #49 for any _additional_ content you'd like to see, especially the priority of the type of content you'd like to see, so that we can try to work through the list in order of importance.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,68 @@
+# Copyright (c) 2023 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+name: Documentation
+
+on:
+  merge_group:
+  pull_request:
+    branches:
+      # Development and release branches
+      - main
+      - release**
+  push:
+    branches:
+      # Development and release branches
+      - main
+      - release**
+      # Bors branches
+      - trying
+      - staging
+
+jobs:
+  build:
+    name: github/documentation/build
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install sphinx and a theme
+      run: |
+          sudo apt update
+          sudo apt install \
+              --allow-downgrades \
+              --allow-remove-essential \
+              --allow-change-held-packages \
+              python3 python3-pip
+          pip install sphinx sphinx-material
+    - name: Build documentation
+      run: sphinx-build -W -n -a -b html docs build/docs
+    - name: Upload HTML output as artifact
+      uses: actions/upload-pages-artifact@v2
+      with:
+        path: build/docs
+
+  deploy:
+    needs: build
+    name: github/documentation/deploy
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main' && github.repository == 'pika-org/pika'
+
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    concurrency:
+      group: "pages"
+      cancel-in-progress: false
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    steps:
+    - name: Deploy documentation to GitHub Pages
+      uses: actions/deploy-pages@v2
+      id: deployment
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -62,7 +62,7 @@ pika optionally depends on:
 * `MPI <https://www.mpi-forum.org/>`_. MPI support can be enabled with ``PIKA_WITH_MPI=ON``.
 * `Boost.Context <https://boost.org>`_ on macOS or exotic platforms which are not supported by the
   default user-level thread implementations in pika. This can be enabled with
-  ``PIKA_WITH_GENERIC_CONTEXT_COROUTINES=ON`` (TODO rename option...).
+  ``PIKA_WITH_BOOST_CONTEXT=ON``.
 * `stdexec <https://github.com/NVIDIA/stdexec>`_. stdexec support can be enabled with
   ``PIKA_WITH_STDEXEC=ON`` (currently tested with commit `7a47a4aa411c1ca9adfcb152c28cc3dd7b156b4d
   <https://github.com/NVIDIA/stdexec/commit/7a47a4aa411c1ca9adfcb152c28cc3dd7b156b4d>`_).  The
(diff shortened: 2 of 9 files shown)


docs/usage.rst
```diff
@@ -0,0 +1,134 @@
...
+* `CUDA <https://docs.nvidia.com/cuda/>`_ 11.0 or greater. CUDA support can be enabled with
+  ``PIKA_WITH_CUDA=ON``. pika can also be built with nvc++ from the `NVIDIA HPC SDK
+  <https://developer.nvidia.com/hpc-sdk>`_. In the latter case, set ``CMAKE_CXX_COMPILER`` to
+  ``nvc++``.
+* `HIP <https://rocmdocs.amd.com/en/latest/index.html>`_ 5.2.0 or greater. HIP support can be
+  enabled with ``PIKA_WITH_HIP=ON``.
+* `MPI <https://www.mpi-forum.org/>`_. MPI support can be enabled with ``PIKA_WITH_MPI=ON``.
+* `Boost.Context <https://boost.org>`_ on macOS or exotic platforms which are not supported by the
+  default user-level thread implementations in pika. This can be enabled with
+  ``PIKA_WITH_GENERIC_CONTEXT_COROUTINES=ON`` (TODO rename option...).
```
RMeli, line 65: Will this be renamed before or after the documentation is released?

docs/usage.rst
```diff
@@ -0,0 +1,134 @@
...
+- The pika implementation uses ``value_types``, ``error_types``, and ``sends_done`` instead of
+  ``completion_signatures`` in sender types, as in the `first 3 revisions of P2300
+  <https://wg21.link/p2300r3>`_.
+
+pika has an experimental CMake option ``PIKA_WITH_STDEXEC`` which can be enabled to use stdexec for
+the P2300 facilities. pika brings the ``stdexec`` namespace into ``pika::execution::experimental``,
+but provides otherwise no guarantees of interchangeable functionality. pika only implements a subset
+of the proposed sender algorithms which is why we recommend that you enable ``PIKA_WITH_STDEXEC``
+whenever possible. We plan to deprecate and remove the P2300 implementation in pika in favour of
+stdexec and/or standard library implementations.
```
RMeli, line 134: Is there any rough/indicative timeline users can expect?

docs/usage.rst
```diff
@@ -0,0 +1,134 @@
...
+---------------------------
+
+Once installed, pika can be used in a CMake project through the ``pika::pika`` target. Other ways of
+depending on pika are likely to work but are not officially supported.
+
+.. code-block:: cmake
+
+   find_package(pika REQUIRED)
+   add_executable(app main.cpp)
+   target_link_libraries(app PRIVATE pika::pika)
```
RMeli, line 89:
```suggestion
Once installed, pika can be used in a CMake project through the ``pika::pika`` target:

.. code-block:: cmake

   find_package(pika REQUIRED)
   add_executable(app main.cpp)
   target_link_libraries(app PRIVATE pika::pika)
   
Other ways of depending on pika are likely to work but are not officially supported.
```

Move the "Other ways of depending on pika [...]" sentence after the code block?

docs/usage.rst
msimberg: Unsure. We have to decide on a name first... I'll probably disentangle the renaming from this PR and do the renaming afterwards. I just wanted a reminder for myself to rename it sooner rather than later.

I'm currently thinking simply `PIKA_WITH_BOOST_CONTEXT` is the most descriptive name for it. Thoughts? Other suggestions?

docs/usage.rst
msimberg: That's a good question. If there were one, would you want it written here?

I think we're realistically looking at 6-12 months if we get stdexec working cleanly in DLA-Future with all configurations, but not all configurations even compile yet so I'm wary to give an estimate here.

docs/usage.rst
msimberg: Yep, good idea!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,66 @@
...
+      cancel-in-progress: false
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    steps:
+    - name: Deploy documentation to GitHub Pages
+      uses: actions/deploy-pages@v2
+      id: deployment
+      # TODO: Only deploy from main when ready
```
msimberg, line 65: To do before merging.

msimberg:
This is "ready" meaning I will not add further content in this PR anymore. Suggestions on layout, structure, etc. are of course very much appreciated. I still need to update the workflow file to not deploy on PRs before merging, but I'm doing that as the last thing once reviews are done.

docs/usage.rst
RMeli: My question was mostly out of personal curiosity. I think a `.. note::` with a very rough timeline might be nice, but if it's too early to tell I'd leave it out then.

docs/usage.rst
```diff
@@ -0,0 +1,146 @@
...
+  run with ``ctest --output-on-failure``.
+- ``PIKA_WITH_EXAMPLES``: Enable examples. Binaries will be placed under ``bin`` in the build
+  directory.
+
+Testing
+-------
+
+Tests and examples are disabled by default and can be enabled with ``PIKA_WITH_TESTS``,
+``PIKA_WITH_TESTS_*``, and ``PIKA_WITH_EXAMPLES``. Note that reconfiguring pika with
+``PIKA_WITH_TESTS=ON`` after the option has been off the individual subcategories of
```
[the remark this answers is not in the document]
msimberg: Yeah, good catch. This was fixed with #356. I'll remove this.

docs/usage.rst
```diff
@@ -0,0 +1,146 @@
...
+Additionally, pika depends on:
+
+- `header-only Boost <https://boost.org>`_ 1.71.0 or greater
+- `hwloc <https://www-lb.open-mpi.org/projects/hwloc/>`_ 1.11.5 or greater
+- `fmt <https://fmt.dev/latest/index.html>`_ 9.0.0 or greater
+
+pika optionally depends on:
+
+* `gperftools/tcmalloc <https://github.com/gperftools/gperftools>`_, `jemalloc
+  <http://jemalloc.net/>`_, or `mimalloc <https://github.com/microsoft/mimalloc>`_. It is *highly*
```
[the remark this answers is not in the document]
msimberg: Hmm, good point. In fact, I think we might as well make it the default on all systems. One can always change it back to `system` if needed. `mimalloc` should be supported on windows and BSDs as well.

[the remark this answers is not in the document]
msimberg: I made it the default everywhere in https://github.com/pika-org/pika/pull/730.

msimberg:
The documentation now lives on https://pikacpp.org.

docs/usage.rst
msimberg: #729.

PR #724, opened 2023-07-31
Add release notes and update versions for pika 0.17.0

aurianer:


CHANGELOG.md
```diff
@@ -4,6 +4,23 @@
 <!--- Distributed under the Boost Software License, Version 1.0. (See accompanying -->
 <!--- file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt) -->
 
+## 0.17.0 (2023-08-02)
+
+### New features
+
+- Add `pika::wait` to wait for the runtime to be idle. ([#704](https://github.com/pika-org/pika/pull/704))
+- The debugger can now be attached after printing failure information. Previously the debugger could be attached before printing the error information. ([#712](https://github.com/pika-org/pika/pull/712))
+- `handle_print_bind` now also prints the thread pool when a mask is empty to be consistent to the printing when a mask is set. ([#710](https://github.com/pika-org/pika/pull/710))
```
msimberg, line 13: Please refer to `--pika:print-bind` rather than `handle_print_bind` since the latter is just an internal helper, and the former is the actual option a user would use to print pool/binding information.

msimberg: E.g.
```suggestion
- `--pika:print-bind` now also prints the thread pool of a thread when thread binding is disabled to be consistent with the output when thread binding is enabled. ([#710](https://github.com/pika-org/pika/pull/710))
```

CHANGELOG.md
```diff
@@ -4,6 +4,23 @@
...
+- Allow to explicitly reset `{unique_,}any_sender`.  ([#719](https://github.com/pika-org/pika/pull/719))
+- Add `execution::unpack` sender adapter to unpack tuple-like types sent by predecessor senders. ([#721](https://github.com/pika-org/pika/pull/721))
+
+### Bugfixes
+
+- Fix warnings when CMake unity build is disabled. ([#697](https://github.com/pika-org/pika/pull/697))
+- Make pika masks consistent with hwloc's ones, pika now reports a full mask when it was previously reported as an empty mask. ([#710](https://github.com/pika-org/pika/pull/710))
```
msimberg, line 20: This is not quite accurate. We still internally store no thread binding as an empty mask. We just make sure they're in the same form for a comparison.

Since the internals aren't really important for the user, perhaps you could say something like: "Fix bogus error when `--pika:print-bind` and `--pika:bind=none` were used together". Details are in the PR after all.

CHANGELOG.md
```diff
@@ -4,6 +4,23 @@
...
 
+## 0.17.0 (2023-08-02)
+
+### New features
+
+- Add `pika::wait` to wait for the runtime to be idle. ([#704](https://github.com/pika-org/pika/pull/704))
+- The debugger can now be attached after printing failure information. Previously the debugger could be attached before printing the error information. ([#712](https://github.com/pika-org/pika/pull/712))
+- `handle_print_bind` now also prints the thread pool when a mask is empty to be consistent to the printing when a mask is set. ([#710](https://github.com/pika-org/pika/pull/710))
+- Allow to explicitly reset `{unique_,}any_sender`.  ([#719](https://github.com/pika-org/pika/pull/719))
+- Add `execution::unpack` sender adapter to unpack tuple-like types sent by predecessor senders. ([#721](https://github.com/pika-org/pika/pull/721))
```
msimberg, line 15:
```suggestion
- Add `execution::unpack` sender adaptor to unpack tuple-like types sent by predecessor senders. ([#721](https://github.com/pika-org/pika/pull/721))
```
only because we've used adaptor over adapter almost excusively so far, but not necessarily because it's better or more correct.

CHANGELOG.md
```diff
@@ -4,6 +4,23 @@
 <!--- Distributed under the Boost Software License, Version 1.0. (See accompanying -->
 <!--- file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt) -->
 
+## 0.17.0 (2023-08-02)
+
+### New features
+
+- Add `pika::wait` to wait for the runtime to be idle. ([#704](https://github.com/pika-org/pika/pull/704))
+- The debugger can now be attached after printing failure information. Previously the debugger could be attached before printing the error information. ([#712](https://github.com/pika-org/pika/pull/712))
```
msimberg, line 12: I would word this the converse: Failure information is now printed before attaching a debugger. I know the difference might sound trivial, but at least to me that puts the focus on the failure information rather than attaching the debugger.

msimberg (changes requested):
Thanks for putting this together!

Please don't forget https://github.com/pika-org/pika/pull/593. Could I ask you to check if there are other PRs/issues that weren't assigned to the 0.17.0 milestone as well?

PR #729, opened 2023-08-08
Rename `PIKA_WITH_GENERIC_CONTEXT_COROUTINES` to `PIKA_WITH_BOOST_CONTEXT`

msimberg:
C.f. https://github.com/pika-org/pika/pull/723.

What do you think? Should we keep the old name? Is the new name better?

_If_ we change this option I would update the spack package to have a new variant `boost_context` which is enforced to be exactly the same as `generic_coroutines` so that one can still use the old variant.

### DIFF
--- a/.jenkins/cscs/env-gcc-9.sh
+++ b/.jenkins/cscs/env-gcc-9.sh
@@ -16,7 +16,7 @@ spack_spec="pika@main +generic_coroutines arch=${spack_arch} %${spack_compiler} 
 configure_extra_options+=" -DPIKA_WITH_CXX_STANDARD=${cxx_std}"
 configure_extra_options+=" -DPIKA_WITH_MAX_CPU_COUNT=128"
 configure_extra_options+=" -DPIKA_WITH_MALLOC=system"
-configure_extra_options+=" -DPIKA_WITH_GENERIC_CONTEXT_COROUTINES=ON"
+configure_extra_options+=" -DPIKA_WITH_BOOST_CONTEXT=ON"
 configure_extra_options+=" -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=ON"
 
 build_extra_options+=" -j10"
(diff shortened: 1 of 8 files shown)


RMeli:
The name change sounds very sensible to me.

msimberg:
> Naive question: what is the reason behind renaming it?
> 
> Currently I would explain this option as "on some platform (e.g. macOS) you need generic context coroutines" and the user cares about them being implemented with `Boost.Context` just because they will need it as dependency.
> 
> Changing it to `*_WITH_BOOST_CONTEXT`, the option is more specific but it declares "I want a dependency" instead of "I need a feature".
> 
> Not sure what's best, without knowing much, probably I slightly lean towards a feature option instead of a dependency option.

It's a good question. I have a few reasons that make me want to rename it to _something_ (not necessarily the current suggestion), with the primary reason being:
- The current name is pretty weird. It has some correct keywords, but "generic context coroutines" doesn't really mean anything on its own. A full, possibly more correct, name would be something like "generic context switching implementation for stackful coroutines" but that's a bit long.

Other reasons:
- The implementation is as "generic" as Boost.Context's implementation is generic. Which means in practice pretty "generic" but still it depends on whatever Boost supports. I'd possibly be ok with using the word generic for something like https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0876r10.pdf, but then we might as well use `STD_FIBER_CONTEXT` or something. Hence I'd like to avoid "generic"/
- The word "coroutines" is ambiguous at least since C++20 came into existence (really before that as well), so I'd also want to avoid "coroutines" in the name.

The above leaves "context" as a useful word to include in the option name. And at that point my preference is to just be explicit and use "boost context". Contrast with `PIKA_WITH_{CUDA,HIP,MPI}` which also say which dependency to enable, not which feature to enable (i.e. GPU support or communication). `PIKA_WITH_MALLOC` suggests another option which could be something like `PIKA_WITH_CONTEXT_SWITCHING_IMPLEMENTATION={builtin,boost}`, but I don't foresee us adding other implementations (except a feature-detected `fiber_context` implementation maybe someday) so I'm not sure I like that path for only two options.

That said, I see where you're coming from. In this instance from the available options prefer explicitly calling out the dependency (especially because it's actually an important dependency to call out, Boost.Context is not header-only).

Does this convince you into the other direction?

msimberg:
I'll wait for feedback from @biddisco and @rasolca as well on this before merging.

albestro (approved):

Issue #568, opened 2023-01-25
Allow passing an explicit CPU mask


msimberg:
Currently pika will use a CPU mask (if available) set on the process to restrict the the number of threads used by its thread pools. In an application that uses OpenMP, if OpenMP is used before pika is initialized, the process mask may be reset to only one core/pu by OpenMP. In these cases pika has no way to recover the original process mask anymore.

One possible solution is to allow passing an explicit process mask as a command line/configuration option to override the process mask that pika can read by itself with hwloc.

Another possible solution is to have a global object read the process mask at an early stage before OpenMP has had time to reset it.

msimberg:
While #739 and #738 partially solve the problem here, they still require manual intervention by the user if OpenMP resets the main thread binding before pika can read it. We may still consider the following options to make sure the bindings work without user's having to do anything:
- Use the `.preinit_array` linker section. As far as I can tell these functions run before e.g. global constructors and would help us read the mask before OpenMP can reset it. This is obviously platform/linker/compiler(?)-specific and not entirely portable, but could help at least on most HPC systems. Some details: https://stackoverflow.com/questions/62293512/inject-a-preinit-array-section-in-an-elf-file-with-libelf. We could inject this in various different ways, each with different pros and cons. Since `.preinit_array` can only be used in executables, an explicit opt-in may be best, but we may be able to find ways to only add it to executables. We could 1. add a `.preinit_array` function in `pika/init.hpp` or in an explicitly included header `pika/workaround_openmp_problems.hpp`, 2. link a static library with a `.preinit_array` to executables either explicitly or implicitly.
- We could provide a wrapper script to set `PIKA_PROCESS_MASK` from `hwloc-get --get --taskset` so that users at least don't have to write the script themselves.
- We could read e.g. SLURM environment variables that contain the mask.
- We could write a plugin for SLURM that exports a `PIKA_PROCESS_MASK` environment variable.
- We could change OpenMP so that it sets the mask of the main thread lazily.

msimberg:
This is closed by #738 and #739. However, #743 still remains a problem for ease of use.

PR #739, opened 2023-08-14
Add `--pika:process-mask` command line option

msimberg:
And corresponding `pika.override_process_mask` configuration option.

Part of #568.

This includes #738. #738 should be merged first.

This allows overriding the process mask detected by pika at startup, for the cases where we are not able to read the mask before e.g. OpenMP binds the main thread to a single core. The option accepts masks in hexadecimal form. The mask itself can be bigger than the number of cores available on the system, but I've now made it stop with an error if there are set bits outside of the number of available cores. E.g. on a 4-core system a mask of `0x0000f` is allowed, but `0x1000f` is not. Related, leading zeros are counted towards the size of the mask, but the internal check in `topology` normalizes it to the number of available cores (with a check).

I currently only parse hexadecimal masks. Binary literals `0b...` are not supported. Literals must have at least one digit, i.e. `0x` is not a valid mask. `--pika:override-process-mask` is currently ignored if `--pika:ignore-process-mask` is set.

In an environment where the process mask has been set, `hwloc-bind` can be used to read the process mask before the pika-enabled application is run and one can then pass that to `--pika:override-process-mask`. E.g.:
```
hello_world --pika:override-process-mask=$(hwloc-bind --get --taskset)
```

Note, the `--taskset` option is important. Otherwise `hwloc-bind` prints the hwloc-specific format, which may include multiple masks separated by commas.

The `taskset` command can also be used but, as far as I'm aware, it needs some additional parsing of the output. If on `bash`, one can get the mask of the current shell like this:
```
> taskset --pid $BASHPID
pid 703180's current affinity mask: ff
```
where `ff` would need to be extracted into `0xff` for use with pika.

Alternatively, with e.g. slurm one may also find the mask in an environment variable such as `SLURM_CPU_BIND_LIST`:
```
> srun -N 1 -n 4 -c 9 --cpu-bind=core env | rg BIND
[snip]
SLURM_CPU_BIND_LIST=0x00000001F00000001F,0x0007C00000007C0000,0x0000003E00000003E0,0x00F8000000
[snip]
```
Extracting the mask from that list is currently left as an exercise for the user, though with enough requests I may consider adding logic for reading that from the environment. For the moment I prefer to keep the functionality as simple as possible in pika though.

Finally, with `hwloc-calc` one can in fact replace the previously removed binding grammar that could be passed to `--pika:bind`. E.g. this can be passed to `--pika:override-process-mask`:
```
> hwloc-calc node:1.pu:2-4
0x000002a0
```

Alternative names: `--pika:force-process-mask`, `--pika:set-process-mask`, others?

Does the above sound reasonable? Do you think any of the constraints and behaviour should be changed?

### DIFF
--- a/libs/pika/command_line_handling/src/parse_command_line.cpp
+++ b/libs/pika/command_line_handling/src/parse_command_line.cpp
@@ -373,6 +373,13 @@ namespace pika::detail {
                   "values. Do not use with --pika:pu-step, --pika:pu-offset, or "
                   "--pika:affinity options. Implies --pika:numa-sensitive=1"
                   "(--pika:bind=none disables defining thread affinities).")
+                ("pika:process-mask", value<std::string>(),
+                  "a process mask in hexadecimal form to restrict cores available for "
+                  "the pika runtime. If a mask has been set externally on the executable, "
+                  "this option overrides that mask. Has no effect if "
+                  "--pika:ignore-process-mask is used. This option does not set the "
+                  "process mask for the main thread. The mask only affects threads spawned "
+                  "by the pika runtime.")
                 ("pika:ignore-process-mask", "ignore the process mask to restrict "
                  "available hardware resources, use all available processing units")
                 ("pika:print-bind",
--- a/libs/pika/runtime_configuration/src/runtime_configuration.cpp
+++ b/libs/pika/runtime_configuration/src/runtime_configuration.cpp
@@ -125,6 +125,7 @@ namespace pika::util {
 
             // add placeholders for keys to be added by command line handling
             "ignore_process_mask = 0",
+            "process_mask = ${PIKA_PROCESS_MASK:}",
             "os_threads = cores",
             "cores = all",
             "localities = 1",
--- a/libs/pika/command_line_handling/include/pika/command_line_handling/command_line_handling.hpp
+++ b/libs/pika/command_line_handling/include/pika/command_line_handling/command_line_handling.hpp
@@ -59,6 +59,7 @@ namespace pika::detail {
         std::string affinity_bind_;
         std::size_t numa_sensitive_;
         bool use_process_mask_;
+        std::string process_mask_;
         bool cmd_line_parsed_;
         bool info_printed_;
         bool version_printed_;
--- a/libs/pika/command_line_handling/src/command_line_handling.cpp
+++ b/libs/pika/command_line_handling/src/command_line_handling.cpp
@@ -109,6 +109,26 @@ namespace pika::detail {
         return "file(" + dest + ")";
     }
 
+    std::string handle_process_mask(detail::manage_config& cfgmap,
+        pika::program_options::variables_map& vm, bool use_process_mask)
+    {
+        std::string mask_string = cfgmap.get_value<std::string>("pika.process_mask", "");
+
+        if (vm.count("pika:process-mask"))
+        {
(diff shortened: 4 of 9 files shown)


msimberg:
bors try

msimberg:
bors try

libs/pika/command_line_handling/src/parse_command_line.cpp
```diff
@@ -373,6 +373,11 @@ namespace pika::detail {
                   "values. Do not use with --pika:pu-step, --pika:pu-offset, or "
                   "--pika:affinity options. Implies --pika:numa-sensitive=1"
                   "(--pika:bind=none disables defining thread affinities).")
+                ("pika:override-process-mask", value<std::string>(),
+                  "a process mask in hexadecimal form to override a mask on the "
+                  "executable. Has no effect if --pika:ignore-process-mask is used. "
+                  "This option does not set the process mask for the main thread. "
+                  "The mask only affects threads spawned by the pika runtime.")
```
msimberg, line 380: Should this simply be called `--pika:process-mask`? In cases where there is no special mask set (like on a desktop or laptop) it's not really overriding anything (just "overriding" the full machine mask which is always there if nothing else is set). Another option is `--pika:set-process-mask` but the `set` feels superfluous to me. Most command line options set something...

libs/pika/runtime_configuration/src/runtime_configuration.cpp
```diff
@@ -125,6 +125,7 @@ namespace pika::util {
 
             // add placeholders for keys to be added by command line handling
             "ignore_process_mask = 0",
+            "override_process_mask = ",
```
msimberg, line 128:
```suggestion
            "override_process_mask = ${PIKA_OVERRIDE_PROCESS_MASK:}",
```
(or `PIKA_PROCESS_MASK` depending on https://github.com/pika-org/pika/pull/739/files#r1294344840).

msimberg: This would allow setting the mask with
```
PIKA_PROCESS_MASK=$(hwloc-bind --get --taskset) myapp
```
Especially useful if pika doesn't get argc/argv.

msimberg:
bors try

msimberg:
bors try

msimberg:
bors try

libs/pika/command_line_handling/src/parse_command_line.cpp
RMeli: I think it might be a better name, for the reason you suggest. It also makes a bit more sense that setting `--pika:process-mask` is ignored if `--pika:ignore-process-mask`, while the `--pika:override-process-mask` name is a bit more confusing in such context IMHO.

libs/pika/runtime_configuration/src/runtime_configuration.cpp
RMeli: This would be useful for use in CP2K.

msimberg: #743.

msimberg:
> LGTM! Should I add a point on #49 to add the examples you cited in the description of the PR? I think it could be useful

I added
> document thread binding behaviour from https://github.com/pika-org/pika/pull/739
earlier already. Did you have something else in mind?

PR #733, opened 2023-08-09
Update stdexec commit tested in CI to `48c52df0f81c6151eecf4f39fa5eed2dc0216204`

msimberg:
- [x] Update Piz Daint spack environments
- [x] Update CI container image
- [x] Check compilation with DLA-Future (https://github.com/eth-cscs/DLA-Future/pull/956)

PR #741, opened 2023-08-15
Replace bors with GitHub merge queues

msimberg:
This is a follow-up to #622 and let's us enable GitHub's merge queue. I'm again going to enable the feature with this PR but if it still turns out to be significantly worse than bors we can go back to bors, as long as it's still supported...

GitHub's merge queue feature uses a single list of required checks both for merging to main and for actually adding a PR to a merge queue. This means that one has to wait for the required checks to first pass on the PR. This PR updates some workflows to always run (even on pushes to PRs) and some workflows to simply be triggered but then report a "skipped" status to satisfy GitHub's branch protection rules. This let's us still have all checks run before an actual merge happens and have some checks run immediately on pushes. In addition to the "linter" checks and the performance tests that were already previously run on pushes, I've enabled
- the "fast" linux/debug build on GitHub actions
- all CircleCI checks
- the CUDA 12 and GCC 13 Jenkins configurations
to run on every push.

GitHub merge queue still doesn't have an equivalent of `bors try` so we can't actually (with the current setup) trigger all CI configurations to run without attempting to merge. Because of this I'm leaving the bors config in place as long as it still works, so `bors try` will keep working for now.


msimberg:
Note, I've disabled the `condition_variable2` test from running on the arm64 CircleCI jobs because of frequent failures... I've added a note to #191.

msimberg:
I've just merged this manually due to a test failure on the hipcc CI configuration. I will check the status of that configuration for a few pull requests. If it keeps failing it may be a problem introduced by #733.

PR #742, opened 2023-08-16
Use more appropriate resource classes for jobs on CircleCI

msimberg:
Uses the small class for things like configure and spellchecking to avoid wasting resources there, use the large class for building and running tests (we don't have access to larger than large on the free plan). inshpect uses the large class as well since it can run in parallel.


msimberg:
bors merge

PR #744, opened 2023-08-17
Clean up unused configuration options

msimberg:
Removes a number of configuration options that were previously required by distributed HPX but that we no longer need in pika.


msimberg:
test cscs-daint

PR #747, opened 2023-08-21
Update stdexec to `4b441f333adaeb8eb0a6c7f38d36af30fa7ab366`

### LINKED ISSUE #687: Investigate segfaults with stdexec in `ensure_started`
#678 updated the stdexec commit used in CI. As a result this test broke with stdexec: https://github.com/pika-org/pika/blob/main/libs/pika/execution/tests/unit/algorithm_ensure_started.cpp#L153-L164. Investigate why, and fix it in stdexec or pika.

msimberg:
Fixes #687.


msimberg:
@aurianer could I ask you to update the ault environment once again for this PR? Thank you!

msimberg:
bors try

msimberg:
bors try

msimberg:
As it turns out the HIP configuration does not like the new version of stdexec. I'll see if I can find the reason for it quickly, otherwise we might try to keep the old version of stdexec for that configuration for the time being.

msimberg:
I've updated the HIP CI configuration to use https://github.com/NVIDIA/stdexec/commit/5e378418d9360a1a53392baae8502c4961f5f4c8 since it uses GCC 12 and https://github.com/NVIDIA/stdexec/commit/4b441f333adaeb8eb0a6c7f38d36af30fa7ab366 includes https://github.com/NVIDIA/stdexec/commit/8f0e55469d410c91a4076c91dc50242a2f1672a7 which breaks compilation with GCC 12. I'm leaving the other CI configurations at https://github.com/NVIDIA/stdexec/commit/4b441f333adaeb8eb0a6c7f38d36af30fa7ab366 since they are unaffected.

PR #764, opened 2023-09-04
Add very early initialization of topology to work around OpenMP resetting process masks

msimberg:
This is a draft. I'm currently just testing if all compilers are happy with the `.preinit_array` section.

If there's a risk of this ever failing (compilation, linking, runtime) we should be quite conservative about this. If it's enabled by defualt, there must be a way to disable it so that one can fall back to using `--pika:process-mask`. Probably we need:
- A feature check for `.preinit_array`. Must do linking as well. However, this may still fail if the user application for some reason uses a linker that doesn't support the section. What's the failure case?
- A CMake flag? Off or on by defualt?
- A CMake target to explicitly link or automatically link to executables?
- A better name for the target.

### DIFF
--- a/cmake/pika_setup_target.cmake
+++ b/cmake/pika_setup_target.cmake
@@ -98,6 +98,8 @@ function(pika_setup_target target)
       ${target} PRIVATE "PIKA_APPLICATION_NAME=${name}"
                         "PIKA_APPLICATION_STRING=\"${name}\""
     )
+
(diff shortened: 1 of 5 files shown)


msimberg:
bors try

msimberg:
I'm going to close this for now. #834 covers the necessary use cases for now. It's more intrusive, but it doesn't rely on the linker doing the right thing. We can always revisit this if there are cases where wrapping executables in `pika-bind` isn't appropriate.

PR #765, opened 2023-09-04
Add release notes for 0.18.0

msimberg:


msimberg:
> LGTM thanks a lot! Should we add a point on the documentation?

I wasn't sure as it's not a code change, but maybe it doesn't hurt to highlight it in the release notes.

Issue #32, opened 2022-01-19
Use CSCS GitLab CI to replace as many Jenkins builds as possible


msimberg:
In practice this means replacing the non-Cray builds currently running on Jenkins. Work was started on https://github.com/STEllAR-GROUP/hpx-local/pull/5.

Make use of the gitlab matrix functionality at least for release/debug builds, if not for different compiler etc. configurations.

Also replace the HIP testing done with Jenkins with `container-runner-hohgant-mi200` (as already started in https://github.com/eth-cscs/DLA-Future/pull/982 for DLA-Future, though that is blocked on an MPI issue; we can initially run all non-MPI tests for pika).

msimberg:
This may be a useful starting point: https://github.com/electronic-structure/SIRIUS/blob/master/ci/.gitlab-ci.yml.

msimberg:
I'll close this as almost all CI configurations have now been moved to CSCS CI. We won't be getting a CPE configuration anytime soon, and moving the performance CI is tracked separately in https://github.com/pika-org/pika/issues/936.

PR #771, opened 2023-09-12
Add initial CI pipeline with CSCS-CI 

aurianer:
Fix partially #32.
Sets the initial pipeline using CSCS-CI with the docker base image.

Note: I didn't remove the `{SOURCE,BUILD}_DIR` defaults in the docker image as it makes it easier to run it locally.

### DIFF
--- a/.gitlab/pipeline.yml
+++ b/.gitlab/pipeline.yml
@@ -41,6 +41,7 @@ clang14_debug_build:
   variables:
     SLURM_CONSTRAINT: mc
     SLURM_PARTITION: normal
+    PROCESSOR_COUNT: 36
 
 .test_common_gpu:
   extends: .test_common
@@ -48,6 +49,7 @@ clang14_debug_build:
   variables:
     SLURM_CONSTRAINT: gpu
     SLURM_PARTITION: normal
+    PROCESSOR_COUNT: 12
 
 clang14_debug_test:
   extends: .test_common_mc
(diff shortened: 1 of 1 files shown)


.gitlab/pipeline.yml
```diff
@@ -0,0 +1,58 @@
...
+    SLURM_PARTITION: normal
+
+clang14_debug_test:
+  extends: .test_common_mc
+  needs:
+    - clang14_debug_build
+  image: $CSCS_REGISTRY_PATH/pika-debug-cpu-clang14:$CI_COMMIT_SHORT_SHA
+  script:
+    - cd /pika/build
+    - ctest --verbose -j2 --timeout 120 --output-on-failure --no-compress-output -R tests --exclude-regex "tests.unit.modules.executors.std_thread_scheduler"
```
msimberg, line 55: Could you increase `-j2` to something higher? I think we should be able to go quite high since we're on the multicore partition. You can also check how many tests we run concurrently in the Jenkins configurations.

.gitlab/docker/Dockerfile.build
```diff
@@ -0,0 +1,29 @@
+ARG BASE_IMAGE=docker.io/pikaorg/pika-ci-base:17
+ARG SOURCE_DIR=/pika/source
+ARG BUILD_DIR=/pika/build
+
+FROM $BASE_IMAGE
+
+ARG SOURCE_DIR
+ARG BUILD_DIR
```
msimberg, line 8: Maybe?
```suggestion
ARG BASE_IMAGE
FROM $BASE_IMAGE

ARG SOURCE_DIR=/pika/source
ARG BUILD_DIR=/pika/build
```
At least we don't need `SOURCE_DIR` and `BUILD_DIR` twice as `ARG`s. Do you think it makes sense to move the definitions of `SOURCE_DIR` and `BUILD_DIR` to `pipeline.yml` and not have any default values here? Not a blocker.

.gitlab/pipeline.yml
```diff
@@ -0,0 +1,60 @@
...
+    SLURM_PARTITION: normal
+    PROCESSOR_COUNT: 12
+
+clang14_debug_test:
+  extends: .test_common_mc
+  needs:
+    - clang14_debug_build
+  image: $CSCS_REGISTRY_PATH/pika-debug-cpu-clang14:$CI_COMMIT_SHORT_SHA
+  script:
+    - cd /pika/build
```
msimberg, line 56: I don't know if environment variables are passed into the container at runtime, but if they do I guess this could be
```suggestion
    - cd $BUILD_DIR
```
Not a blocker.

msimberg (approved):
Thanks for all the work so far on this @aurianer!

This is ready to be used already in my opinion and you can go ahead and merge it if you like.

I've added two more potential cleanup comments, but I'm not sure both of them are possible to do, and they are not blockers. They can be done at a later point as well.

For later: we should add this to the GitHub merge queue required checks, but at the moment I think this pipeline will not be triggered because of the merge queue branch name.

msimberg (approved):

Issue #745, opened 2023-08-17
Clean up configuration and command line handling


msimberg:
The runtime configuration is currently in a pretty messy state with many ad-hoc workarounds. It parses configurations multiple times, environment variables provide defaults instead of overriding configuration options, and many options are ignored if put in the wrong place (see #743). I would propose that we streamline the handling of configuration and command line options.

The following is an attempt to summarize an "ideal" flow of processing options, attempting to be as close as possible to the current behaviour. A later step always overwrites options set from previous steps.

- Read a default configuration hardcoded in a source file or in a separate configuration (read-only) file stored on disk
- Read a configuration file from `PIKA_CONFIG` if set, otherwise read a configuration file from `~/.config/pika/config.ini` if exists
- Read configuration options from `PIKA_INI` if set
- Read a configuration file from `--pika:config` if set
- Read configuration options from `--pika:ini` if set
- Read configuration options from other environment variables, such as `PIKA_PROCESS_MASK`
- Read configuration options from the command line, such as `--pika:process-mask`
- Normalize, expand, replace options if that's something we want to support (we currently support referring to other configuration options)

An alternative workflow collapses some of the handling of configuration files:

- Read a default configuration hardcoded in a source file or in a separate configuration (read-only) file stored on disk
- Read a configuration file from `--pika:config` if set, otherwise from `PIKA_CONFIG` if set, otherwise from `~/.config/pika/config.ini` if exists
- Read configuration options from `--pika:ini` if set, otherwise from `PIKA_INI` if set
- Read configuration options from other environment variables, such as `PIKA_PROCESS_MASK`
- Read configuration options from the command line, such as `--pika:process-mask`
- Normalize, expand, replace options if that's something we want to support (we currently support referring to other configuration options)

A third, yet simpler worflow that leaves out the `--pika:ini`/`PIKA_INI` options could look like this:

- Read a default configuration hardcoded in a source file or in a separate configuration (read-only) file stored on disk
- Read a configuration file from `--pika:config` if set, otherwise from `PIKA_CONFIG` if set, otherwise from `~/.config/pika/config.ini` if exists
- Read configuration options from other environment variables, such as `PIKA_PROCESS_MASK`
- Read configuration options from the command line, such as `--pika:process-mask`
- Normalize, expand, replace options if that's something we want to support (we currently support referring to other configuration options)

[...]

PR #775, opened 2023-09-15
Add workaround to ensure `PIKA_PROCESS_MASK` is respected

### LINKED ISSUE #743: Make sure `PIKA_PROCESS_MASK` is actually honored by the runtime configuration
https://github.com/pika-org/pika/pull/739#discussion_r1295965920

Also make sure to add documentation about `PIKA_PROCESS_MASK`.

msimberg:
Fixes #743. However, this only special cases the handling for `PIKA_PROCESS_MASK` and doesn't fix the behaviour for any other configuration options. The proper fix would require doing #745. This also adds an additional check for empty process masks when set through `PIKA_PROCESS_MASK` or `--pika:process-mask` to give a better error message when the mask is empty.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -142,12 +142,14 @@ how many threads to use for the runtime. ``hwloc-bind`` can also be used to manu
 mask on the application. When a process mask is set, the default behaviour is to use only one thread
 per core in the process mask. Setting ``--pika:threads`` to a number higher than the number of cores
 available in the mask is not allowed. Using ``--pika:threads=all`` will use all the hyperthreads in
-the process mask. The process mask can explicitly be ignored with the option
-``--pika:ignore-process-mask``. In that case pika behaves as if no process mask is set. In addition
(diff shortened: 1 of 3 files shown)


RMeli (approved):
Thanks @msimberg! Is it worth adding it to the documentation?

msimberg:
> Thanks @msimberg! Is it worth adding it to the documentation?

Yes, thanks for the reminder!

msimberg:
@RMeli added documentation, would you mind checking if the explanation is clear to you?

In theory `--pika:ignore-process-mask` could get the same treatment with an environment variable workaround. @RMeli do you think you will be needing `PIKA_IGNORE_PROCESS_MASK` at some point soon, or do you think we can wait with a workaround to see if we can do #745 before the workaround?

RMeli (approved):
Thanks @msimberg. Documentation is clear, and specifies the precedence which is great. LGTM!

> or do you think we can wait with a workaround to see if we can do https://github.com/pika-org/pika/issues/745 before the workaround?

I think we can wait and eventually adapt if the need arises.

PR #779, opened 2023-09-19
Change one Jenkins CI configuration to use C++23

### LINKED ISSUE #756: Add support for C++23 to CMake configuration and spack package
This should "just work" with `PIKA_WITH_CXX_STANDARD` but there are no CI configurations testing it.

aurianer:
Fixes #756 once https://github.com/spack/spack/pull/40078 is merged

### DIFF
--- a/.jenkins/cscs/env-clang-16.sh
+++ b/.jenkins/cscs/env-clang-16.sh
@@ -4,7 +4,7 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-cxx_std="20"
+cxx_std="23"
 clang_version="16.0.2"
 boost_version="1.82.0"
 hwloc_version="2.9.1"


cmake/pika_check_cxx_standard.cmake
```diff
@@ -31,11 +31,5 @@ endif()
 set(CMAKE_CXX_STANDARD ${PIKA_WITH_CXX_STANDARD})
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-# We explicitly set the default to 98 to force CMake to emit a -std=c++XX flag.
-# Some compilers (clang) have a different default standard for cpp and cu files,
-# but CMake does not know about this difference. If the standard is set to the
-# .cpp default in CMake, CMake will omit the flag, resulting in the wrong
-# standard for .cu files.
-set(CMAKE_CXX_STANDARD_DEFAULT 98)
```
msimberg, line 39: I'd like to keep these around. Even though we explicitly now use clang as  `CUDA` compiler, NVHPC is still a `CXX`-only compiler but it still compiles `.cu` files.

Ignoring any issues with CMake doing the wrong thing, I still like to have the flag explicitly emitted. Explicit is usually better than implicit.

.jenkins/cscs/env-clang-16.sh
```diff
@@ -4,7 +4,7 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-cxx_std="20"
+cxx_std="23"
```
msimberg, line 7: Did you test this with the spack package? Doesn't the spack package require updating first? I'm happy to make the change for the Piz Daint spack instance until it's changed upstream. It looks like the boost and fmt packages also need a 23 `cxxstd` added. They'd need to be tested as well.

msimberg (approved):

msimberg:
@aurianer I've rebuilt the clang 16 configuration on Piz Daint with C++23 now. Let's give this a go.

Issue #777, opened 2023-09-18
Fix the segfault happening on CSCS CI for `tests.unit.modules.executors.std_thread_scheduler`


aurianer:
Segfault happening on CSCS CI configuration with pika-ci-image:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/5110101781

msimberg:
Closing this as #784 was merged. The test does at least not seem to fail _every_ run. The test is still listed in #191 so it's not completely ignored if it keeps failing occasionally.

PR #784, opened 2023-09-20
Add first GCC 11 spack pipeline with CSCS CI

aurianer:
Fix partially #32.
- Use spack in the pipeline instead of the pika-org/pika-ci-image.
- 3 stages `spack_base`, `spack_configs`, `build` and `test`. The `spack_base` stage installs spack and common utilities. The `spack_configs` installs pika dependencies for our set of CI spack spec (for now just gcc-11). The container images of those 2 stages are supposed to be rarely rebuilt (only at a change of spack version used or at a change of configuration, e.g. compiler/build_type/etc.). The `build` and `test` stages are the ones running at each push.

TODO:
- [x] replace processor count with `nproc`
- [x] factorize build and test image tags

### DIFF
--- a/.gitlab/pipeline.yml
+++ b/.gitlab/pipeline.yml
@@ -123,7 +123,7 @@ gcc11_debug_test:
     - gcc11_debug_build
   script:
     - spack arch
-    - spack build-env $spack_spec -- bash -c "ctest --label-exclude COMPILE_ONLY --test-dir ${BUILD_DIR} --verbose -j$(nproc) --timeout 120 --output-on-failure --no-compress-output -R tests --exclude-regex tests.unit.modules.executors.std_thread_scheduler tests.unit.build"
+    - spack build-env $spack_spec -- bash -c "ctest --label-exclude COMPILE_ONLY --test-dir ${BUILD_DIR} --verbose -j$(nproc) --timeout 120 --output-on-failure --no-compress-output -R tests --exclude-regex tests.unit.build"
   image: $PERSIST_IMAGE_NAME
 
 ###################################################################################################
@@ -147,5 +147,4 @@ clang14_debug_test_pika-ci-image:
   image: $CSCS_REGISTRY_PATH/pika-debug-cpu-clang14:$CI_COMMIT_SHORT_SHA
   script:
     - ctest --test-dir ${BUILD_DIR} --verbose -j$(nproc) --timeout 120 --output-on-failure \
-      --no-compress-output -R tests \
-      --exclude-regex "tests.unit.modules.executors.std_thread_scheduler"
+      --no-compress-output -R tests


.gitlab/docker/Dockerfile.spack_base
```diff
@@ -0,0 +1,47 @@
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+ARG SPACK_REPO=https://github.com/spack/spack
+ARG SPACK_VERSION
```
msimberg, line 5: This applies to all dockerfiles: could you please move the ARGs close to where they are actually used? Since every ARG creates a new layer, changing an ARG invalidates all later layers even if an ARG isn't necessarily used by the immediate next RUN commands.

.gitlab/docker/Dockerfile.spack_base
```diff
@@ -0,0 +1,47 @@
...
+    xz-utils \
+  && rm -Rf /var/lib/apt/lists/*
+
+# Install spack
+RUN mkdir -p $SPACK_ROOT \
+    && git clone --depth 1 -b ${SPACK_VERSION} ${SPACK_REPO} ${SPACK_ROOT}
+
+ENV PATH $SPACK_ROOT/bin:$PATH
+
+# Install utility
```
msimberg, line 43: What does this comment say? Nothing is installed by the spack commands below.

.gitlab/docker/Dockerfile.spack_base
```diff
@@ -0,0 +1,47 @@
...
+RUN spack external find --scope site
+RUN spack compiler find --scope site
```
msimberg, line 45: Can these be combined into a single layer?

.gitlab/docker/Dockerfile.spack_base
```diff
@@ -0,0 +1,47 @@
...
+
+ENV CMAKE_GENERATOR=Ninja
```
msimberg, line 47: Does this need to be in the base image? Or does it need to be set at all? We should definitely use ninja for building pika but it might be enough to use `-GNinja`?

.gitlab/docker/Dockerfile.spack_build
```diff
@@ -0,0 +1,17 @@
...
+ARG BUILD_DIR=/pika/build
+
+ARG CMAKE_COMMON_FLAGS
+ARG CMAKE_FLAGS
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+COPY . ${SOURCE_DIR}
+
+SHELL ["/bin/bash", "-c"]
```
msimberg, line 14: Here and elsewhere: is this required? Good to set explicitly?

.gitlab/docker/Dockerfile.spack_build
```diff
@@ -0,0 +1,17 @@
...
+
+# Configure & Build
+RUN spack build-env $spack_spec -- bash -c "cmake -B${BUILD_DIR} ${SOURCE_DIR} $CMAKE_COMMON_FLAGS $CMAKE_FLAGS && cmake --build ${BUILD_DIR} --target tests"
```
msimberg, line 17: Could you please also explicitly build the `examples` or `all` targets? There are a few examples that are not enabled as tests, and they would not get build with just `tests`.

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,32 @@
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+ARG ARCH
+ARG BOOST
+ARG BUILD_TYPE
+ARG COMPILER
+ARG CXXSTD
+ARG HWLOC
```
msimberg, line 9: See comment about using ARGs close to where they're used.

.gitlab/docker/Dockerfile.spack_build
```diff
@@ -0,0 +1,17 @@
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+ARG SOURCE_DIR=/pika/source
+ARG BUILD_DIR=/pika/build
+
+ARG CMAKE_COMMON_FLAGS
+ARG CMAKE_FLAGS
+
+ENV DEBIAN_FRONTEND=noninteractive
```
msimberg, line 10: Necessary? It's already set, and you don't use apt anymore after the base image right?

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,32 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
```
msimberg, line 11: See other comment about not using apt anymore this point.

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,32 @@
...
+
+SHELL ["/bin/bash", "-c"]
+
+# Disable host compatibility to be able to compile for broadwell on zen2
+RUN echo -e "concretizer:\n  targets:\n    host_compatible: false" \
+    > $SPACK_ROOT/etc/spack/concretizer.yaml
```
msimberg, line 17: I think this can be done with `spack config add ...`. The json can also be written without newlines: `concretizer:targets:host_compatible:false`. I think the comment above is also unnecessarily specific. It'll be compiling for haswell, zen3, etc. in the future as well.

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,32 @@
...
+
+# Install compiler
+RUN export spack_compiler=${COMPILER/clang/llvm} && \
+    spack install $spack_compiler languages=c++ arch=$ARCH && \
+    spack compiler add --scope site $(spack location -i $spack_compiler)
+
+ENV spack_spec "pika@main build_type=$BUILD_TYPE arch=$ARCH %$COMPILER malloc=system cxxstd=$CXXSTD ^$BOOST ^$HWLOC"
```
msimberg, line 24: Could this be a single `ARG SPACK_SPEC` instead? Then we'd reduce the number of args, and make it easier to explicitly specify e.g. the fmt version, malloc, etc. in the pipeline config file directly instead of adding an ARG for each dependency.

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,32 @@
...
+
+RUN echo "spack arch: $(spack arch)"
+
+# Print spec
+RUN spack spec -lI $spack_spec
+
+# Install dependencies for this configuration
+RUN spack install --fail-fast --only dependencies $spack_spec
```
msimberg, line 32: These can be a single layer.

.gitlab/docker/Dockerfile.spack_compiler
msimberg, line 24: I would leave out the build type at this stage. All the dependencies are built with a build type independent of the pika build type so if we leave it out here we can use the same dependencies images for release and debug builds.

.gitlab/pipeline.yml
```diff
@@ -10,56 +10,138 @@ include:
...
+  extends:
+    - .container-builder
+    - .variables_gcc11_config
+    - .cmake_variables_common
+  needs:
+    - gcc11_spack_image
+  before_script:
+    - configuration=$CMAKE_COMMON_FLAGS-$CMAKE_FLAGS-$CI_COMMIT_SHORT_SHA
+    - configuration=${configuration//-D/}
+    - CONFIG_TAG=`echo $configuration | sha256sum - | head -c 8`
```
msimberg, line 84:
```suggestion
    - CONFIG_TAG=`echo $configuration | sha256sum - | head -c 16`
```

msimberg (changes requested):
This looking pretty good! Cosmetic changes remain, and of course the test job needs to be fixed, but it's getting very close.

.gitlab/pipeline.yml
```diff
@@ -10,56 +10,138 @@ include:
...
+.cmake_variables_common:
+  variables:
+    CMAKE_COMMON_FLAGS: "-DPIKA_WITH_COMPILER_WARNINGS=ON -DPIKA_WITH_COMPILER_WARNINGS_AS_ERRORS=ON -DPIKA_WITH_CHECK_MODULE_DEPENDENCIES=ON -DPIKA_WITH_EXAMPLES=ON -DPIKA_WITH_TESTS=ON"
+
+.variables_gcc11_config:
+  variables:
+    ARCH: linux-ubuntu22.04-broadwell
+    BOOST: boost@1.78.0
+    BUILD_TYPE: Debug
+    COMPILER: gcc@11.2.0
```
msimberg, line 32: Why not 11.4.0?

.gitlab/pipeline.yml
```diff
@@ -10,56 +10,138 @@ include:
...
+
+gcc11_debug_test:
+  extends:
+    - .cmake_variables_common
+    - .variables_gcc11_config
+    - .test_common_mc
+  needs:
+    - gcc11_debug_build
+  script:
+    - spack build-env $spack_spec -- bash -c ctest --test-dir ${BUILD_DIR} --verbose \
```
msimberg, line 122: Is `--test-dir` a ctest option?

.gitlab/docker/Dockerfile.spack_build
msimberg: This doesn't necessarily need to be removed. I was just curious what the benefit is of setting it.

> since it makes the next argument evaluated as $0 in bash

I'm not sure I understand what this is referring to. Do you have an example?

.gitlab/docker/Dockerfile.spack_compiler
msimberg: The following at least works for me locally on a recent spack develop (see `<--`, added by me to the output):
```
~> spack config blame concretizer                                                                                                           [0/0]
---                                                         concretizer:
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:18    reuse: dependencies
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:23    targets:
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:28      granularity: microarchitectures
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:32      host_compatible: True <---
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:38    unify: True
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:40    duplicates:
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:44      strategy: none
~> spack config --scope defaults add "concretizer:targets:host_compatible:false"
~> spack config blame concretizer
---                                                         concretizer:
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:7     reuse: dependencies
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:12    targets:
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:17      granularity: microarchitectures
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:18      host_compatible: False <---
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:19    unify: True
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:21    duplicates:
/home/mjs/src/spack/etc/spack/defaults/concretizer.yaml:25      strategy: none
```

.gitlab/docker/Dockerfile.spack_compiler
msimberg: I mean you can still do
```
ARG SPACK_SPEC
ENV SPACK_SPEC=$SPACK_SPEC
```
if you want that behaviour. My comment was mainly about how and where we construct the spec.

msimberg: Or as a dotenv. Whichever you find nicer to work with.

.gitlab/pipeline.yml
```diff
@@ -10,56 +10,140 @@ include:
...
+      -j$(nproc) --timeout 120 --output-on-failure --no-compress-output -R tests \
+      --exclude-regex "tests.unit.modules.executors.std_thread_scheduler"
```
msimberg, line 126: By the way, I think it would make sense already at this point to move the compile-only tests to the build/configure stage to build them on the rosa build farm instead of on daint.

.gitlab/pipeline.yml
msimberg: Ok, but that doesn't mean we should downgrade the compiler. I can think of at least ways forward here:
- Check if the system compiler is the compiler we want to install through spack, and if it is we skip building the compiler through spack.
- Check if we can install the same version alongside the system compiler. I'm not quite sure what spack will think of this.
I think the second option would be interesting to know if it works and nice since then we can control exactly how the compiler is built, but this one is more work. I'd be happy with the first option. Finally, I'd also be happy to have this done in a follow-up PR. This is not a blocker, I just wanted to know why we weren't using the latest version in the 11.X series. If you prefer to do this separately, please open an issue for it.

.gitlab/pipeline.yml
msimberg: Yes please.

.gitlab/docker/Dockerfile.spack_build
```diff
@@ -0,0 +1,20 @@
...
+
+# Configure & Build
+RUN spack build-env $spack_spec -- bash -c "cmake -B${BUILD_DIR} ${SOURCE_DIR} \
+    $CMAKE_COMMON_FLAGS $CMAKE_FLAGS && cmake --build ${BUILD_DIR} --target all tests examples"
+
+# Run compile only tests and tests.unit.build
+RUN spack build-env $spack_spec -- bash -c "ctest -L COMPILE_ONLY --test-dir ${BUILD_DIR} \
+    --verbose -j$(nproc) --timeout 30 --output-on-failure --no-compress-output -R tests \
+    && ctest --test-dir ${BUILD_DIR} --verbose -j$(nproc) --timeout 15 --output-on-failure \
+    --no-compress-output -R tests.unit.build"
```
msimberg, line 20: Let's go with this for now, but see https://github.com/pika-org/pika/issues/816 for a potential follow-up. I think `nproc` may be overkill here, but it's not a blocker as long as we don't kill the build nodes by using too much memory...

.gitlab/pipeline.yml
```diff
@@ -8,56 +8,146 @@ include:
...
+    - .cmake_variables_common
+    - .variables_gcc11_config
+    - .test_common_mc
+  needs:
+    - gcc11_debug_build
+  script:
+    - spack arch
+    - spack build-env $spack_spec -- bash -c "ctest -LE COMPILE_ONLY --test-dir ${BUILD_DIR} \
+      --verbose -j$(nproc) --timeout 120 --output-on-failure --no-compress-output -R tests \
+      --exclude-regex tests.unit.modules.executors.std_thread_scheduler tests.unit.build"
```
msimberg, line 128: I think after this PR you can simply try removing this exclusion (i.e. do #777). It's possible that #797 or #795 were the cause of the segfaults.

cmake/pika_add_compile_test.cmake
```diff
@@ -5,7 +5,7 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 function(pika_add_compile_test category name)
-  set(options FAILURE_EXPECTED NOLIBS OBJECT)
+  set(options FAILURE_EXPECTED NOLIBS OBJECT COMPILE_ONLY)
```
msimberg, line 8: Are not all tests added through `pika_add_compile_test` `COMPILE_ONLY`? In other words, I don't think you need the option here. Then you can revert the explicit flags you've added in the modules.

cmake/pika_add_compile_test.cmake
```diff
@@ -50,6 +50,10 @@ function(pika_add_compile_test category name)
     WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
   )
 
+  if(${${name}_COMPILE_ONLY})
+    set_tests_properties("${category}.${name}" PROPERTIES LABELS "COMPILE_ONLY")
+  endif()
```
msimberg, line 55:
```suggestion
  set_tests_properties("${category}.${name}" PROPERTIES LABELS "COMPILE_ONLY")
```

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,25 @@
...
+RUN spack compiler info $COMPILER > /dev/null 2> /dev/null; compiler_missing=$?; \
+    export spack_compiler=${COMPILER/clang/llvm} && \
+    if [[ $compiler_missing != 0 ]]; then \
+        spack install $spack_compiler languages=c++ arch=$ARCH && \
+        spack compiler add --scope site $(spack location -i $spack_compiler);\
+    fi
+
+RUN echo "spack arch: $(spack arch)"
+
+# Print spec
```
msimberg, line 21:
```suggestion
```

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -0,0 +1,25 @@
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+SHELL ["/bin/bash", "-c"]
+
+# Disable host compatibility to be able to compile for broadwell on zen2
```
msimberg, line 6: I should perhaps been more specific about this comment: this will be compiling for other architectures in the future. Instead of having a comment about which architectures we're targeting, the comment should explain _why_ the architecture will/may be different (builds are done on a different system from where tests are run).

msimberg (changes requested):
All good now except for the `COMPILE_ONLY` label and a couple of very minor cosmetic cleanups. I would simply add the `COMPILE_ONLY` unconditionally to all compile-only tests unless there is a good reason to make it opt-in?

Remaining comments

msimberg (approved):
Thanks @aurianer for all the updates here. This is blocked because of CI being down... but let's hit merge as soon as those come back up.

msimberg:
@aurianer FYI, something's still not right with running the tests on the last run: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/5375604892#L65.

PR #797, opened 2023-10-04
Fix one more lifetime bug in `ensure_started` for cases when the sender is dropped without connecting it

msimberg:
If the `ensure_started` sender is not connected (i.e. there are no continuations) the `ensure_started_receiver` holds the last reference to the shared state. This will be released when releasing the operation state until the `ensure_started` operation state, since the `ensure_started_receiver` will be released with the operation state. This happens because the `ensure_started_receiver` is taken by rvalue reference in `set_*`. This PR changes them to take the receiver by value, ensuring that the reference count stays non-zero also after the operation state has explicitly been reset.

This is another way to fix #795 but I'm keeping that change as it doesn't hurt. Importantly, this fixes a distinct bug that is reported by valgrind in the `std_thread_scheduler` test (the test that calls `ensure_started` but drops the sender immediately). I've enabled valgrind testing with the `std_thread_scheduler_test` in this PR.

### DIFF
--- a/.github/workflows/linux_valgrind.yml
+++ b/.github/workflows/linux_valgrind.yml
@@ -58,6 +58,7 @@ jobs:
       shell: bash
       run: |
           cmake --build build --target tests.unit.modules.execution
+          cmake --build build --target std_thread_scheduler_test
     - name: Test
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


msimberg:
The failure doesn't reproduce every run in CI, but has now reproduced 2/3 times. I'm going to push the fix next.

PR #798, opened 2023-10-04
Add linter for yaml files

aurianer:
Attempt to give more useful error messages than CSCS-CI when the yml file is not correct. Only lints modified files.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,25 @@
+# Copyright (c) 2023 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+name: Lint Yaml Files
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Lint Yaml Files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Lint Yaml files
+        uses: github/super-linter/slim@v5
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          VALIDATE_YAML: true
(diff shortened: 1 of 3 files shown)


.github/linters/.yaml-lint.yml
```diff
@@ -0,0 +1,9 @@
+---
```
msimberg, line 1:
```suggestion
# Copyright (c) 2023 ETH Zurich
#
# SPDX-License-Identifier: BSL-1.0
# Distributed under the Boost Software License, Version 1.0. (See accompanying
# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)

---
```

.github/linters/.yaml-lint.yml
```diff
@@ -0,0 +1,9 @@
...
+
```
msimberg, line 2: How is this configuration file picked up by yaml-lint/super-linter? I see no mention of it in the workflow or the documentation, but I've probably just missed it.

msimberg (changes requested):

.github/workflows/linter.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  build:
+    name: Lint Yaml Files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
msimberg, line 19: Why does this need to be set? Presumably that's what triggers "You are not currently on a branch" (https://github.com/pika-org/pika/pull/798/checks#step:4:72). I don't know if it matters that it's not on a branch.

msimberg: Thanks for the reference, all good. So this is set to allow super linter to get the list of changed files only (from the example in the super linter README: https://github.com/super-linter/super-linter#example-connecting-github-action-workflow):
```
      - name: Checkout Code
        uses: actions/checkout@v4
        with:
          # Full git history is needed to get a proper
          # list of changed files within `super-linter`
          fetch-depth: 0
```

msimberg:
I've added the linter check to the required checks for merges.

msimberg (approved):

PR #801, opened 2023-10-05
Add a test in `drop_operation_state` with a predecessor sender sending a tuple

aurianer:
When piping a `just` sender of a `tuple` into a `drop_operation_state` sender, the values resulting were the one contained in the `tuple` directly. It has to do with the fact that `std::tuple<std::tuple<Ts...>>` is a `std::tuple<Ts>` which was later discarded by the `std::apply` used in the `drop_operation_state`'s receiver `set_value`. We ensure in this PR that the nesting is conserved.

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
@@ -87,7 +87,7 @@ namespace pika::drop_op_state_detail {
 
             try
             {
-                auto ts_local = std::tuple(PIKA_FORWARD(Ts, ts)...);
+                std::tuple<std::decay_t<Ts>...> ts_local(PIKA_FORWARD(Ts, ts)...);
                 r.op_state->op_state.reset();
 
                 std::apply(pika::util::detail::bind_front(pika::execution::experimental::set_value,


libs/pika/execution/tests/unit/algorithm_drop_operation_state.cpp
```diff
@@ -165,5 +165,11 @@ int main()
 
     test_adl_isolation(ex::drop_operation_state(my_namespace::my_sender{}));
 
+    {
+        auto s = ex::just(std::tuple<int, double>()) | ex::drop_operation_state() |
+            ex::then([](std::tuple<int, double>&&) {});
+        tt::sync_wait(std::move(s));
+    }
+
```
msimberg, line 173: Would you mind restructuring this test in the same form as the tests from line 26 to 107 (you can move it up there as well). Maybe change one of the tuple types to `std::string` so we have one type where copy isn't the same as move?

msimberg (changes requested):

msimberg (approved):
Thanks!

We may want to consider a patch release of 0.19 with this next week.

msimberg:
Approved, modulo formatting and inshpect failures.... I think auto-merge will keep it in the queue if you fix those.

PR #804, opened 2023-10-09
Check if release tag already exists before tagging

aurianer:
When using a `GITHUB_TOKEN` with the wrong permissions, the tag is created but the release is not pushed. This allows to rerun the script to push the release even if the tag already exists.

### DIFF
--- a/tools/roll_release.sh
+++ b/tools/roll_release.sh
@@ -102,9 +102,20 @@ else
 fi
 
 echo ""
-echo "Tagging release."
-git tag --annotate "${VERSION_FULL_TAG}" --message="${VERSION_TITLE}"
-git push origin "${VERSION_FULL_TAG}"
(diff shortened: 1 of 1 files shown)


msimberg (approved):
Good idea, thanks!

tools/roll_release.sh
```diff
@@ -103,8 +103,12 @@ fi
 
 echo ""
 echo "Tagging release."
-git tag --annotate "${VERSION_FULL_TAG}" --message="${VERSION_TITLE}"
-git push origin "${VERSION_FULL_TAG}"
+if [[ "$(git tag -l ${VERSION_FULL_TAG})" == "" ]]; then
```
msimberg, line 106: Actually, does this check if the tag exists locally or on the remote? I assume locally only. Could the push to `origin` be done even if the tag has already been created (and maybe also pushed)?

msimberg: Thanks! Yeah, wouldn't have been a big deal as it was, but the update looks even better :)

msimberg (approved):

PR #807, opened 2023-10-12
Fix handling of explicitly specified process masks

msimberg:
Process masks come in two flavours: using physical/OS indices and logical indices. The latter is what hwloc prefers to use for most operations, while the former is what the operating system reports (and what `taskset`, `hwloc-bind --get --taskset`, and slurm masks use). In pika we deal with logical masks. When reading the mask from the process, we already convert the mask from physical to logical indices. When I did https://github.com/pika-org/pika/pull/739 I didn't take this into account and simply used the mask passed to `--pika:process-mask` directly as if it contained logical indices. This adds the conversion from physical to logical indices before the explicit mask is stored.

The physical/OS/logical terminology is what's used in the hwloc manual. It contains more information about how they're handled.

This also adds a simple test where we check that the output of `--pika:print-bind` is what we expect with the fixed indices. The test fails before these changes. The test is somewhat system specific in that it expects two hardware threads per core. We may also have to disable it e.g. for macOS.

### DIFF
--- a/libs/pika/runtime/tests/unit/CMakeLists.txt
+++ b/libs/pika/runtime/tests/unit/CMakeLists.txt
@@ -4,8 +4,9 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-set(tests thread_mapper)
+set(tests process_mask_flag thread_mapper)
 
+set(process_mask_flag_PARAMETERS THREADS 2 RUN_SERIAL)
 set(thread_mapper_PARAMETERS THREADS 4)
 
 foreach(test ${tests})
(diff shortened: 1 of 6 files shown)


libs/pika/topology/src/topology.cpp
```diff
@@ -1169,7 +1169,28 @@ namespace pika::threads::detail {
                 fmt::streamed(mask));
         }
 
-        main_thread_affinity_mask_ = std::move(mask);
+        // The mask is assumed to use physical/OS indices (as returned by e.g. hwloc-bind --get
+        // --taskset or taskset --pid) while pika deals with logical indices from this point
+        // onwards. We convert the mask from physical indices to logical indices before storing it.
+        mask_type logical_mask = mask_type();
```
RMeli, line 1175:
```suggestion
        auto logical_mask = mask_type();
```
?

RMeli (approved):
LGTM (with limited knowledge of `hwloc`'s C API and internals).

libs/pika/topology/src/topology.cpp
msimberg: Updated to simply:
```suggestion
        mask_type logical_mask();
```

msimberg: Well, it turned into
```suggestion
        mask_type logical_mask{};
```
in the end because of the most vexing parse. But updated in any case.

libs/pika/runtime/tests/unit/CMakeLists.txt
```diff
@@ -22,5 +23,17 @@ foreach(test ${tests})
...
   pika_add_unit_test("modules.runtime" ${test} ${${test}_PARAMETERS})
-
 endforeach()
+
+string(
+  CONCAT
+    process_mask_flag_expected_output
+    "   0: PU L#0\\(P#0\\), Core L#0\\(P#0\\), Socket L#0\\(P#0\\), on pool \"default\"\n"
+    "   1: PU L#2\\(P#1\\), Core L#1\\(P#1\\), Socket L#0\\(P#0\\), on pool \"default\"\n"
+    "All tests passed."
```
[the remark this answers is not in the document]
msimberg: Ah, well spotted. I think in that case it's system dependent. The output I put in the regex is what I get locally (and what we get e.g. on CircleCI). Will update the regex to account for that.

Issue #802, opened 2023-10-06
Run all tests under valgrind


msimberg:
#795 and #797 added a few tests to be run under valgrind in CI. The changes in #795 makes it possible to opt-in tests to be tested under valgrind. We should expand the list of tests to improve our coverage. This is another issue probably best tackled roughly module by module (or even test by test). Below is a list of modules (and the examples directory) that are fully covered by valgrind:

- [x] examples (#971)
- [x] affinity (nothing to do)
- [x] allocator_support (nothing to do)
- [x] assertion (#841)
- [x] async_base (nothing to do)
- [x] async_combinators (module removed in #774)
- [ ] async_cuda
- [x] async (module removed in #774)
- [x] async_mpi (#951)
- [x] command_line_handling (#947)
- [x] concepts (nothing to do)
- [x] concurrency (#809)
- [x] config (nothing to do)
- [x] coroutines (#809)
- [x] datastructures (#809)
- [x] debugging (#849)
- [x] errors (#809)
- [x] execution (#795, #809, #839)
- [x] execution_base (#809)
- [x] executors (#797)
- [x] filesystem (nothing to do)
- [x] functional (#851)
- [x] hashing (nothing to do)
- [x] include (nothing to do)
- [x] ini (nothing to do)
- [x] init_runtime
- [x] iterator_support
- [x] itt_notify (nothing to do)
- [x] lcos
- [x] lock_registration
- [x] logging
- [x] memory (#847)
- [x] mpi_base (#951)
- [x] pack_traversal (module doesn't exist)
- [x] prefix (nothing to do)
- [x] preprocessor (nothing to do)
- [x] program_options (#951)
- [x] properties (#849)
- [x] resource_partitioner (#950)
- [x] runtime_configuration (nothint to do)
- [x] runtime (#949)
- [x] schedulers (#946)
- [x] string_util (#847)
- [x] synchronization (#945)
- [x] tag_invoke ( #852)
- [x] thread_pool_util (module doesn't exist)
- [x] thread_pools (nothing to do)
- [x] thread_support (nothing to do)
- [x] threading (#840)
- [x] threading_base (#840)
- [x] thread_manager (nothing to do)
- [x] timing (nothing to do)
- [x] topology (#840)
- [x] type_support
- [x] util (nothing to do)
- [x] version (nothing to do)

msimberg:
#1042 enabled valgrind testing for all modules. The only ones not tested are the MPI and CUDA/HIP modules.

PR #809, opened 2023-10-13
Enable valgrind testing for more modules

msimberg:
This is part of #802 and enables valgrind testing for the following modules:
- `hello_world` in examples
- concurrency
- coroutines
- datastructures
- errors
- execution (not legacy executor tests)
- execution_base

There was one invalid read that is reported across all tests using the pika runtime, which I have suppressed until I have new information about it. From what I can tell it's a false positive. It's an invalid read on this line: https://github.com/pika-org/pika/blob/aca574707cea9d5419add285bb516c69054078f8/libs/pika/coroutines/include/pika/coroutines/detail/posix_utility.hpp#L170. However, I can trigger invalid reads at various places of the mmapped stack if I simply read through all bytes in the stack and as far as I can tell they're all within the bounds of the stack. Time will tell if this is real or not.

Other than the one suppressed error I've made the `contiguous_index_queue` test run a bit faster under valgrind by reducing the problem size. I've also added a number of valgrind flags to the CI configuration. In particular, `--fair-sched=yes` makes the tests run noticeably faster under valgrind.


msimberg:
`split_continuation_clear` still seems to be causing problems, at least in CI, so I've disabled it again from (valgrind) CI testing. Curiously it doesn't seem to trigger any problems locally for me, but maybe I've just been lucky so far. I will merge what works so far, and continue with `split_continue_clear` and the rest of the tests in follow-up PRs.

PR #810, opened 2023-10-13
Set maximum number of threads for macOS CI tests dynamically

msimberg:
The macOS test runner sometimes gets allocated three threads and sometimes four. Our tests expect to know the right number, so this sets the maximum number dynamically.


msimberg:
The problem does not seem to be the number of threads available in CI. Back to testing.

msimberg:
For example, this job got three threads: https://github.com/pika-org/pika/actions/runs/6517546601/job/17702487999#step:5:23 wheras this one got four: https://github.com/pika-org/pika/actions/runs/6517546601/job/17702552567#step:5:23.

PR #821, opened 2023-10-23
Use the `NUM_PROCS` variable provided by the custom cscs build gitlab-runner

### LINKED ISSUE #816: Use reasonable number of cores in CSCS CI build jobs
CI build jobs don't necessarily have all cores available for use in which case `nproc` will return too many cores. The CI templates have some helpers for this, but I'm not 100% sure if/how we can use them. Look into using the number of cores defined by the templates and update our pipelines accordingly. For example: https://gitlab.com/cscs-ci/recipes/-/blob/50f5cef27812f15326d31cdbf9010281ad44c644/templates/v2/.ci-ext-devel.yml#L16.

aurianer:
Fix #816

CSCS's `.container-builder` is actually setting a default of build arguments for the dockerfile provided. I simply added `NUM_PROCS` as an argument in our dockerfile to be able to use it. I also set a default argument in case we want to build the dockerfile locally.

### DIFF
--- a/.gitlab/docker/Dockerfile.spack_build
+++ b/.gitlab/docker/Dockerfile.spack_build
@@ -6,6 +6,8 @@ ARG BUILD_DIR
 
 ARG CMAKE_COMMON_FLAGS
 ARG CMAKE_FLAGS
+# Provided by the gitlab runner of .container-builder
+ARG NUM_PROCS
 
 COPY . ${SOURCE_DIR}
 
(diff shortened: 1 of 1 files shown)


.gitlab/docker/Dockerfile.spack_build
```diff
@@ -6,6 +6,8 @@ ARG BUILD_DIR
 
 ARG CMAKE_COMMON_FLAGS
 ARG CMAKE_FLAGS
+# Overridden automatically when using .container-builder
+ARG NUM_PROCS=16
```
msimberg, line 10: Should we leave this as
```suggestion
ARG NUM_PROCS
```
so that we notice if it's for some reason not defined anymore? I think I'd rather have this be an error than silently fall back to 16 if the internals of `.container-builder` change for some reason.

PR #826, opened 2023-10-25
Cleanup runtime module - part 1

aurianer:
Fixes partially https://github.com/pika-org/pika/issues/16. Part 2 in #1091.


msimberg:
@aurianer what's the status of this? Is there anything missing? Would you need help with anything?

libs/pika/command_line_handling/src/late_command_line_handling.cpp
```diff
@@ -33,7 +33,7 @@ namespace pika::detail {
         return str;
     }
 
-    int handle_late_commandline_options(util::runtime_configuration& ini,
+    int handle_late_commandline_options(pika::util::runtime_configuration& ini,
```
msimberg, line 36: Just curious, did you add this here and elsewhere because things weren't compiling without the extra `pika::` or just to be explicit (I'm asking since quite a few of these changes touch files that have nothing to do with the runtime module)? Not asking for a change here now.

libs/pika/runtime/include/pika/runtime/runtime.hpp
```diff
@@ -38,24 +38,23 @@
...
-        extern std::list<startup_function_type> global_pre_startup_functions;
-        extern std::list<startup_function_type> global_startup_functions;
-        extern std::list<shutdown_function_type> global_pre_shutdown_functions;
-        extern std::list<shutdown_function_type> global_shutdown_functions;
-    }    // namespace detail
+namespace pika::detail {
+
+    ///////////////////////////////////////////////////////////////////////
+    // There is no need to protect these global from thread concurrent
+    // access as they are access during early startup only.
```
msimberg, line 45: Old typo, but maybe worth fixing:
```suggestion
    // access as they are accessed during early startup only.
```

msimberg (approved):

PR #827, opened 2023-10-26
Report exceptions caught by `pika_main` immediately

msimberg:
In #774 I noticed the following failure case that is badly reported by the runtime:
- `pika_main` starts some work that depends on later work being scheduled
- `pika_main` throws an exception
- the runtime stores the exception so that it can report it in `pika::init` or `pika::stop`
- `pika_main` never schedules the work that would allow the previously scheduled work to finish
- `pika::init` or `pika::stop` waits for the runtime to be empty
- the runtime hangs waiting for the runtime to be empty _and_ the error is never reported

I'm not attempting to change how we terminate the runtime here, but I'm simply reporting the caught exception as soon as the runtime gets it. I still let the runtime _attempt_ to shut down cleanly, but at least the error message has been printed.

I think this is another case in favour of not having an "entry point" to the runtime, but simply relying on explicit scheduling of tasks and `sync_wait`. This makes the error reporting much more straightforward (`sync_wait` will simply rethrow an exception immediately and, if uncaught in `main`, terminate the application).

### DIFF
--- a/libs/pika/runtime/src/runtime.cpp
+++ b/libs/pika/runtime/src/runtime.cpp
@@ -45,7 +45,8 @@
 # include <common/TracySystem.hpp>
 #endif
 
-#include <fmt/format.h>
+#include <fmt/ostream.h>
+#include <fmt/printf.h>
 
 #include <atomic>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


msimberg:
cscs-ci run

PR #828, opened 2023-10-26
Allow printing bindings by setting `PIKA_PRINT_BIND` environment variable

msimberg:
The main question is how to handle the truthiness of the environment variable. This first implementation simply checks if `PIKA_PRINT_BIND` is set at all, and if it is it will print the bindings. Other alternatives are to consider certain values truthy, and the rest falsy (or vice versa).


RMeli (approved):
> The main question is how to handle the truthiness of the environment variable.

Yeah... Having `export PIKA_PRINT_BIND=0` printing the binding might be quite confusing for the end user, but I'm not sure this is worth a more sophisticated treatment. It is clear in the documentation.

For reference, this is what CMake does for generator expressions:
```
 $<BOOL:string>

    Converts string to 0 or 1. Evaluates to 0 if any of the following is true:

        string is empty,

        string is a case-insensitive equal of 0, FALSE, OFF, N, NO, IGNORE, or NOTFOUND, or

        string ends in the suffix -NOTFOUND (case-sensitive).

    Otherwise evaluates to 1.
```

msimberg:
> Yeah... Having export PIKA_PRINT_BIND=0 printing the binding might be quite confusing for the end user, but I'm not sure this is worth a more sophisticated treatment. It is clear in the documentation.

Indeed. I know it's a somewhat selfish reason but it is simpler to treat any value, also empty, as true. It's also simple to explain. Some other examples to support this approach is ccache's treatment of boolean values (https://ccache.dev/manual/latest.html#_boolean_values; however, they additionally error out on a few falsy values to avoid surprises). Cray's MPICH documentation seems to have _some_ variables that are considered true simply from being set. Other variables seem to allow truthy and falsy values.

Shall we go with the current simple approach and constrain the allowed values later if there seems to be demand for it? I think `PIKA_PRINT_BIND` is at least a non-critical option (in the sense that it doesn't affect the behaviour of the runtime, it's purely a "debuggin" option) where we can hopefully afford to make a breaking change later on.

RMeli:
> Shall we go with the current simple approach and constrain the allowed values later if there seems to be demand for it?

It's a yes from me (that's why I approved the PR despite the comment).

msimberg:
cscs-ci run

PR #814, opened 2023-10-14
Split the CSCS CI pipeline in different files for more clarity

aurianer:

PR #830, opened 2023-10-27
Trigger error in CI pipeline if no tests run + small changes of ctest options

### LINKED ISSUE #829: Find a way to make the status check fail in CI when no tests are run

aurianer:
Based on #814 
Fixes #829

- `ctest` now returns an error in case no tests are run.
- Removes `--verbose` option from `ctest` as `--output-on-failure` is sufficient.
- Make `-L` more exhaustive.


msimberg (approved):

msimberg:
cscs-ci run

msimberg:
This seems to be working as intended: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/5413650946#L1664! Perhaps unintuitively `tests.unit.build` isn't actually a test target, but only a build target, so you can remove the separate `ctest` step for `tests.unit.build` (you can see they get built in the earlier steps).

PR #831, opened 2023-10-27
Revert #774

msimberg:
This reverts #774 for compatibility-related reasons for the 0.20.0 release. I will un-revert the changes again for 0.21.0.

The CP2K integration of DLA-Future (and thus pika) benefits greatly from #807, which will be included in 0.20.0. The latest release of DLA-Future is 0.2.1 which still requires pika-algorithms, and including #774 in 0.20.0 breaks pika-algorithms. By keeping #774 in 0.20.0 there would be no released version of pika that can be used with CP2K and DLA-Future 0.2.1 (0.20.0 breaks pika-algorithms, and 0.19.0 doesn't have #807). By postponing #774 to 0.21.0 the initial CP2K integration can use DLA-Future 0.2.1 and pika 0.20.0.


msimberg:
cscs-ci run

PR #835, opened 2023-10-31
Add more gcc configurations in CSCS CI

aurianer:
Fix partially #32.

### DIFF
--- a/.gitlab/includes/gcc9_pipeline.yml
+++ b/.gitlab/includes/gcc9_pipeline.yml
@@ -1,0 +1,73 @@
+# Copyright (c) 2023 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file BOOST_LICENSE_1_0.rst or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+include:
+  - local: '.gitlab/includes/common_pipeline.yml'
+  - local: '.gitlab/includes/common_spack_pipeline.yml'
+
+.variables_gcc9_config:
+  variables:
+    ARCH: linux-ubuntu22.04-broadwell
+    BUILD_TYPE: Debug
+    COMPILER: gcc@9.3.0
+    CXXSTD: 17
+    SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} malloc=system cxxstd=$CXXSTD \
+                 +generic_coroutines ^boost@1.71.0 ^hwloc@1.11.5"
+
+gcc9_spack_image:
+  stage: spack_configs
+  needs: [base_spack_image]
+  extends:
+    - .container-builder
+    - .variables_gcc9_config
+  before_script:
+    - CONFIG_TAG=`echo $SPACK_SPEC-$SPACK_COMMIT | sha256sum - | head -c 16`
+    - compiler=${COMPILER/@/-}
+    - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/pika-$compiler:$CONFIG_TAG
+    - echo -e "compiler=$compiler\nBASE_IMAGE=$PERSIST_IMAGE_NAME" >> compiler.env
+  variables:
+    BASE_IMAGE: $CSCS_REGISTRY_PATH/pika-spack-base:$SPACK_COMMIT
+    DOCKERFILE: .gitlab/docker/Dockerfile.spack_compiler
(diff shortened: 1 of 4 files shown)


msimberg:
You'll have to leave out the cce configuration. As long as we don't have a Cray base image we won't get be able to test the Cray compilers in the current setup. We can consider asking for access to the bare metal runners for that configuration, but I would definitely leave that out of this PR.

Running the MPI tests will require a slight restructuring, but shouldn't be a big issue if we label the MPI tests as well. I also suggest leaving this to a separate PR.

msimberg:
cscs-ci run

.gitlab/includes/gcc10_apex_pipeline.yml
```diff
@@ -0,0 +1,73 @@
...
+  - local: '.gitlab/includes/common_spack_pipeline.yml'
+
+.variables_gcc10_apex_config:
+  variables:
+    ARCH: linux-ubuntu22.04-broadwell
+    BUILD_TYPE: Debug
+    COMPILER: gcc@10.3.0
+    CXXSTD: 17
+    SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} +apex malloc=system cxxstd=$CXXSTD ^apex@2.4.1 \
+                 ~openmpi ~papi"
```
msimberg, line 18:
```suggestion
                 ~mpi ~papi"
```
?

.gitlab/includes/gcc9_pipeline.yml
```diff
@@ -0,0 +1,73 @@
...
+  - local: '.gitlab/includes/common_pipeline.yml'
+  - local: '.gitlab/includes/common_spack_pipeline.yml'
+
+.variables_gcc9_config:
+  variables:
+    ARCH: linux-ubuntu22.04-broadwell
+    BUILD_TYPE: Debug
+    COMPILER: gcc@9.3.0
+    CXXSTD: 17
+    SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} malloc=system cxxstd=$CXXSTD ^boost@1.71.0 \
```
msimberg, line 17:
```suggestion
    SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} malloc=system cxxstd=$CXXSTD +generic_coroutines ^boost@1.71.0 \
```

.gitlab/includes/gcc12_cuda12_pipeline.yml
```diff
@@ -0,0 +1,76 @@
...
+    - .variables_gcc12_cuda12_config
+    - .test_common_gpu
+  needs:
+    - gcc12_cuda12_debug_build
+  script:
+    - spack arch
+    - spack build-env $spack_spec -- bash -c "ctest --label-exclude COMPILE_ONLY --test-dir ${BUILD_DIR} -j$(nproc) --timeout 120 --output-on-failure --no-compress-output --no-tests=error"
+      # All async_cuda tests are disabled from running because the driver on Piz Daint is too old
+      # for CUDA 12
+  image: $PERSIST_IMAGE_NAME
```
msimberg, line 76: This test stage will not work on daint because CUDA 12 requires a newer driver which daint doesn't have (specifically, the CUDA tests won't run). We can leave it out for now, and consider running on clariden/hohgant later. The jenkins CI configuration with CUDA 12 also doesn't _run_ the CUDA tests.

.gitlab/includes/gcc9_cuda11_pipeline.yml
```diff
@@ -0,0 +1,73 @@
...
+  - local: '.gitlab/includes/common_spack_pipeline.yml'
+
+.variables_gcc9_cuda11_config:
+  variables:
+    ARCH: linux-ubuntu22.04-broadwell
+    BUILD_TYPE: Debug
+    COMPILER: gcc@9.3.0
+    CXXSTD: 17
+    SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} +cuda malloc=system cxxstd=$CXXSTD \
+                 ^boost@1.75.0 ^hwloc@2.0.3 ^cuda@11.2.0 ^fmt@9"
```
msimberg, line 18: This should not be necessary now that you've updated the spack commit:
```suggestion
                 ^boost@1.75.0 ^hwloc@2.0.3 ^cuda@11.2.0"
```

msimberg: All right sounds good. In that case there may be new/other problems with fmt 10 and CUDA, and we may have to extend the conflicts in the spack package.

msimberg (approved):

PR #838, opened 2023-11-01
Remove unnecessary ctest command line arguments from GCC 11 pipeline config

msimberg:
`-R tests` is unecessary because all tests contain `tests` already, and `--exclude-regex tests.unit.build` is unnecessary because the build unit tests are only build targets, not tests (https://github.com/pika-org/pika/pull/830#issuecomment-1784777396).

@aurianer this would be useful to include in the other configurations in #835.


msimberg:
cscs-ci run

PR #839, opened 2023-11-02
Fix more lifetime bugs in `split`

msimberg:
Enables valgrind testing for all of the `execution` module. Description of the bug in https://github.com/pika-org/pika/commit/3291a045f9644591e97e2fe9d0c50f6f59fab5a6. The bug is very timing dependent and I can't think of a way to reproduce it with a unit test. However, for the record, here is a reproducer that triggers the problem easily: https://github.com/msimberg/pika/tree/split-continuation-clear-valgrind-reproducer. It relies on adding sleeps inside the implementation.

With these changes I've also cleaned up both `split` and `ensure_started`'s handling of clearing continuations, since in both cases the receiver holds a reference to the shared state now.


msimberg:
cscs-ci run

PR #842, opened 2023-11-03
Set maximum number of threads for GitHub actions Linux tests dynamically

msimberg:
Like the macOS runners, the Linux runners apparently also now can have different hardware configurations, and may have more than two threads as we've so far assumed. C.f. https://github.com/pika-org/pika/pull/810.


msimberg:
cscs-ci run

PR #843, opened 2023-11-03
Set maximum number of threads for GitHub actions Linux tests dynamically, take two

msimberg:
C.f. #842. Use processing units, not cores, as the maximum number of threads in tests because they are what `hardware_concurrency` report, and the actual maximum pika supports ignoring real oversubscription.


msimberg:
I will rerun the affected workflows a few times before merging this to make sure it works properly this time.

msimberg:
cscs-ci run

msimberg:
This seems to work better now... let's give it a go.

PR #844, opened 2023-11-03
Reuse `any_sender` storage when moving to a `unique_any_sender`

### LINKED ISSUE #556: Optimize `any_sender` to `unique_any_sender` conversion
Currently the conversion works because a `unique_any_sender` can be constructed from any r-value connectable sender, which `any_sender` is. However, the conversion could be optimized to simply move the internal buffer from the `any_sender` to the `unique_any_sender` and potentially avoid an allocation.

msimberg:
Fixes #556.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #848, opened 2023-11-03
Remove extra _test suffix to avoid having _test_test

aurianer:
Just noticed the functional tests end up with executable named `<name>_test_test`


msimberg (approved):
Good catch, thanks!

PR #850, opened 2023-11-04
Remove leftover range unit test, already in iterator_support

aurianer:
The `range` unit test in `thread_support` was not tested. It seems to be a copy of the one in `iterator_support` modulo small updates like `PIKA_TEST_EQ` => `PIKA_TEST`


msimberg (approved):
Interesting... it looks like I mistakenly created this in the big hpx-local revert PR. Thanks for the cleanup!

PR #854, opened 2023-11-06
Run valgrind tests with both release and debug mode

msimberg:


msimberg:
The `contiguous_index_queue` test failed the first time I tested this presumably because the threads don't get scheduled fairly enough. We'll have to see if we disable the test completely or try to adjust the testing criteria for it.

msimberg:
> The `contiguous_index_queue` test failed the first time I tested this presumably because the threads don't get scheduled fairly enough. We'll have to see if we disable the test completely or try to adjust the testing criteria for it.

I've disabled that particular failing check with valgrind enabled in release mode in https://github.com/pika-org/pika/pull/854/commits/42affad7f99a3b6551cea5e5da85e825c85f2d12. Since it's a sanity check in the sense that if that fails the test may not be testing concurrent use of the queue, but we still keep it enabled in other configurations, I think it's ok to disabled it for that one configuration.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
Note that the required checks change with this PR, as the valgrind workflow generates two separate statuses. Other PRs will need rebasing once this is merged.

PR #856, opened 2023-11-08
Make sure deadlock detection is triggered when `yield_while` isn't allowed to suspend

### LINKED ISSUE #688: Fix spinlock deadlock detection when suspension is disabled in `yield_while`
Currently we manually `% 16` the iteration counter `k` to signal that threads should not suspend when calling `yield_k`. This does disable suspension but it also circumvents the check for spinlock deadlocks (i.e. the check that throws if `k` becomes very large) since it resets `k` after it reaches `16`. We should refactor `yield_while`/`yield_k` so that the deadlock detection still triggers. We should also ideally hide the magic number `16` somewhere deeper so that callers of `yield_k` don't have to know about it.

msimberg:
Fixes #688.

- Adds a `spin_k` member to `agent_base`, `agent_ref`. This only ever spins, never yields, unlike `yield_k`.
- Updates `yield_while` and variants to use `spin_k` when suspension is not allowed.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #857, opened 2023-11-13
Add clang configurations for CSCS CI

aurianer:
Fixes partially #32.


msimberg:
> Just to have some references here, it seems like I can't install clang@15.0.0 with libffi@3.4.4, I'll try another version for libffi and see if we use the same in the jenkins pipeline https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/5529513221

Looks like it's fixed by this possibly: https://github.com/libffi/libffi/pull/764/files? However, there's no release with those changes yet. Perhaps worth a patch to spack?

msimberg (approved):
Thanks!

We definitely need to try to reduce the duplication between the configs, but does not need to be done here. Opened https://github.com/pika-org/pika/issues/862 to keep track of it.

PR #860, opened 2023-11-17
Remove `CSCS_REBUILD_POLICY` always

aurianer:
I committed this when I added the rm on the tarball (with the spack commit unchanged), but this doesn't need to be here, we only need this step to run when the `SPACK_COMMIT` is changing now


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #861, opened 2023-11-19
Add Clang 14 CUDA 11 pipeline with CSCS CI

aurianer:
Part of #32

### DIFF
--- a/.gitlab/includes/clang14_cuda11_pipeline.yml
+++ b/.gitlab/includes/clang14_cuda11_pipeline.yml
@@ -15,8 +15,7 @@ include:
     COMPILER: clang@14.0.6
     CXXSTD: 17
     SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} +cuda malloc=system cxxstd=$CXXSTD \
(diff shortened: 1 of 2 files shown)


msimberg:
Is the failure you're trying to work around https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/5570278524#L533? If yes, we're getting that because spack does not model using clang as a CUDA compiler, and the `conflicts` with C++20 is wrong for that case. If possible, I would maybe suggest that we right now use `cxxstd=17` in the spec, but build pika with `CMAKE_CXX_STANDARD=20`.

msimberg:
Hmm, still failing. I think it's because in Jenkins we are still using a spack commit that doesn't have the conflict. I'm not sure yet what a good solution here would be. A couple of options:

- Keep an internal `package.py` with exceptions/modifications (I don't love this...)
- Try to model the clang-as-cuda-compiler option in spack (with a variant? I've asked on the spack slack if they have any better suggestions)
- Something else?

Any ideas?

msimberg (approved):
Thanks! Could you open an issue for tracking us adding back a clang/CUDA stdexec configuration?

msimberg:
cscs-ci run

PR #864, opened 2023-11-20
Don't create more streams than required in `cuda_pool`

msimberg:
Create number of streams based on `get_num_worker_threads` if the runtime is up, otherwise fall back to `hardware_concurrency`. This is intended to work around issues with HIP event creation being significantly slower if many streams are created. E.g. on an AMD EPYC system with 128 hardware threads (`hardware_concurrency == 128`) the `cuda_pool` would create `(3 + 3) * hardware_concurrency = 768` streams per rank, even if the runtime is only started with e.g. 8 threads (8 ranks per node, without hyperthreading). If the actual number of runtime threads is used, the number of threads created is instead 48. This can significantly improve performance if many kernels are submitted.

The `cuda_pool` still falls back to using `hardware_concurrency` if it's created before the pika runtime is started.


msimberg:
cscs-ci run

PR #865, opened 2023-11-21
Print spack spec in CSCS CI pika building step

aurianer:
In case the compiler image is already up-to-date on jfrog, this image is used and the spec is not printed, I find it handy to have it printed in the pika build step too


msimberg (approved):
Good idea!

msimberg:
cscs-ci run

PR #869, opened 2023-11-24
Add `require_started` sender adaptor

### LINKED ISSUE #790: Add debug sender adaptor to diagnose unstarted senders
In some use cases it's fine to leave senders unstarted, but in others it's a bug. If it's a bug it may be hard to tell which senders were unstarted. Currently `split` is the only adaptor that diagnoses if it's unstarted because it's required to be connected and started at least once, but other senders don't do it.

We could add a simple debug adaptor (`debug_unstarted`?) which simply checks in the destructor of the sender if the sender has been started at least once. Otherwise the adaptor would simply forward any and all values that are sent from the previous sender. This would likely look similar to #772.

msimberg:
Fixes #790.

Comments on the name are especially welcome. I found `require_started` quite descriptive as it signals that the user is required to make sure that the sender is started. It also mirrors `ensure_started` which actually connects and starts a sender.

The `require_started` sender has an empty state to make it somewhat sensible to detect unstarted senders. E.g. `just()` is copyable, but it's also possible to connect it after moving it. This is not allowed by the `require_started` sender. The `require_started` sender stores the wrapped sender internally in an `optional`, and when moving a `require_started` sender the `optional` is emptied.

A `require_started` sender can be allowed to not be started, but it has to be explicitly asked for with `discard`.

When a `require_started` sender is not connected, or when it's connected but the operation state not started, the destructor will terminate (by default). Assigning to unstarted senders also leads to termination.

Mainly for testing purposes the behaviour on unstarted senders is customizable. The default behaviour is to terminate, since that reports the error immediately and cannot be ignored. The behaviour can be changed to throw instead with a flag passed to the sender constructor, or changed afterwards with `set_mode`. I don't expect this second mode to be used much in practice, but it's useful to have as an alternative at least for testing. The choice of mode is disabled with stdexec, since stdexec requires nothrow destructible senders. In this case `terminate` is always used.

### DIFF
--- a/libs/pika/execution/CMakeLists.txt
+++ b/libs/pika/execution/CMakeLists.txt
@@ -17,6 +17,7 @@ set(execution_headers
     pika/execution/algorithms/just.hpp
     pika/execution/algorithms/let_error.hpp
     pika/execution/algorithms/let_value.hpp
+    pika/execution/algorithms/require_started.hpp
     pika/execution/algorithms/schedule_from.hpp
     pika/execution/algorithms/split.hpp
     pika/execution/algorithms/split_tuple.hpp
--- a/libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
@@ -20,6 +20,7 @@
 #include <pika/type_support/detail/with_result_of.hpp>
 #include <pika/type_support/pack.hpp>
 
+#include <fmt/core.h>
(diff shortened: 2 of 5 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #866, opened 2023-11-22
Reduce test parallelism in CI to `$(nproc)/2`

aurianer:
As pointed by @msimberg the timeout in the binary semaphore tests might be due to oversubscription, this PR is an attemps to fix that

PR #873, opened 2023-11-30
Add `PIKA_WITH_PARALLEL_TESTS_BIND_NONE` in CSCS CI

aurianer:
It was enabled by default in `ctest.cmake` in jenkins (cf https://github.com/pika-org/pika/blob/main/.jenkins/cscs/ctest.cmake#L69)


msimberg (approved):
Excellent, thanks. This is likely a required addition to https://github.com/pika-org/pika/pull/866 to avoid test timeouts. If we still see issues after this I think we can start suspecting the tests themselves rather than the test setup.

msimberg:
Are you planning further changes to this PR? It looks ready to go.

PR #875, opened 2023-12-01
Remove globbing fortran source files

aurianer:


msimberg (approved):
Thanks for the cleanup, this is really nice! A possible addition to this or a separate PR: could you check if we even use the `AUTOGLOB` option anywhere? I think it used to be used back in HPX, but I'm not sure we need it anymore.

msimberg:
cscs-ci run

PR #876, opened 2023-12-04
Add `pika::is_runtime_initialized` function

### LINKED ISSUE #808: Add function to check if runtime is initialized
We currently don't have a way for users to up-front check if the pika runtime is initialized. We do internally check if the runtime is initialized, e.g. when getting a thread pool, but in some cases it can be useful to do that check as the user and either report a better error or simply avoid using the runtime if it's not initialized.

We have an internal function for getting the runtime pointer, but I don't think we should be exposing that. We have runtime state-specific functions (like `is_running`), but they don't tell a user if the runtime is simply initialized.

Maybe we need simply `bool pika::is_initialized()` which returns true between `pika::start()` and `pika::stop()`, and otherwise false? Alternatively, if we need the information on the status of the runtime, we could return an optional of a status. The optional being empty would indicate that the runtime has not been initialized.

msimberg:
Fixes #808.

However, not 100% sure about the name. I would like to include `runtime` in the function name, so it's not simply taking the generic name `pika::is_initialized`. I don't think we want to use that particular name for anything else, but we might want to check if other things (thread pools, cuda polling, etc.) are "initialized". Adding the extra `runtime` qualifier seems better to me, but open to other suggestions.

As for the actual word to use here: this function checks if the runtime exists, meaning it will return `false` before `pika::start/init` and after `pika::stop/init`. It'll return `true` between `pika::start` and `pika::stop` and from other threads while `pika::init` hasn't returned.

`initialized` can be slightly misleading. E.g. `MPI_Initialized` returns true forever after `MPI_Init` has been called, even after `MPI_Finalize`, but then `MPI_Init` cannot be called multiple times, unlike `pika::start/init`.

`active` can be slightly misleading given that the runtime can be suspended. I think that `inactive` could be confused with runtime exists but is suspended.

`exists` is another option that I actually think is quite accurate, but I haven't seen it used for this kind of a check before so may simply be unfamiliar.


msimberg:
cscs-ci run

RMeli (approved):
I think `pika::is_runtime_initialized` is good enough (especially so in light of your explanation/analysis). I can't think of anything better, and for what it's worth I don't like `exists`.

albestro:
Thanks for the overview describing the full mental process that lead to current naming. I agree with @RMeli, and it sounds perfectly good!

Justifying a bit my comment, not to change your mind, but to give you additional food for thoughts: I was thinking that `running`, ie `is_runtime_running`, could be added to the list of considered terms. I have some memories of reading comments about the going on cleanup/refactoring of `init`/`start`. It is not clear to me where pika is headed about it (my fault, I'm not following closely).

IMHO `is_runtime_running` loose sense when the sequence of actions on runtime can have something different than `start`/`stop`. And, in the current status, the sequence could have also `init`/`finalize`/`suspend`/`resume` (which I'm not sure if they are interchangeable/synonymous or not).

msimberg:
Thanks all for the comments! I think we'll go with `is_runtime_initialized`.

@albestro to respond explicitly to your comment about `is_runtime_running`:

We in fact already have a `pika::is_running` already! The reason I didn't want to go in that direction is the potential mixup with the _state_ of the runtime, when it's "initialized" (according to the definition of this PR). I would not consider an "initialized" but "suspended" runtime to be "running". Currently "running" is a synonym for "active" or "worker threads are processing tasks". A "suspended" runtime would still be "initialized" but not "running".

Regarding cleanup of `init`/`start` etc., I think that would not change any of the discussion here. The potential cleanup that I have in mind is to make use of `start`/`stop` explicitly required, and remove the combined `init`.

msimberg:
I've added some documentation. I also realized that I was in fact using the terms "active" and "inactive" previously for what I now call "initialized" and "not initialized". I've updated the documentation for other functions as well. Please have another look to see if you actually think it's an improvement or not.

msimberg:
cscs-ci run

PR #880, opened 2023-12-05
Fix minor cmake var use when undefined problem

biddisco:
When reusing some pika cmake macros in another project, I noticed a couple of places where variables are dereferenced when possibly unused, these should be used 'bare' to avoid errors

### DIFF
--- a/cmake/pika_add_test.cmake
+++ b/cmake/pika_add_test.cmake
@@ -55,13 +55,13 @@ function(pika_add_test category name)
   endif()
 
   set(args "--pika:threads=${${name}_THREADS}")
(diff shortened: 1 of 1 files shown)


msimberg (approved):
Thanks. The change is good (hence the approval), but just for my understanding: in what situations could these be undefined (because we may also need to fix the cases where they are undefined)?

msimberg:
cscs-ci run

biddisco:
> Thanks. The change is good (hence the approval), but just for my understanding: in what situations could these be undefined (because we may also need to fix the cases where they are undefined)?

I reuse some macros in my project without actually setting up all the pika variables, so it is probably because I'm bypassing some setup that the vars are not initialized.

PR #891, opened 2023-12-11
Push performance benchmark data to elastic database

msimberg:
Pushes performance benchmark data to elastic (via logstash). The json blob that is sent is structured as follows:
```
{
  "build_configuration": {
    "ARCH": "linux-ubuntu22.04-broadwell",
    "BUILD_TYPE": "Release",
    "CMAKE_COMMON_FLAGS": "-GNinja -DPIKA_WITH_COMPILER_WARNINGS=ON -DPIKA_WITH_COMPILER_WARNINGS_AS_ERRORS=ON -DPIKA_WITH_CHECK_MODULE_DEPENDENCIES=ON -DPIKA_WITH_EXAMPLES=ON -DPIKA_WITH_TESTS=ON -DPIKA_WITH_PARALLEL_TESTS_BIND_NONE=ON",
    "CMAKE_FLAGS": "",
    "COMPILER": "gcc@13.1.0",
    "SPACK_COMMIT": "27c0dab5ca1944d686894f17e2a05bc7ddac3061",
    "SPACK_SPEC": "pika@main arch=linux-ubuntu22.04-broadwell %gcc@13.1.0 malloc=mimalloc cxxstd=20 ^boost@1.83.0 ^hwloc@2.9.1"
  },
  "ci": {
    "CI_COMMIT_AUTHOR": "Mikael Simberg <mikael.simberg@iki.fi>",
    "CI_COMMIT_BRANCH": "__CSCSCI__pr891",
    "CI_COMMIT_DESCRIPTION": "",
    "CI_COMMIT_MESSAGE": "Add CSCS CI job to submit benchmark data to elasticsearch\n",
    "CI_COMMIT_SHA": "12399ef6cd1d27043fbbbecc6f3829ad420dde27",
    "CI_COMMIT_SHORT_SHA": "12399ef6",
    "CI_COMMIT_TIMESTAMP": "2023-12-18T11:51:57+01:00",
    "CI_COMMIT_TITLE": "Add CSCS CI job to submit benchmark data to elasticsearch",
    "CI_JOB_IMAGE": "...",
    "organization": "pika-org",
    "repository": "pika"
  },
  "data_stream": {
    "dataset": "service.pika",
    "namespace": "alps",
    "type": "logs"
  },
  "metric": {
    "benchmark": {
      "command": "task_size_test",
      "command_options": "--tasks-per-thread=10000 --task-size-growth-factor=1.01 --target-efficiency=0.95 --perftest-json",
[...]

PR #893, opened 2023-12-11
Add `task_size` performance test

msimberg:
From the comment in the test:

> This test finds the smallest grain/task size that yields a parallel efficiency at least as large as the target efficiency. The timing is always compared to an ideal single-threaded execution, i.e. the number of tasks multiplied by the grain size. The search starts from the minimum grain size. The grain size is then grown geometrically until the efficiency of the parallel execution of the given number of tasks is above the target efficiency.
>
> This test is useful for finding an appropriate minimum task size for a system, i.e. one that gives resonable parallel efficiency.

I'm adding this for potential use in #891. The grain size is a useful compressed way of expressing how well pika performs on a system (or how well a system performs with pika). The smaller the grain size with a given efficiency, the better it performs. Obviously for more detailed analysis a full sweep of task sizes, number of threads, etc. is still useful, but this gives a quick indication of what a good task size is for a given system.

I'm using the simpler `schedule`/`when_all`/`sync_wait` pattern for launching tasks as it gives a pessimistic estimate. Hierarchical scheduling might give a smaller grain size, but it's not the typical way of launching tasks.

### DIFF
--- a/tests/performance/local/CMakeLists.txt
+++ b/tests/performance/local/CMakeLists.txt
@@ -19,6 +19,7 @@ set(benchmarks
     skynet
     task_overhead
     task_overhead_report
+    task_size
 )
 
 if(NOT PIKA_WITH_SANITIZERS)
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run

msimberg:
Renamed all `grain_size` occurrences to the easier-to-understand `task_size`.

msimberg:
cscs-ci run

PR #900, opened 2023-12-14
Task overhead cleanup

msimberg:
Separates some test cleanup that was originally part of #891.

- Remove the `--test-all` command line option from `task_overhead_report_test` since there is only one test, and it wasn't previously run by default (it is now always run)
- Remove the test-specific `executor` field from the performance test json output to allow generalization to other types of tests


msimberg:
cscs-ci run

PR #902, opened 2023-12-14
Add command line option to output json for performance CI to `task_size` test

msimberg:
Only a single number is printed for the timing "series": the task size which results in at least the target efficiency. Since the test is run multiple times with increasing task sizes, this number should be relatively stable (at least if the growth factor of the task size is small).


msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #912, opened 2024-01-03
Update matplotlib dep to avoid critical CVE

aurianer:


msimberg (approved):
Thanks! If you didn't already, don't forget to update the environment on daint as well.

PR #913, opened 2024-01-03
Reenable build unit tests on clang 14/CUDA 11 CI configuration

msimberg:
With Jenkins this configuration was using Cray's clang. In the new CSCS CI setup this is no longer the case, and the test seems to be working with vanilla clang.


msimberg:
cscs-ci run

PR #915, opened 2024-01-04
Don't explicitly set permissions on pika-bind when installing

### LINKED ISSUE #911: Question about the permission of pika-bind script
https://github.com/pika-org/pika/blob/dc24ede8eb987a2c877874a13aa6ca809107cd31/CMakeLists.txt#L1759-L1764

With setting of `PERMISSIONS OWNER_READ GROUP_READ WORLD_READ OWNER_EXECUTE GROUP_EXECUTE WORLD_EXECUTE` to the file,  and I do the RPM packaging of the project. The results show that the file is write-protected regular file, when I want to remove the file from the RPM package.

msimberg:
The `PROGRAMS` mode of install will automatically set the appropriate bits for the scripts: https://cmake.org/cmake/help/latest/command/install.html#programs. Specifically, it seems to be common to include the write bit for the owner, which was not in the explicit list of permissions, but is set by default.

Fixes #911.


msimberg:
@RMeli do you think this looks reasonable as well? See #911 for context.

msimberg:
cscs-ci run

RMeli (approved):
@msimberg looks reasonable, thanks! IIRC, I iterated a bit on this and initially used `FILES` for which execute permissions are not set. I must have forgot to remove the explicit permission when changing it to `PROGRAMS`.

PR #922, opened 2024-01-05
Relax regex used to check `process_mask_flag` test output

msimberg:
The important aspect of the output is that the first two threads use physical PU numbers 0 and 1, the other indices can vary by machine.


msimberg:
cscs-ci run

PR #924, opened 2024-01-05
Use moodycamel queue also for terminated tasks in `shared_priority_queue_scheduler`

msimberg:
This improves the performance of the scheduler on systems where lockfree 128-bit atomics are not available or can't be generated. GCC 12 and older on e.g. Nvidia Grace CPUs seems to still produce lockful 128-bit atomics.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #927, opened 2024-01-05
Improve `spin_k` and `yield_k` performance by pausing more

msimberg:
- In `yield_k`, pause every iteration before actually yielding
- In `spin_k`, add a simple backoff by spinning inside `spin_k` `k` times before returning

Especially the latter improves performance of contended barriers on Grace.


msimberg:
cscs-ci run

PR #940, opened 2024-01-09
Replace barrier implementation with libc++'s tree barrier

msimberg:
Significantly improves performance of the barrier by using libc++'s tree barrier implementation (https://github.com/llvm/llvm-project/blob/06286a553280fc843d6f5df477a2c776aa2ece35/libcxx/include/barrier, https://github.com/llvm/llvm-project/blob/06286a553280fc843d6f5df477a2c776aa2ece35/libcxx/src/barrier.cpp). This implementation requires no locks which already improves scalability. The tree implementation should further reduce memory traffic, especially across NUMA nodes. It also uses polling instead of a condition variable to avoid the use of locks.

### DIFF
--- a/.gitlab/scripts/run_performance_benchmarks.sh
+++ b/.gitlab/scripts/run_performance_benchmarks.sh
@@ -135,7 +135,7 @@ pika_test_options=(
 "--method=barrier
 --tasks-per-thread=1000 \
 --task-size-growth-factor=1.05 \
---target-efficiency=0.5 \
+--target-efficiency=0.9 \
 --perftest-json"
 
 "--method=bulk


libs/pika/synchronization/include/pika/synchronization/barrier.hpp
```diff
@@ -1,18 +1,33 @@
...
 
 #pragma once
 
 #include <pika/config.hpp>
 #include <pika/assert.hpp>
 #include <pika/concurrency/spinlock.hpp>
-#include <pika/synchronization/detail/condition_variable.hpp>
 #include <pika/thread_support/assert_owns_lock.hpp>
 
 #include <chrono>
```
msimberg, line 33: Check for missing/unused includes.

libs/pika/synchronization/include/pika/synchronization/barrier.hpp
```diff
@@ -88,24 +129,21 @@ namespace pika {
...
-        using arrival_token = bool;
+        using arrival_token = detail::barrier_phase_t;
 
-        // Returns:        The maximum expected count that the implementation
-        //                 supports.
-        static constexpr std::ptrdiff_t(max)() noexcept
+        static constexpr std::ptrdiff_t max() noexcept
         {
-            return (std::numeric_limits<std::ptrdiff_t>::max)();
+            return std::numeric_limits<std::ptrdiff_t>::max();
```
msimberg, line 146: Put back parentheses around `max`.

msimberg:
cscs-ci run

msimberg:
@aurianer and @biddisco I would kindly ask you to look quite carefully through this to see if you can spot any potential problems with the pika-specific parts (I trust the upstream implementation to be mostly correct). In particular, the thread id mapping I'm expecting to not cause problems but I don't excluded that it'll come back to haunt us. pika's thread ids are really just addresses, and as long as the `std::thread` ids are that as well I don't expect conflicts but that may of course not be the case. Performance is significantly better in the tests that I ran, almost everywhere. There are a few configurations with DLA-Future on Piz Daint that seem to actually slightly slower but that's a tradeoff I think I'm willing to make. It may be that the tree barrier has slightly higher overhead for a low number of threads.

libs/pika/synchronization/include/pika/synchronization/barrier.hpp
```diff
@@ -88,22 +130,19 @@ namespace pika {
...
+        detail::barrier_algorithm_base base;
 
     public:
-        using arrival_token = bool;
+        using arrival_token = detail::barrier_phase_t;
 
-        // Returns:        The maximum expected count that the implementation
-        //                 supports.
-        static constexpr std::ptrdiff_t(max)() noexcept
+        static constexpr std::ptrdiff_t max() noexcept
```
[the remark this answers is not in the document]
msimberg: Ah, no this should be kept as the old version (even though it's only a potential problem on Windows). The parentheses were exactly to avoid expansion of the Windows macro `max`.

msimberg:
cscs-ci run

PR #941, opened 2024-01-09
Update NVHPC CI to 23.9

aurianer:
Need installation on daint but it worked with my spack installation. I added gcc as a compiler for fmt as it didn't seem to find the corresponding flags for CXX20. I tried with fmt@10.1.1, fmt@10.0.0, fmt@9 and the same error occured.

### DIFF
--- a/.jenkins/cscs/env-nvhpc.sh
+++ b/.jenkins/cscs/env-nvhpc.sh
@@ -8,7 +8,7 @@ cxx_std="20"
 boost_version="1.78.0"
 hwloc_version="2.7.0"
 stdexec_version="git.nvhpc-23.09.rc4=main"
-nvhpc_version="22.11"
+nvhpc_version="23.9"
 spack_compiler="nvhpc@${nvhpc_version}"
 spack_arch="cray-cnl7-haswell"
 


msimberg:
What is the error you get related to C++20?

msimberg:
> > What is the error you get related to C++20?
> 
> [error fmt cxx20 nvhpc](https://gist.github.com/aurianer/2e96cbaafae30f6212c7af2d759dbbea)

> fmt@10.2.1 seems to work with nvhpc@23.9 though

The error message seems to be installing fmt 10.2.1 though? What CMake version did you use when you got the error message? Typically those kind of errors go away with newer versions of CMake, because they explicitly add support for C++ standard versions per compiler instead of detecting the flags.

msimberg:
@aurianer I've added NVHPC 23.9 to the spack environment, and I was able to compile at least fmt 10.1.1 (latest available in spack 0.21.0) with C++20 and CMake 3.26.3, so I've removed the constraint on fmt from the spec here. Let's see if the CI agrees with this.

msimberg:
cscs-ci run

msimberg (approved):

msimberg:
test cscs-daint

PR #944, opened 2024-01-10
Add NVHPC pipeline with MPI enabled in CSCS CI

aurianer:
This adds a CSCS CI pipeline for nvhpc and add a constraint to spack to require the dependencies to be built with the spack compilers installed.

original error (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/5902194011), now trying to solve it with the workaround found (https://github.com/spack/spack/issues/32022)

### DIFF
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -8,7 +8,7 @@ include:
   - local: '.gitlab/includes/common_pipeline.yml'
 
 variables:
-  SPACK_COMMIT: 7d4dcd52d96b593fb5e1557794ad57a7c5eb2304
+  SPACK_COMMIT: 6cfcbc01679ebabe640f4462c8fc07626a543015
 
 base_spack_image:
   stage: spack_base


msimberg:
I suppose getting https://github.com/pika-org/pika/pull/969 in sooner would help with this PR.

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -20,6 +21,13 @@ RUN spack compiler info $COMPILER > /dev/null 2> /dev/null; compiler_missing=$?;
     if [[ $compiler_missing != 0 ]]; then \
         spack install $spack_compiler arch=$ARCH && \
         spack compiler add --scope site $(spack location -i $spack_compiler) && \
+        if [[ ! -z "$NVHPC_COMPILER" ]]; then \
+            nvhpc_version=${NVHPC_COMPILER#nvhpc@} && \
+            spack install $NVHPC_COMPILER%$spack_compiler arch=$ARCH && \
+            spack compiler rm gcc@11 && \
```
msimberg, line 27: I don't particularly love this since it's specific to the gcc that happens to be installed in the base image.

As an alternative, could you try a combination of the below:
- Use `--fresh` for the pika environment concretization, since this should end up preferring the compiler(s) you specify in the spec
- Add `COMPILER` and `NVHPC_COMPILER` to package preferences in the spack configuration. These should then be preferred to the bootstrap compiler.

.gitlab/includes/nvhpc_mpi_pipeline.yml
```diff
@@ -0,0 +1,64 @@
...
+    - compiler=${compiler//@/-}
+    - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/base/pika-$compiler:$CONFIG_TAG
+    - echo -e "compiler=$compiler" >> compiler.env
+    - echo -e "BASE_IMAGE=$PERSIST_IMAGE_NAME" >> compiler.env
+  variables:
+    DOCKERFILE: .gitlab/docker/Dockerfile.spack_compiler
+    DOCKER_BUILD_ARGS: '["BASE_IMAGE","ARCH","COMPILER","NVHPC_COMPILER"]'
+  artifacts:
+    reports:
+      dotenv: compiler.env
```
msimberg, line 45: Can this use the new templates instead?

.gitlab/pipeline.yml
```diff
@@ -9,3 +9,4 @@ include:
   - local: '.gitlab/includes/clang14_cuda11_pipeline.yml'
   - local: '.gitlab/includes/performance_gcc13_pipeline.yml'
   - local: '.gitlab/includes/gcc12_hip5_pipeline.yml'
+  - local: '.gitlab/includes/nvhpc_mpi_pipeline.yml'
```
msimberg, line 12: Maybe something in this direction:
```suggestion
  - local: '.gitlab/includes/nvhpc23_9_pipeline.yml'
```
?

msimberg (changes requested):

.gitlab/docker/Dockerfile.spack_compiler
msimberg: I meant to comment this earlier but forgot: note also that we don't mind if `build`-only dependencies (like `ninja`) are using the system compiler.

.gitlab/includes/nvhpc_mpi_pipeline.yml
msimberg: Good point. Maybe something like this:
```
~> COMPILER="gcc@11.4.0"; echo "${COMPILER}${NVHPC_COMPILER:+-}${NVHPC_COMPILER}"
gcc@11.4.0
~> COMPILER="gcc@11.4.0" NVHPC_COMPILER="nvhpc@23.9"; echo "${COMPILER}${NVHPC_COMPILER:+-}${NVHPC_COMPILER}"
gcc@11.4.0-nvhpc@23.9
```
? `${NVHPC_COMPILER:+-}` gets expanded to `-` if `NVHPC_COMPILER` is set (non-empty) and to nothing otherwise (https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html). I don't know how this interacts with `set -u` (not sure if it's set in that script) so you may need to experiment a bit, but I think it should be possible to get that to work.

Note by the way that you don't have `NVHPC_COMPILER` in the `CONFIG_TAG`. Maybe it would be good to add it there as well (to be consistent with everything else we put in `CONFIG_TAG`).

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -20,6 +21,16 @@ RUN spack compiler info $COMPILER > /dev/null 2> /dev/null; compiler_missing=$?;
     if [[ $compiler_missing != 0 ]]; then \
         spack install $spack_compiler arch=$ARCH && \
         spack compiler add --scope site $(spack location -i $spack_compiler) && \
+        if [[ ! -z "$NVHPC_COMPILER" ]]; then \
+            # Require the spack compilers instead of the system compiler
+            printf \
+            "  all:\n    require:\n    - any_of: [\"\045${NVHPC_COMPILER}\",\"\045${COMPILER}\"]" \
```
msimberg, line 27: Just wondering: what is the `\045` for?

.gitlab/docker/Dockerfile.spack_compiler
```diff
@@ -20,6 +21,16 @@ RUN spack compiler info $COMPILER > /dev/null 2> /dev/null; compiler_missing=$?;
     if [[ $compiler_missing != 0 ]]; then \
```
msimberg, line 21: I think it makes sense to add the `all:require:` also when nvhpc isn't used.

.gitlab/docker/Dockerfile.spack_compiler
msimberg: I think you can do the same as in DLA-Future: https://github.com/eth-cscs/DLA-Future/blob/70dd621e8038bba67d69be9f7e6499c4b44f7eea/ci/docker/build.Dockerfile#L55.

.gitlab/pipeline.yml
msimberg: I think you should leave the mpi variant, especially if the tests run successfully.

.gitlab/docker/Dockerfile.spack_compiler
albestro: About the `spack config add`, I think it is a limit of the CLI tool. I found also other comments about it, like https://spackpm.slack.com/archives/C5W7NKZJT/p1672863096551769.

The problem I see is that `require:` has to be an array, and `any_of` has to be an array again. So it should look like `packages:all:require:[any_of:["%gcc", "%nvcc]]`, and the parser is not so advanced to parse similar syntax.

msimberg:
Could you update the PR description here as well?

msimberg:
@aurianer could you force-push again to retrigger the CI?

.gitlab/docker/Dockerfile.spack_base
```diff
@@ -25,9 +25,6 @@ RUN apt-get update && apt-get -yqq install --no-install-recommends \
     openssl \
     pipx \
     pkgconf \
-    python3 \
-    python3-dev \
-    python3-pip \
```
msimberg, line 30: Note that python is installed even if you remove these: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/6173414941#L5232. Presumably because of pipx?

I would probably do the following:
- Keep them deleted here if we don't need it _explicitly_ otherwise. Since pipx uses python it can request installation of python, but if we remove pipx so should python.
- Instead do `spack external find --exclude python` to let spack use its own python.
- Remove the workaround for meson (at least if I understand correctly that this workaround is only required when using an external python).

.gitlab/docker/Dockerfile.spack_dependencies
```diff
@@ -1,6 +1,10 @@
 ARG BASE_IMAGE
 FROM $BASE_IMAGE
 
+# Tmp! Modify the meson package.py to find the binary
+RUN sed -i "s/self.spec.prefix.bin.meson/self.spec.prefix.local.bin.meson/" \
+    $SPACK_ROOT/var/spack/repos/builtin/packages/meson/package.py
+
```
msimberg, line 7: See other comment about not using the system python. I think if that works you should be able to remove this workaround.

.gitlab/pipeline.yml
```diff
@@ -9,6 +9,7 @@ include:
   - local: '.gitlab/includes/clang14_cuda11_pipeline.yml'
   - local: '.gitlab/includes/performance_gcc13_pipeline.yml'
   - local: '.gitlab/includes/gcc12_hip5_pipeline.yml'
+  - local: '.gitlab/includes/nvhpc23_9_pipeline.yml'
```
msimberg, line 12: While I expect NVHPC to cause "interesting" compilation failures, I think we should not trigger it automatically on pushes since, at least on the Jenkins configurations, it's by far the slowest compiler as well.

It's fine to leave this here until this is ready for merging, but I'm commenting just as a reminder that we move it down before merging.

msimberg (changes requested):

msimberg (approved):
Let's get this in. Thanks!

Don't forget to update the required checks for merges to not include the Jenkins nvhpc configurations before pressing merge (and no other merges at the same time).

PR #959, opened 2024-01-12
Add fallback to `invoke_with_exception_info` to report `what()` from `std::exception`

### LINKED ISSUE #868: Report non-pika exceptions in `pika_main` better
Currently if a non-pika exception is thrown in `pika_main`, the early reporting of that exception (added in #827) will simply print `<unknown>` for the exception, e.g.:
```
The pika runtime caught the following exception in the entry point (typically pika_main). The exception may be rethrown later by pika::init or pika::stop. The pika runtime will wait for all tasks to finish, but may be deadlocked because of the exception.
<unknown>
```
If the application then goes on to finish normally `pika::stop` or `pika::init` will then correctly report the exception later, e.g.:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  error message
```
However, if the application terminates because of other errors or never gets to shutting down the runtime, the user will be left with `<unknown>`. We can improve that error message to at least print `runtime_error::what()`.

msimberg:
Fixes #868 in a very basic way. This trickles up to e.g. `report_exception_and_*` to report `std::exception::what()` instead of `<unknown>`. Non-`std::exception`s are still reported as `<unknown>`. This does not specialize for different types of `std::*_error`s and reports them all as `std::exception`.

The message will now be e.g.:
```
The pika runtime caught the following exception in the entry point (typically pika_main). The exception may be rethrown later by pika::init or pika::stop. The pika runtime will wait for all tasks to finish, but may be deadlocked because of the exception.
std::exception:
  what():  hello
terminate called after throwing an instance of 'std::runtime_error'
  what():  hello
```
where the first time is when `pika_main` catches the exception, and the second is when `pika::init/stop` rethrows the exception and leads to termination. The duplication is better than not printing anything, as the runtime may never quit in case of an exception in `pika_main`. Previously this would have been:
```
The pika runtime caught the following exception in the entry point (typically pika_main). The exception may be rethrown later by pika::init or pika::stop. The pika runtime will wait for all tasks to finish, but may be deadlocked because of the exception.
<unknown>
terminate called after throwing an instance of 'std::runtime_error'
  what():  hello
```

### DIFF
--- a/libs/pika/errors/include/pika/errors/exception_info.hpp
+++ b/libs/pika/errors/include/pika/errors/exception_info.hpp
@@ -11,6 +11,8 @@
 #include <pika/errors/error_code.hpp>
 #include <pika/errors/exception_info.hpp>
 
+#include <fmt/format.h>
+
 #include <cstddef>
 #include <exception>
 #include <memory>
@@ -223,6 +225,10 @@ namespace pika {
         catch (exception_info const& xi)
         {
             return PIKA_FORWARD(F, f)(&xi);
+        }
+        catch (std::exception const& e)
+        {
+            return fmt::format("std::exception:\n  what():  {}", e.what());
         }
         catch (...)
         {


msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #960, opened 2024-01-12
Use fewer iterations in `async_rw_mutex` test with valgrind

msimberg:
The test can take near 300 seconds to run in release mode with valgrind. This reduces the number of iterations to something a bit more reasonable.


msimberg:
It now runs in a slightly more reasonable 30 seconds.

PR #970, opened 2024-01-16
Reenable `process_mask_flag` test everywhere in CI

msimberg:
After #922 was merged, the test should now run pretty much in any CI environment. However, this was not quite the case. This PR further slightly relaxes the regex to not care about the socket and numa node indices. For macOS it introduces a completely separate regex which expects no bindings, because thread binding is not supported on macOS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -578,7 +578,7 @@ jobs:
               -j2 \
               --timeout 500 \
               -T test \
-              -E "tests.unit.modules.threading.condition_variable2"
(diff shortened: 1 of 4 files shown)


msimberg:
cscs-ci run

msimberg:
cscs-ci run

msimberg:
I've opened #973 to slightly speed up the channel tests that take too long to run under valgrind. I've restarted that job here to get this through to merging.

PR #973, opened 2024-01-17
Reduce sizes of various channel tests when using valgrind

msimberg:
In the same vein of #960, reduce the number of iterations/tests in a few of the channel tests when using valgrind to make them run in a more reasonable time.


msimberg:
cscs-ci run

PR #975, opened 2024-01-18
Refactor CSCS CI using gitlab CI templates

aurianer:


.gitlab/includes/clang11_pipeline.yml
```diff
@@ -19,77 +19,28 @@ include:
     CMAKE_FLAGS: "-DCMAKE_BUILD_TYPE=$BUILD_TYPE -DPIKA_WITH_CXX_STANDARD=$CXXSTD \
                   -DPIKA_WITH_MAX_CPU_COUNT=256 -DPIKA_WITH_MALLOC=system \
                   -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=ON"
+    CSCS_REBUILD_POLICY: always
```
msimberg, line 22: We don't want this everywhere do we? The pika build stage would be ok to always rebuild, but the compiler and dependencies not? Or does it fetch an already built image to avoid actually rebuilding, even if this says "always"?

.gitlab/includes/common_spack_pipeline.yml
```diff
@@ -26,3 +26,69 @@ base_spack_image:
   artifacts:
     reports:
       dotenv: base.env
+
+.compiler_image_template: &compiler_image_template
```
msimberg, line 30: Is this needed?

```suggestion
.compiler_image_template:
```
? As I understand it the `&compiler_image_template` makes this into an anchor which is an alternative way to reuse parts of the configuration but I think you're not using that?

.gitlab/includes/common_spack_pipeline.yml
```diff
@@ -26,3 +26,69 @@ base_spack_image:
...
+    - configuration=$DOCKERFILE_SHA-$BASE_IMAGE-$BUILD_DIR-$CMAKE_COMMON_FLAGS-$CMAKE_FLAGS-$CI_COMMIT_SHORT_SHA-$SOURCE_DIR
+    - configuration=${configuration//-D/}
+    - CONFIG_TAG=`echo $configuration | sha256sum - | head -c 16`
+    - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/pika-$compiler-$build_type-build:$CONFIG_TAG
+    - echo -e "PERSIST_IMAGE_NAME=$PERSIST_IMAGE_NAME" >> build.env
+  artifacts:
+    reports:
+      dotenv: build.env
+
+.test_template: &test_template
```
msimberg, line 84: Same question here regarding anchors:
```suggestion
.dependencies_image_template:
  stage: spack_dependencies
  extends:
    - .container-builder
  before_script:
    - export DOCKERFILE_SHA=`sha256sum $DOCKERFILE | head -c 16`
    - CONFIG_TAG=`echo $DOCKERFILE_SHA-$BASE_IMAGE-$SPACK_SPEC | sha256sum - | head -c 16`
    - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/base/pika-$compiler-dependencies:$CONFIG_TAG
    - echo -e "compiler=$compiler" >> dependencies.env
    - echo -e "BASE_IMAGE=$PERSIST_IMAGE_NAME" >> dependencies.env
  variables:
    DOCKERFILE: .gitlab/docker/Dockerfile.spack_dependencies
    DOCKER_BUILD_ARGS: '["BASE_IMAGE","SPACK_SPEC"]'
  artifacts:
    reports:
      dotenv: dependencies.env

.build_template:
  stage: build
  extends:
    - .container-builder
    - .build_variables_common
    - .cmake_variables_common
  before_script:
    - export DOCKERFILE_SHA=`sha256sum $DOCKERFILE | head -c 16`
    - build_type=`echo $BUILD_TYPE | tr '[:upper:]' '[:lower:]'`
    - configuration=$DOCKERFILE_SHA-$BASE_IMAGE-$BUILD_DIR-$CMAKE_COMMON_FLAGS-$CMAKE_FLAGS-$CI_COMMIT_SHORT_SHA-$SOURCE_DIR
    - configuration=${configuration//-D/}
    - CONFIG_TAG=`echo $configuration | sha256sum - | head -c 16`
    - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/pika-$compiler-$build_type-build:$CONFIG_TAG
    - echo -e "PERSIST_IMAGE_NAME=$PERSIST_IMAGE_NAME" >> build.env
  artifacts:
    reports:
      dotenv: build.env

.test_template:
```

.gitlab/includes/clang14_cuda11_pipeline.yml
```diff
@@ -21,78 +21,28 @@ include:
                   -DPIKA_WITH_CUDA=ON -DPIKA_WITH_MALLOC=system \
                   -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=ON -DCMAKE_CUDA_COMPILER=c++ \
                   -DCMAKE_CUDA_ARCHITECTURES=80 -DPIKA_WITH_CXX_LAMBDA_CAPTURE_DECLTYPE=OFF"
+    CSCS_REBUILD_POLICY: always
 
 clang14_cuda11_spack_compiler_image:
-  stage: spack_compiler
-  timeout: 2 hours
```
msimberg, line 27: Do you think you still need the higher timeout here?

.gitlab/includes/clang11_pipeline.yml
```diff
@@ -19,77 +19,28 @@ include:
...
 
 clang11_spack_compiler_image:
-  stage: spack_compiler
-  needs: [base_spack_image]
   extends:
-    - .container-builder
+    - .compiler_image_template
     - .variables_clang11_config
```
msimberg, line 27: Just checking for my own understanding: Are these sensitive to the order in which blocks are listed under `extends:`? If later blocks override/use earlier blocks, would it make sense (just for "sanity") to put the `variables_clang11_config` block (and others) always first since they don't depend on other variables? I don't know if that's always the case thought.

The same comment applies of course to other configurations and the other stages as well.

msimberg (changes requested):
Only some minor cleanup and a few clarifying questions before approval. This looks like huge improvement, thank you!

.gitlab/includes/clang14_cuda11_pipeline.yml
msimberg: We can always readd it if needed.

.gitlab/includes/clang11_pipeline.yml
msimberg: What happens if a variable is redefined (we might e.g. want defaults for some things)? Or if a variable depends on another variable?

.gitlab/includes/common_spack_pipeline.yml
```diff
@@ -26,3 +26,69 @@ base_spack_image:
...
+    - echo -e "PERSIST_IMAGE_NAME=$PERSIST_IMAGE_NAME" >> build.env
+  artifacts:
+    reports:
+      dotenv: build.env
+
+.test_template:
+  extends:
+    - .cmake_variables_common
+  needs:
+    - gcc9_debug_build
```
msimberg, line 88: I think this one doesn't belong here (likely overridden by the actual configurations?).

.gitlab/includes/clang11_pipeline.yml
msimberg: Thanks for checking and the link! I was thinking of this case specifically: https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#merge-details, but all of this is good to have here for reference.

msimberg (approved):
Always welcome: `+160 −894`. Thank you!

msimberg:
Merging manually because of clariden CI runners being down. CI looks good otherwise.

PR #980, opened 2024-01-18
Prefer using predicated `condition_variable::wait`

msimberg:
This is mostly a question of style, but it gets rid of annoying warnings that have to be suppressed with `NOLINTNEXTLINE`, and is generally a bit more compact than the manual while-loop.


msimberg:
cscs-ci run

PR #981, opened 2024-01-19
Make submission to logstash quiet by default

msimberg:
Now that a few PRs have gone in that push data to logstash/elastic and it seems to be working quite well, I'm removing the verbose output of the submission. The results can anyway be browsed through kibana, and errors are reported as usual in the log. The verbose output can still be turned on as needed with the environment variable `PIKA_CI_LOGSTASH_SUBMIT_VERBOSE`, e.g. for debugging.


msimberg:
cscs-ci run

PR #982, opened 2024-01-19
Disable Jenkins configurations covered by CSCS CI

msimberg:
Now that we have pretty extensive, and well-working, coverage of CI configurations on CSCS's CI system it makes little sense to keep configurations in both Jenkins and CSCS CI. This is an attempt at disabling the debug configurations in Jenkins, except for CCE and NVHPC, which are not (yet) covered by the CSCS CI setup.

Based on https://www.jenkins.io/blog/2019/11/22/welcome-to-the-matrix/.

### DIFF
--- a/.jenkins/cscs/Jenkinsfile
+++ b/.jenkins/cscs/Jenkinsfile
@@ -49,6 +49,18 @@ pipeline {
                          values 'Debug', 'Release'
                     }
                 }
+                excludes {
+                    exclude {
+                        axis {
+                            name 'build_type'
+                            values 'Debug'
+                        }
+                        axis {
+                            name 'configuration_name'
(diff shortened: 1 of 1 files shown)


msimberg:
This seems to be working as intended. @aurianer does this look ok to you?

msimberg:
> should we run the CSCS CI at each push by default now instead of commenting?

I'd still avoid doing that. The Jenkins setup actually only runs a few of the pipelines automatically on pushes (most of them "run" but simply skip the build). We could add another `pipeline.yaml` that is automatically triggered, but only with a small subset of the pipelines.

msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #984, opened 2024-01-19
Store thread pool name by value instead of reference in `thread_pool_init_parameters`

msimberg:
Storing it by reference is a huge footgun as it requires the user to make sure the name stays alive long enough. This changes the member to a regular value to avoid surprises. The member doesn't need to be a reference for performance or other reasons. The `standalone_thread_pool_scheduler` test did not keep the name alive long enough (it was passed as a prvalue).


msimberg:
cscs-ci run

PR #994, opened 2024-01-23
Fix forwarding of return type from `execution::then` callable

msimberg:
The result of the callable in `then` (and `cuda::then_on_host`) was previously taken by universal reference (`auto&&`) but then moved into the successor receiver. This would be wrong when the return type of the callable is a reference, and fail to compile if the successor receiver also expects a reference. This moves the call of the callable to simply directly pass the result into `set_value` instead of going via an intermediate local variable. This also adds a regression test for `then`.


msimberg:
cscs-ci run

PR #995, opened 2024-01-23
Fix data race in `condition_variable2` test

msimberg:
The `t1_feedback` variable is read and written without synchronization. Use relaxed atomics for the variable instead.


msimberg:
cscs-ci run

msimberg:
cscs-ci run

PR #997, opened 2024-01-23
Add a second pipeline to run a subset of the CI and the full pipeline on merge

### LINKED ISSUE #996: Run only a subset of CSCS CI pipelines on PRs before merge
With Jenkins we only run a small subset of CI pipelines when testing a PR itself, and then the full set of CI pipelines before a merge. Currently we run _all_ CSCS CI pipelines on `cscs-ci run`, which is probably a bit overkill. We should reduce the number of pipelines run before the merge to avoid overloading the CI.

aurianer:
Fix #996

Now we have 2 pipelines:
- one with a subset which runs at each push to a PR against or push directly to one of the CI enabled branches.
- one that runs only on push against one of the CI enabled branches.
- the 2 pipelines i.e. the full set of CI is triggered on the merge queue as it is a CI enable branch in both pipelines.
- the comment to trigger the pipelines are respectively `cscs-ci run ci` and `cscs-ci run ci-on-merge`

### DIFF
--- a/.gitlab/pipeline_on_merge.yml
+++ b/.gitlab/pipeline_on_merge.yml
@@ -5,7 +5,6 @@
 # file BOOST_LICENSE_1_0.rst or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 include:
-  - local: '.gitlab/pipeline.yml'
   - local: '.gitlab/includes/dockerhub_pipeline.yml'
   - local: '.gitlab/includes/gcc9_pipeline.yml'
   - local: '.gitlab/includes/gcc9_cuda11_pipeline.yml'


.gitlab/pipeline_on_merge.yml
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2024 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file BOOST_LICENSE_1_0.rst or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+include:
+  - local: '.gitlab/pipeline.yml'
```
msimberg, line 8: Nice, good idea!

msimberg (approved):

.gitlab/pipeline.yml
```diff
@@ -6,19 +6,8 @@
 
 include:
   - local: '.gitlab/includes/dockerhub_pipeline.yml'
```
msimberg, line 8: Sorry, I think this can also be removed from here since it's more or less a duplicate of the CircleCI configuration.

msimberg (changes requested):

msimberg (approved):

PR #1003, opened 2024-01-24
Update CI image to use Ubuntu 23.10 as base

msimberg:
Updates clang and related tools to version 16, based on ubuntu 23.10.

clang-format 16 introduces some major and some minor formatting changes. The major change is that it seems to change the indentation of multiline macros. I couldn't find a setting to change this back to the previous formatting so these changes we'll have to accept as they are. Additionally, I've chosen some hopefully reasonable values for a few new options in 15 and 16. Interestingly, there is now an option to insert (or remove) separators in integer literals. I've set the configuration to add separators every four digits for binary and hexadecimal literals as I think this makes it much easier to count the positions of the digits. For decimal literals I've set the option to "keep whatever is there". Another alternative would be every three digits but I found this makes some version literals (which are logically grouped by 100s) read a bit weirdly. Happy to change this if you prefer something else. The current option means that we can add separators as we see fit and clang-format won't change it. It's also possible to set it to "always remove separators".

clang-tidy also has some new checks. In the MoodyCamel concurrent queue it warns about bugprone assignment in `if` statements. I've silenced these. It also has a new check for optional access without a `has_value/bool()` guard. I've silenced the warning where we assert first that the optional is non-empty. I've also updated a number of accesses to the unchecked `operator*` since we already check for a value first, instead of using the checking and throwing `value()`.


libs/pika/coroutines/include/pika/coroutines/detail/context_posix.hpp
```diff
@@ -165,7 +165,7 @@ namespace pika::threads::coroutines {
         // some platforms need special preparation of the main thread
         struct prepare_main_thread
         {
-            constexpr prepare_main_thread() {}
+constexpr prepare_main_thread() {}
```
msimberg, line 168: To do: why is clang-format doing this?

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -116,7 +116,7 @@ inline std::uint32_t prev_rank(std::uint32_t rank, std::uint32_t size)
 // at the same time
 inline std::uint32_t make_tag(std::uint32_t rank, std::uint32_t iteration, std::uint32_t ranks)
 {
-    std::int64_t tag = (rank + (iteration * ranks)) & 0xffffffff;
+    std::int64_t tag = (rank + (iteration * ranks)) & 0xffff'ffff;
```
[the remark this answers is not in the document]
msimberg: Absolutely, though not set in stone. This is configurable if you prefer another style. See the box about separators https://en.cppreference.com/w/cpp/language/integer_literal and https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#integerliteralseparator.

While you commented I was updating the PR description, also see that for an explanation.

msimberg:
This is ready for review though I would still like to at least briefly see if I can convince clang-format to do something different in `context_posix.hpp` as it creates very weird indentation in that file right now.

libs/pika/coroutines/include/pika/coroutines/detail/context_posix.hpp
msimberg: Changing the ridiculous `namespace pika { namespace threads { namespace coroutines { namespace detail { namespace posix { namespace pth {` to use a slightly less ridiculous, but still ridiculous, `namespace pika::threads::coroutines::detail::posix::pth {` seems to have convinced clang-format to format the file sensibly.

biddisco (commented):
LGTM

msimberg:
cscs-ci run ci-on-merge

PR #1006, opened 2024-01-25
Skip the CI pipeline ci-on-merge on pushes, trigger it on merge

### LINKED ISSUE #999: Investigate how to skip the ci-on-merge pipeline before it is added to the merge queue
Currently we require ci-on-merge pipeline status to pass both before adding to the merge queue and before merging into main.
We would like to limit the requirement to trigger it only before actually merging into main.

aurianer:
Fix #999

We have two pipelines for pika in CI: `ci` to be run at each push on PRs or CI enabled branches (this works well so is unchanged in this PR) and `ci-on-merge` to be run only on merges to main.
**Problem**: To merge a PR, we are requiring the `ci-on-merge` status to pass (github settings of the main protected branch). In those settings, the required check is required both to add the PR to the merge queue (we don't want that since we want this pipeline only to run on the merge queue right before the merge to main) and to actually merge it to main (we want this run). The problem is the following: on github actions `merge_group` dissociate adding to the merge queue from actually merging the PR whereas the github settings for required checks don't and require a passed status check before adding the PR to the merge queue.
**Temporary solution from before**: The temporary solution adopted was to manually trigger ci-on-merge with a comment on the PR just before adding it to the merge queue.
**Changes of this PR**: However this can be automated, in this PR we add rules on the jobs of `ci-on-merge` so that they only run if the branch is `gh-readonly*`, `main` or `release*`. This PR also merges back the 2 pipelines, since the dissociation is not needed anymore (same settings in the setup CI page to run at each push).


.gitlab/pipeline_on_merge.yml
```diff
@@ -1,20 +1,43 @@
-# Copyright (c) 2024 ETH Zurich
-#
-# SPDX-License-Identifier: BSL-1.0
-# Distributed under the Boost Software License, Version 1.0. (See accompanying
-# file BOOST_LICENSE_1_0.rst or copy at http://www.boost.org/LICENSE_1_0.txt)
-
 include:
+  - local: '.gitlab/includes/skip_job.yml'
```
msimberg, line 2: An alternative would perhaps be to put all jobs back in `pipeline.yml`, and skip the same set of jobs there instead. Then we wouldn't need the `ci-on-merge` pipeline at all. As far as I can tell this would be equivalent, but simpler as there's only one pipeline and there would be no need for the dummy job.

.gitlab/pipeline_on_merge.yml
```diff
@@ -1,20 +1,43 @@
...
+    rules:
+      - if: $CI_COMMIT_BRANCH !~ /^gh-readonly-queue/
   - local: '.gitlab/includes/dockerhub_pipeline.yml'
+    rules:
+      - if: $CI_COMMIT_BRANCH == "main" || $CI_COMMIT_BRANCH =~ /^gh-readonly-queue/
```
msimberg, line 7: What about release branches?

.gitlab/pipeline_on_merge.yml
```diff
@@ -1,20 +1,43 @@
-# Copyright (c) 2024 ETH Zurich
-#
-# SPDX-License-Identifier: BSL-1.0
-# Distributed under the Boost Software License, Version 1.0. (See accompanying
-# file BOOST_LICENSE_1_0.rst or copy at http://www.boost.org/LICENSE_1_0.txt)
-
 include:
+  - local: '.gitlab/includes/skip_job.yml'
+    rules:
+      - if: $CI_COMMIT_BRANCH !~ /^gh-readonly-queue/
```
msimberg, line 4: _If_ we end up needing the dummy job (see other comment) then make sure this is the complement of the filter in the other jobs.

msimberg:
Could you update the title and description please?

msimberg (approved):

PR #1008, opened 2024-01-27
Install pika in CSCS CI configurations

### LINKED ISSUE #906: Install pika in CSCS CI configurations
This allows testing that pika installs correctly, including running the build unit tests against the installed pika.

aurianer:
Fix #906

### DIFF
--- a/.gitlab/docker/Dockerfile.spack_build
+++ b/.gitlab/docker/Dockerfile.spack_build
@@ -16,7 +16,8 @@ RUN spack -e pika_ci spec -lI $spack_spec
 
 # Configure & Build
 RUN spack -e pika_ci build-env $spack_spec -- bash -c "cmake -B${BUILD_DIR} ${SOURCE_DIR} \
-    $CMAKE_COMMON_FLAGS $CMAKE_FLAGS && cmake --build ${BUILD_DIR} --target all tests examples"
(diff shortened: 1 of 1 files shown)


.gitlab/docker/Dockerfile.spack_build
```diff
@@ -49,3 +49,7 @@ RUN CTEST_XML=$PWD/ctest.xml; \
             --no-compress-output \
             -R tests \
             -no-tests=error"
+
+# Install pika
+RUN spack -e pika_ci build-env $spack_spec -- bash -c "cmake --install ${BUILD_DIR}" && \
+    ./hello_world --pika:bind=none
```
msimberg, line 55: Sorry, I was definitely not clear in #906.

We already have unit tests that test the installed pika: `tests.unit.build*`. There are two sets, one that uses the build directory which can always be run and one that uses the install directory which is silently skipped if pika is not installed (because it'd be a constant source of test failures when developing pika).

I think it should be sufficient to simply add `install` to the build targets on line 19.

msimberg: That said, if we add `install` to the targets, you could additionally add a `hello_world` to the test stage, since the examples should get installed.

msimberg (approved):

PR #1014, opened 2024-02-01
Keep `-Werror` flags in CSCS CI builds of pika

msimberg:
Spack disables `-Werror` flags by default. This makes some sense for building dependencies in pika, but is not so good for CI. This enables `-Werror` flags again for the build stage of pika itself, but not while building pika dependencies.


msimberg:
The `-Werror` flag seems to be working again. The merge failed because of an unused variable warning: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/6098112258#L544.

PR #1017, opened 2024-02-05
Unrevert #872

msimberg:
Reapplies #872, i.e. ensures that cuBLAS and cuSOLVER handles are managed by the `cuda_pool` instead of being released at thread exit (since they were previously `thread_local`s). First, #872 was buggy because of the indexing being made based on the global thread number, which was wrong for non-default pools (the number was always local to the thread pool). See #1016 for more details. Second, the index used for choosing a handle outside of the pika thread pools was done without wrapping the index, leading to out-of-bounds accesses. I've expanded the tests to cover this better.

Note: in the last commit I've reverted the indexing fix to see if tests fail in CI. I also need to verify this in SIRIUS before merging.


msimberg:
On the previous commit, without the fix to the indexing, the `cuda_device_reset` test segfaults and times out: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/6104363449#L471. The PR is now in the "fixed" state and ready for reviews.

msimberg:
The only differences from #872 are the two last commits, i.e. the fix to the indexing and an update to the unit test.

msimberg:
@aurianer @biddisco, I'm going to ahead with this PR and #1016 to start preparing the 0.22.2 release. Please feel free to comment even after merging. I'd like to aim for a 0.22.2 release by the end of the week, assuming no surprises show up.

PR #1018, opened 2024-02-05
Run debug and release builds for each CSCS CI configuration

aurianer:
- Use the matrix functionality to run both Debug and Release builds for each configuration.
- We are using 2 different jobs for the test stage (one release and one debug). 

The `image:` field of a gitlab CI job doesn't allow customization with a variable (build type for example). Since it's evaluated before running the script of the same job, a scripting logic to change the image name is not possible either. That's why we currently have 2 jobs for the test step (one for the release mode, the other for the debug mode)

### DIFF
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -63,10 +63,18 @@ base_spack_image:
     reports:
       dotenv: dependencies.env
 
+.parallel_build_types:
+  parallel:
+    matrix:
+      - BUILD_TYPE:
+          - Debug
+          - Release
+
 .build_template:
   stage: build
   extends:
     - .container-builder
+    - .parallel_build_types
     - .build_variables_common
     - .cmake_variables_common
   before_script:
@@ -84,6 +92,7 @@ base_spack_image:
 .test_template:
   extends:
     - .cmake_variables_common
+    - .parallel_build_types
   script:
     - spack arch
     - export CTEST_XML=$PWD/ctest.xml
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -69,33 +69,34 @@ base_spack_image:
       - BUILD_TYPE:
           - Debug
           - Release
+  variables:
+    DOTENV_FILE: "build-$BUILD_TYPE.env"
 
 .build_template:
   stage: build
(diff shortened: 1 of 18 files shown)


.gitlab/includes/common_spack_pipeline.yml
```diff
@@ -89,4 +105,3 @@ base_spack_image:
     - export CTEST_XML=$PWD/ctest.xml
     - trap "${SOURCE_DIR}/.gitlab/scripts/collect_ctest_metrics.sh ${CTEST_XML}" EXIT
     - spack -e pika_ci build-env $spack_spec -- bash -c "ctest --output-junit ${CTEST_XML} --label-exclude COMPILE_ONLY --test-dir ${BUILD_DIR} -j$(($(nproc)/2)) --timeout 120 --output-on-failure --no-compress-output --no-tests=error"
-  image: $PERSIST_IMAGE_NAME
```
msimberg, line 92: How does the test stage now know which image to use if `image` is not set explicitly?

msimberg: Ah, I see. I missed that, thanks!

All right, let's go with this for now and if we become wiser in the future we can change it.

My only very minor aesthetic nitpick would be to name the variables `PERSIST_IMAGE_NAME_{RELEASE,DEBUG}`, i.e. keep it all-caps if possible.

msimberg (approved):
Thank you!

.gitlab/includes/performance_gcc13_pipeline.yml
```diff
@@ -41,7 +41,7 @@ performance_gcc13_release_test:
     - .test_common_mc
     - .cmake_variables_common
   needs: [performance_gcc13_release_build]
-  image: $PERSIST_IMAGE_NAME
+  image: $PERSIST_IMAGE_NAME_RELEASE
```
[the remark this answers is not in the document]
msimberg: Feel free to add a debug run step also for the performance tests, then it's even closer to the non-performance CI jobs.

Of course, running performance tests in debug mode isn't to figure out how well we can do, but it's useful to see how much of a penalty we have in debug mode.

msimberg:
I've just merged this manually as it looks like all the pipelines were correctly triggered. There was a test failure in one of the release builds and we'll have to keep a bit of a closer eye on those from now on to see if there are new tests that regularly fail.

PR #1019, opened 2024-02-06
Disable recycling of threads when `PIKA_WITH_THREAD_STACK_MMAP` is disabled

msimberg:
When not using `mmap` for thread stacks we can rely on the allocator to do performant recycling instead of us manually recycling threads. In some situations it seems like not using `mmap` and relying on the allocator recycling memory is faster than us recycling threads. I'm leaving `mmap` as the default though as it's not clear if it's always faster to not use `mmap`.

### DIFF
--- a/.gitlab/includes/clang15_pipeline.yml
+++ b/.gitlab/includes/clang15_pipeline.yml
@@ -16,7 +16,8 @@ include:
     SPACK_SPEC: "pika@main arch=$ARCH %${COMPILER} malloc=system cxxstd=$CXXSTD ^boost@1.80.0 \
                  ^hwloc@2.8.0"
     CMAKE_FLAGS: "-DPIKA_WITH_CXX_STANDARD=$CXXSTD -DPIKA_WITH_MALLOC=system \
-                  -DPIKA_WITH_SPINLOCK_DEADLOCK_DETECTION=ON -DPIKA_WITH_UNITY_BUILD=OFF"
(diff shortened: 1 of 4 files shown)


msimberg:
I've done some preliminary tests of this and here's a brief summary:

|  | mmap | malloc |
| - | - | - |
| Thread/stack recycling  | Current default, performs well  | Performs the same as the default |
| No recycling | Performs terribly (mmap is expensive?) | Performs ~10% faster than the default |

I'm guessing the reason for the difference between recycling+mmap and no recycling+malloc is that e.g. mimalloc's memory reuse is far more optimized than our naive lock-based thread recycling.

By not using mmap we do lose e.g. the guard page, and possibly lazy allocation of pages. I've yet to test if this has a performance impact in DLA-Future.

msimberg:
Like #1023, this is ready for reviews but it needs an update to the performance CI reference before merging.

msimberg:
This seems to not have a very impact in DLA-Future, but does have a big impact in specialized tests in pika. At this point I'm going to make the change but keep `PIKA_WITH_THREAD_STACK_MMAP=ON` as the default and we can do further testing to consider changing the default.

msimberg:
I'm separating out the change to use mimalloc in performance tests into another PR just to see the effect of this change in isolation.

msimberg:
At least the `task_overhead` test in CI reports a 50% slowdown when disabling mmap with jemalloc. I'll definitely keep mmap enabled for now. I suspect jemalloc and mimalloc may perform quite differently, and I'll try disabling mmap again in the other PR to use mimalloc in performance tests.

PR #1023, opened 2024-02-07
Always use moodycamel queue for default FIFO queue

msimberg:
Like in #924, this is important for systems where the compiler doesn't emit 128-bit lockfree atomics, but instead falls back to locks such as on Grace with GCC 12. Since `lockfree_fifo` is used in the default scheduler, changing `lockfree_fifo` to use the moodycamel queue affects the default scheduler significantly.


msimberg:
The CI performance tests show:

msimberg:
This is ready for review, but I will have to update the performance CI references before merging if approved.

msimberg:
This seems to be a good change especially on Grace when tested in DLA-Future. There's also a small improvement on x86. I've updated the performance CI reference.

@biddisco, @aurianer any comments?

PR #1025, opened 2024-02-08
Add santis runner pipeline

aurianer:
This PR uses the fact that variables gets overriden by the latest one of the same name. So the `.container-builder-cscs-gh200` is the one setting the latest tag. See the tag in [one of the first pipeline of this PR](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/6169523086), it is set to `santis-container-builder`.
I originally tried set the container builder in a gitlab variable and then use the variable in a extend but the variable was not expanded (cf [the first job](https://cicd-ext-mw.cscs.ch/ci/pipeline/results/error/371fcef737f4ab2a78eb1dac61aa3899ac4a7ee6f9ccb42b89af965edd9533e0.html)).
I also tried to set the container builder extends clause at the latest possible, but it was complaining when I was removing it from the template job.

Note: it changes the fetch method from urllib to curl because of the recurrent `FetchError`s in spack while downloading the sources.

### DIFF
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -28,14 +28,19 @@ variables:
     reports:
       dotenv: base.env
 
+# We allow failure since santis is unstable
 base_spack_image_aarch64:
   extends: [.container-builder-cscs-gh200, .base_spack_image]
+  allow_failure: true
 base_spack_image_x86_64:
   extends: [.container-builder-cscs-zen2, .base_spack_image]
 
 make_multiarch_image:
   stage: spack_base
-  needs: [base_spack_image_aarch64, base_spack_image_x86_64]
+  needs:
+    - job: base_spack_image_aarch64
+      optional: true
+    - job: base_spack_image_x86_64
   extends: .make-multiarch-image
   after_script:
     - echo -e "BASE_IMAGE=$PERSIST_IMAGE_NAME" >> multiarch.env
--- a/.gitlab/includes/gcc13_santis_pipeline.yml
+++ b/.gitlab/includes/gcc13_santis_pipeline.yml
@@ -23,12 +23,14 @@ gcc13_santis_spack_compiler_image:
   extends:
     - .variables_gcc13_santis_config
     - .compiler_image_template_santis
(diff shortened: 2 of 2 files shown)


msimberg:
> Looks like the build for santis is finally working. I would be in favour of moving the `make_multiarch_image` as a part of the `spack_base` stage so that all the pipelines still fit in the screen without having to scroll right, @msimberg what do you think?

That depends on the screen size ;) but sure, if you mean `spack_base` stage, and still as separate jobs then no problem.

> For this PR, I was thinking to remove the `gcc13_pipeline.yml` and replacing it with the santis one, is that fine with you? or would you like to keep both?

I would absolutely still keep the non-santis pipeline as we have no guarantees about the availability of santis (plus removing the non-santis pipeline would lose the test run stage for gcc 13).

.gitlab/pipelines_on_push.yml
```diff
@@ -6,6 +6,7 @@
 
 include:
   - local: '.gitlab/includes/gcc13_pipeline.yml'
+  - local: '.gitlab/includes/gcc13_santis_pipeline.yml'
```
msimberg, line 9: Can we move this to the "on merge" section?

msimberg (approved):
Looks good to me, thanks.

> I would be in favour of moving the make_multiarch_image as a part of the spack_base stage so that all the pipelines still fit in the screen without having to scroll right

Did you decide not to do the above in the end? If so, why?

> cf this job

Could you update the above in the PR description? The last `cf this job` has no link.

msimberg:
> > Did you decide not to do the above in the end? If so, why?
> 
> I did it, it was not pushed sorry
> 
> > Could you update the above in the PR description? The last `cf this job` has no link.
> 
> I removed the cf since the links will anyway point to nothing after a while with the cleanup Andreas but in place. It would take me some unecessary time to look it up again. Will try to be more thorough next time not to forget the link

No worries, and thanks!

It looks like jfrog is unhappy. My approval stands so let's get this merged, but I'd like to see it get merged cleanly through CI, so let's see if jfrog starts behaving better soon and try again.

PR #1028, opened 2024-02-09
Bump spack sha to the same as the nvhpc PR

aurianer:


msimberg:
I suppose you would need the patch to meson to continue to the `hipify-clang` stage: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/6135413074#L959.

msimberg:
@aurianer can this be closed? If I understand correctly this was only used for testing for #944?

Issue #1031, opened 2024-02-15
Segmentation fault in `task_overhead` test


msimberg:
E.g. https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/6178114719#L41 (it doesn't say explicitly which test, but `task_overhead_report_test` is the first test to run). This seems to be something newly introduced but I'm not entirely sure what would have caused it. It's reproducible in the same container image on Piz Daint but I have not backtrace.

msimberg:
This seems to have been worked around by #1033.

PR #1033, opened 2024-02-15
Use moodycamel queue by default for terminated items

msimberg:
Despite being a FIFO queue, it seems like the benefit of using the moodycamel queue outweighs the the benefits of using a LIFO queue (for better cache reuse) for terminated threads.


msimberg:
On the performance CI this gives a roughly 15% speedup:

msimberg:
It's possible that the cause of the segfaults in #1031 is the `deque` and that this will "fix" the issue (at least work it around for the default scheduler, which with this PR no longer uses the deque). From local testing this PR does not fail on the task overhead test, while master regularly does. It's of course possible that the moodycamel queue simply hides another underlying issue.

msimberg:
I'm going to go ahead with this as it seems to improve performance and apparently helps with failing tests.

PR #1035, opened 2024-02-16
Store scheduler counter data on separate cache lines

msimberg:
This doesn't make a measurable difference in performance, but the counters occasionally show up in `perf c2c` output and it makes sense to separate them to different cache lines.


biddisco (approved):
LGTM

Issue #926, opened 2024-01-05
CSCS-CI related issues


aurianer:
- [ ] Create one HIP pipeline (#846)
- [ ] Add the gcc10 config with APEX (#930)
- [ ] Reenable stdexec in clang14 configuration (#894)  (for now the spack cxx standard for this configuration is 17, the cxx standard for building is 20 and stdexec is disabled)
- [ ] Install pika in configurations (#906)
- [ ] Add CSCS CI pipeline with nvhpc (#925)
- [ ] Reenable gcc12 cuda12 CSCS CI pipeline once cuda driver sufficient on clariden (#884)
- [ ] Try to use the CSCS spack build cache in CI (#789)
- [ ] Reduce duplication in CSCS CI configurations (#862)
- [ ] #962
- [ ] Enable both release and debug build types for all configurations
- [ ] Don't forget to add the `-Wno-error=array-bounds` in the GCC 13 in Release mode (for now we only build debug)
- [ ] Potentially move the configurations to another platform than daint once one is officially available for CI. Clariden is not intended for that but more for ML.
- [ ] Move the performance tests to CSCS CI.

Notes: 
- the gcc11 configuration is using gcc@11.4.0 in CSCS CI whereas it uses gcc@11.2.0 in jenkins
- the gcc9 cuda11 configuration is pinning fmt to version 9

msimberg:
Perhaps worth opening issues for the following as well:
- GCC 10 + APEX configuration
- Release builds
- Performance CI (the one on Jenkins, not the one pushing to elastic)
?

PR #1037, opened 2024-02-16
Move mc pipelines to eiger in CSCS CI

aurianer:
Part of #926

On eiger jobs are submitted through firecrest. The related tokens are on the setup page and were added by Andreas so that the job is submitted through the jenkssl account.


msimberg (approved):
Very nice, thank you.

PR #1041, opened 2024-02-19
Don't reference count in `start_detached`

msimberg:
Since there are no other references to the operation state, we statically know to release the operation state whenever `set_*` is called on the `start_detached` receiver. This avoids a an unnecessary atomic increment and decrement.


msimberg:
Nice, both valgrind and asan seem to report that the operation state isn't actually freed now. I have some more work to do.

msimberg:
The dummy scheduler in the `algorithm_execute` was a bit too dummy, in the sense that it was never calling `set_*`. A scheduler that does this would be very much ill-formed in a real situation, so I don't think it's worth going back to the reference-counted version to cover for a case like this.

PR #1042, opened 2024-02-19
Use explicit blacklist for known failing tests in CI

msimberg:
This allows running all tests despite known failures (in particular for sanitizer configurations), and the full workflow can be set to required for merges.


msimberg:
This is ready for review but merging needs to be coordinated with other PRs. I have not yet required the sanitizer status checks in the repo configuration and that needs to be done before merging. Requiring them will currently block merging of other PRs so I'm waiting with enabling them.

PR #1043, opened 2024-02-21
Boost or BOOST cmake change

biddisco:
I can't remember why I have this patch in my branch, but the cmake policies changed to use Project name as opposed to PROJECT name in places and this shut up some warnings in my builds.

### DIFF
--- a/cmake/pika_setup_boost.cmake
+++ b/cmake/pika_setup_boost.cmake
@@ -54,9 +54,9 @@ if(NOT TARGET pika_dependencies_boost)
 
   # We are assuming that there is only one Boost Root
   if(NOT BOOST_ROOT AND "$ENV{BOOST_ROOT}")
-    set(BOOST_ROOT $ENV{BOOST_ROOT})
+    set(Boost_ROOT $ENV{BOOST_ROOT})
   elseif(NOT BOOST_ROOT)
-    string(REPLACE "/include" "" BOOST_ROOT "${Boost_INCLUDE_DIRS}")
+    string(REPLACE "/include" "" Boost_ROOT "${Boost_INCLUDE_DIRS}")
   endif()
 
   add_library(pika_dependencies_boost INTERFACE IMPORTED)


msimberg:
Does this relate to https://github.com/pika-org/pika/pull/885? Will your change work with older versions of CMake? Do we need to bump the minimum version (not above 3.24 ideally, because it's useful to keep around for HIP CMake issues)? @aurianer do you know? See https://cmake.org/cmake/help/latest/policy/CMP0144.html.

biddisco:
aha. my bad. I didn't realize that another policy change already fixed this warning. 
To be correct "The OLD behavior for this policy is to ignore <PACKAGENAME>_ROOT variables if the original <PackageName> has lower-case characters. The NEW behavior for this policy is to use <PACKAGENAME>_ROOT variables."

The old behaviour was to use Boost_ROOT, but you can still do that, however, now BOOST_ROOT is also allowed. The old method isn't being removed, but the case sensitive restrivtion is being dropped. I will close this and ditch the patch from my branch.

msimberg:
Thanks both for the clarification. Should we be setting the new policy in `pika-config.cmake` as well (as locally as possible, I don't know if it's scoped or sticks around)?

msimberg:
> I would vote to change nothing or to delete the line setting the CMP0074 policy. What do you think?

I think independently of CMP0144 we should be able to delete setting the CMP0074 policy. It should anyway use the new behaviour since 3.12 if I understand correctly (https://cmake.org/cmake/help/latest/policy/CMP0074.html).

It sounds like they basically reversed CMP0074 with CMP0144. CMP0074 says:

> The OLD behavior for this policy is to ignore `<PackageName>_ROOT` variables. The NEW behavior for this policy is to use `<PackageName>_ROOT` variables.

CMP0144 says:

> The OLD behavior for this policy is to ignore `<PACKAGENAME>_ROOT` variables if the original `<PackageName>` has lower-case characters. The NEW behavior for this policy is to use `<PACKAGENAME>_ROOT` variables.

PR #1044, opened 2024-02-22
Explicitly mark destructors of `any_sender` and related classes `noexcept`

msimberg:
Using senders with `noexcept(false)` destructors (such as `require_started`) with `any_sender` otherwise fails to compile because of mismatching exception specifications on the base class (`noexcept`) and the deriving class (`noexcept(false)`). This adds a test which attempts to compile `require_started` senders in `any_sender`.

### DIFF
--- a/libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
@@ -31,12 +31,24 @@
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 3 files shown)


msimberg:
I've disabled the `require_started_mode` with GCC 9 (in addition to when using stdexec) since as far as I can tell it's 1. a compiler bug and 2. the mode is mostly useful for testing and shouldn't really be used directly by consumers. At the same time I've slightly refactored the choice of enabling or disabling `require_started_mode` into something hopefully a bit more readable.

PR #1051, opened 2024-03-04
Fix data race in when polling for CUDA events

msimberg:
See commit message for description of the data race. The MPI polling already does the same as the fixed version on this PR, and does not have a data race.


biddisco (approved):
LGTM

PR #1052, opened 2024-03-04
Add `jthread1` and `thread_mf` tests to tsan blacklist in CI

msimberg:
These tests also occasionally fail with thread sanitizer segfaults, not data races or similar.


msimberg:
Merging directly as there are no code or CI changes, besides the addition of the two tests to the blacklist.

PR #1057, opened 2024-03-12
Report lines of code to elastic

msimberg:
This reports the number of lines of code to elastic using https://github.com/flosse/sloc. I filter out the line counts for individual files before sending the data. I also only send the data for the `lib/pika` directory, excluding tests and examples. What remains are the total lines of code in `libpika`, optionally categorized by filetype and type of line (source line, comment, empty line etc.).

The CI job runs on the CSCS CI infrastructure to be able to send data to elastic. The `sloc` tool now runs on a compute node, which is somewhat overkill. An alternative would be to run that step on rosa as well, but it leads to a slightly more complicated job description (since only whitelisted images are allowed to run on rosa, it would require running docker inside docker, which is awkward in the configuration file). If in the future it's possible to run this on rosa with fewer resources we should change it.

### DIFF
--- a/.gitlab/docker/Dockerfile.sloc_base
+++ b/.gitlab/docker/Dockerfile.sloc_base
@@ -1,0 +1,12 @@
+ARG BASE_IMAGE
+FROM $BASE_IMAGE
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get -yqq install --no-install-recommends \
+    curl \
+    jq \
+    moreutils \
+    npm \
+  && rm -Rf /var/lib/apt/lists/* && \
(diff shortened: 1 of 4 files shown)


msimberg:
@aurianer would you mind having a look? In particular the `stage` for these jobs is a bit awkward now. In #1066 I've removed the stages altogether, which implicitly puts everything into a `test` stage. In this case I don't want sloc jobs to depend on the regular build jobs, and vice versa. For that I need to put the sloc jobs in the `spack_base` stage. Do you have other suggestions for how to do this cleanly? I'd be ready to go ahead with #1066 but I don't want to go ahead and merge it before you've had a look.

PR #1060, opened 2024-03-13
Add faster `invoke_result` for cases where we know we're not dealing with member functions

msimberg:
In the spirit of #1058, add a `invoke_result_plain_function` type trait for use in cases where we know that the function is not a member function and we don't need the full `std::invoke_resutl`. This is true at least in cases where we're checking if a CPO is invocable. More generic cases where the callable comes from the user continue to use `std::invoke_result`.


msimberg:
Merging manually. All important CI configurations finished successfully, and the NVHPC configuration is taking too long.

PR #1066, opened 2024-03-15
Remove stages from CSCS CI configurations

msimberg:
I'm just testing, don't take this too seriously. I found the stages to be awkward for #1057 and apparently it's possible to not use stages in newer versions of GitLab. I'm testing this to see how the UI handles the lack of stages. Dependencies are still of course respected, so everything should work as before.


msimberg:
Merging manually since the CSCS CI pipelines are triggered correctly with these changes. I will not wait for the builds themselves to finish.

PR #1081, opened 2024-03-19
Reenable nvhpc CI configuration with longer timeout on test step

msimberg:


msimberg:
I'm closing this without merging. The nvhpc image is too big to be reliably used on the Piz Daint GPU nodes. We'll reenable the test step once the maintenance is over and hopefully won't have regressed too much by then.

PR #1083, opened 2024-03-21
Add "Getting help" section to documentation

msimberg:
Link to GitHub issues, discussion, and the matrix space.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -36,6 +36,13 @@ The example above should output something like:
    Is this pi: 3.14?
    The result is 785942.2826
 
+Getting help
+============
+
+Please report issues on `GitHub <https://github.com/pika-org/pika/issues>`_. General questions and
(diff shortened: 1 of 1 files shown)


msimberg:
> I'd consider having a single entry point for general questions and discussions, not to spread the information around too much.

That's a good point. I thought GitHub discussions (forum style) and matrix (real time chat) would be different enough to justify having both, but I'd definitely consider only having one of them if you think it's more straightforward for users. Neither of them has actually been used yet by anyone so I can't say which one is more popular (if either of them will ever be "popular"...).

msimberg:
Based on a further discussion I've updated the "Getting help" section to only include GitHub issues and discussion at the moment. The matrix space still exists but I will not be advertising it explicitly for the reasons outlined by @RMeli above. Thank you @RMeli for the feedback!

PR #1093, opened 2024-04-16
Replace logging implementation with spdlog

msimberg:
This replaces the internal (boost logging-based) logging implementation with spdlog (https://github.com/gabime/spdlog). At the same time I've noticeably simplified the logging setup.

Compared to before, there is only one log now which can be used with `PIKA_LOG(level, "msg {}", arg)`. Logging is done to `stderr` by default. The default log level is warnings and more severe. The default logging output looks like this now (a snippet from `hello_world`):
```
[2024-04-16 17:27:38.942414484] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180341] [pool:0000/0003/0003] [parent:----/----] [task:0x7f01ec028b00/----] [runtime.cpp:400/set_state] runtime_state::startup
[2024-04-16 17:27:38.942424521] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180341] [pool:0000/0003/0003] [parent:----/----] [task:0x7f01ec028b00/----] [runtime.cpp:1023/run_helper] (4th stage) run_helper: ran startup functions
[2024-04-16 17:27:38.942433060] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180341] [pool:0000/0003/0003] [parent:----/----] [task:0x7f01ec028b00/----] [runtime.cpp:1026/run_helper] (4th stage) runtime::run_helper: bootstrap complete
[2024-04-16 17:27:38.942440892] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180341] [pool:0000/0003/0003] [parent:----/----] [task:0x7f01ec028b00/----] [runtime.cpp:400/set_state] runtime_state::running
[2024-04-16 17:27:38.942450837] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180341] [pool:0000/0003/0003] [parent:----/----] [task:0x7f01ec028b00/----] [runtime.cpp:1032/run_helper] (last stage) runtime::run_helper: about to invoke pika_main
Hello World!
[2024-04-16 17:27:38.942485701] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180336] [pool:----/0004/----] [parent:----/----] [task:----/----] [runtime.cpp:1189/wait] runtime: about to enter wait state
[2024-04-16 17:27:38.942594025] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180342] [pool:----/----/----] [parent:----/----] [task:----/----] [runtime.cpp:1158/wait_finalize] runtime: about to enter wait state
[2024-04-16 17:27:38.942607966] [pika] [info] [host:pf3dx1(-1)] [pid:2180336] [tid:2180342] [pool:----/----/----] [parent:----/----] [task:----/----] [runtime.cpp:1160/wait_finalize] runtime: exiting wait state
```

- The time and `pika` is always printed first (useful to users who might not know that pika is running under the hood, at least they have something to search for)
- The log level comes next, and is printed in color (see color: https://github.com/gabime/spdlog/wiki/3.-Custom-formatting#pattern-flags)
- The hostname is printed along with the MPI rank (if defined), based on https://github.com/pika-org/pika/blob/adb63f0b9d0b532a5882b529ea3d3e091c6df820/libs/pika/debugging/include/pika/debugging/print.hpp#L280-L287
- The process and thread id are printed with the formatting provided by spdlog (`%P` and `%t` from https://github.com/gabime/spdlog/wiki/3.-Custom-formatting#pattern-flags)
- The pool and worker thread are printed in the format `<pool num>/<global thread num>/<local thread num>`. `----` is printed when any of them is not set.
- The parent and self task ids are printed with the default `thread_id` formatter.
- The filename, line number, and function are always printed to add some context.
- The actual log message is always printed last.

The log format is customizable with `PIKA_LOG_FORMAT` or `--pika:log-format`. The pika information (thread ids, hostname, worker thread information) is implemented using custom spdlog formatters.

The log level follows spdlog's log levels, from 0 (trace) to 6 (off) and can be set with `PIKA_LOG_LEVEL` or `--pika:log-level`.

The destination can be set with `PIKA_LOG_DESTINATION` or `--pika:log-destination`, and defaults to `cerr`. This can be `cerr` or `cout` for stderr and stdout, or a file otherwise.


msimberg:
I'm not asking for detailed reviews at this point (and I would suggest not looking at the diff yet). However, I'd be grateful for comments on the high-level changes:
- configuration options (these will be documented once done)
- style of the log pattern
- logged information (too verbose? can other information be useful?)

I'm not set on the `[x] [y]` pattern for logging. I went with it initially because:
1. it's similar to the default pattern set by spdlog
2. it's relatively easy to separate the different components by regexes etc.
3. everything is on one line

Currently I'm not making sure the logged fields are always the same width. Would that be useful and/or essential to have?

In this new version there is only one logger. Is it useful to bring back different types/categories of loggers? Can we do so without making the user-facing API of enabling logging more complicated?

msimberg:
Overheads need a bit of work:

msimberg:
Performance looks better with https://github.com/pika-org/pika/pull/1093/commits/196666550ac9281d5ab79bb23648b32f936481a3:

msimberg:
In https://github.com/pika-org/pika/pull/1093/commits/f6493a14064c4df2901c1387d9d298010b866266 I've pushed some preliminary documentation about logging.

docs/usage.rst
```diff
@@ -189,6 +189,62 @@ Since version ``0.20.0``, the ``pika-bind`` helper script is bundled with pika.
 and then runs the given command. ``pika-bind`` is a more convenient alternative to manually setting ``PIKA_PROCESS_MASK``
 when pika is used together with a runtime that may reset the process mask of the main thread, like OpenMP.
 
+.. _logging:
+
+Logging
+=======
+
+The pika runtime logs warnings and more severe messages to stderr by default, using `spdlog
```
RMeli, line 197:
```suggestion
The pika runtime logs warnings and more severe messages using `spdlog
```
The default logging location is already described below.

docs/usage.rst
```diff
@@ -189,6 +189,62 @@ Since version ``0.20.0``, the ``pika-bind`` helper script is bundled with pika.
...
+.. code-block::
+
+   [2024-04-18 13:45:07.095279283] [pika] [info] [host:machine(-1)] [pid:2786603] [tid:2786607] [pool:0000/0003/0003] [parent:----/----] [task:0x7fa6a4077cf0/pika_main] [set_thread_state.cpp:205:set_thread_state] set_thread_state: thread(0x7fa6a802c8d0), description(<unknown>), new state(pending), old state(suspended)
+
+The fields are as follows:
+
+- ``[2024-04-18 13:45:07.095279283]``: The timestamp of the message.
+- ``[pika]``: An identifier present in all pika's logs.
+- ``[info]``: The severity level of the message.
+- ``[host:machine(-1)]``: The hostname and the MPI rank of the process (``-1`` if MPI is disabled).
```
RMeli, line 218: Is it too complex/slow to just have `[hostname]` instead of `[hostname(-1)]` when MPI is not used?

docs/usage.rst
```diff
@@ -189,6 +189,62 @@ Since version ``0.20.0``, the ``pika-bind`` helper script is bundled with pika.
...
+- ``[pid:2786603]``: The process id as reported by the operating system.
+- ``[tid:2786607]``: The thread id as reported by the operating system.
+- ``[pool:0000/0003/0003]``: The pika thread pool and worker thread ids: the first component is the
+  thread pool id, the second is the global worker thread id (unique across all thread pools), and
+  the third is the local worker thread id (unique only within the current thread pool). If the
+  message is not logged from a thread on the pika runtime you may see ``[pool:----/----/----]``. You
+  may also see ``[pool:----/0004/----]``. This corresponds to the main thread of the application,
+  which does not belong to a thread pool, but still has a global worker thread id.
```
RMeli, line 226: Some of the entries, like this one, are quite verbose. I'm wondering if it would be beneficial to have simply
```
- ``[pool:0000/0003/0003]``: The pika thread pool and worker thread ids: the first component is the
  thread pool id, the second is the global worker thread id (unique across all thread pools), and
  the third is the local worker thread id (unique only within the current thread pool).
```
And move the details in dedicated per-item subsections. Same thing for some of the entries below.

docs/usage.rst
RMeli: But I see below that `%j` is set to the MPI rank, for custom logging. So an empty string might be surprising to some users.

RMeli (commented):
I feel the default is a bit verbose, but probably it's a good thing. And it should not be an issue since users can easily customize the message.

docs/usage.rst
msimberg: That's a good point. I was worried that the special log entries (like `----/0004/----`) would be too confusing without further explanation so I opted for being more verbose. Is your suggestion to structure it a bit more like this (in pseudo-markdown):
```
# Logging

Intro to logging.

Description of fields:
- list
- of
- short
- descriptions

## Thread pool and worker threads

Longer description of the possible values.

## Task id and description

Longer description of the possible values.
```
? Or perhaps:
```
# Logging

Intro to logging.

Description of fields:
- list
- of
- short
- descriptions

Longer description of the possible values of thread pool and worker thread ids.

Longer description of the possible values of task ids and descriptions.
```
?

Since not all fields need further explanation I thought it makes sense to keep it straightforward with all the information in the list. Seeing the above options I think perhaps the second could be a bit nicer to read. Did you have something else in mind? What do you think?

RMeli: I was suggesting the first option, but both are equivalent IMO. I was just concerned that having very long bullet points might be difficult to read/follow.

msimberg: I've pushed a small restructuring in https://github.com/pika-org/pika/pull/1093/commits/bfd61c516b1f8607969bd45bacde5f186b0d6e8e. Rendered here for convenience:
2024-04-19-141838_792x1226_scrot
Does that look any better to you?

docs/usage.rst
msimberg: I updated the "hostname" field to be `[host:<hostname>/----]` in the case that the MPI rank is unavailable, with the motivation that the output is more predictable in the sense that _something_ is always printed in the MPI rank field, but not the magic number `-1`. This is consistent with how I've treated the other fields so far (task description/id, thread pool ids, etc.), i.e. indicate that the field exists in general, but isn't available at that moment.

@RMeli do you think this is ok, or do you think it's more confusing to have these "empty fields"?

PR #1094, opened 2024-04-17
Install cmake from repositories in ARM CircleCI configuration

msimberg:
Updates the base image to Ubuntu 22.04. This is not yet enough to install fmt from the repositories (22.04 still has fmt 8, whereas we require 9 or newer).


msimberg:
I think we're seeing https://github.com/sphinx-doc/sphinx/issues/12295#issuecomment-2060772469 on the documentation builds now. Let's wait a bit to see if they fix it upstream.

PR #1099, opened 2024-04-19
Minor debug cleanup

biddisco:
part 1 : Simplifies the debug helper code by removing ifdefs and using `__has_include` for c++ name demmangling
part 2 : declutter some of the mpi code for example by moving a namespace check into a macro so it doesn't need to be used in many places.
pushing this PR because it is in my incoming MPI change PR and can be reviewed separately

Note that removing the debug code and replacing it with `fmt::` is a next step.

### DIFF
--- a/libs/pika/debugging/include/pika/debugging/demangle_helper.hpp
+++ b/libs/pika/debugging/include/pika/debugging/demangle_helper.hpp
@@ -16,17 +16,19 @@
 // gcc and clang both provide this header
 #if __has_include(<cxxabi.h>)
 # include <cxxabi.h>
-using cxxabi_supported__ = std::true_type;
+namespace pika::debug::detail {
+    using cxxabi_supported__ = std::true_type;
+    using abi::__cxa_demangle;
+}    // namespace pika::debug::detail
 #else
-using cxxabi_supported__ = std::false_type;
-// create some dummy function to make the compiler happy in the true_type instantiation
-namespace abi {
+namespace pika::debug::detail {
+    using cxxabi_supported__ = std::false_type;
     template <typename... Ts>
     char* __cxa_demangle(Ts... ts)
     {
         return nullptr;
     }
-}    // namespace abi
+}    // namespace pika::debug::detail
 #endif
 
 // --------------------------------------------------------------------
@@ -43,7 +45,7 @@ namespace pika::debug::detail {
     struct demangle_helper<T, std::true_type>
     {
         demangle_helper()
-          : demangled_{abi::__cxa_demangle(typeid(T).name(), nullptr, nullptr, nullptr), std::free}
+          : demangled_{__cxa_demangle(typeid(T).name(), nullptr, nullptr, nullptr), std::free}
         {
         }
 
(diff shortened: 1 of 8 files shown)


libs/pika/debugging/include/pika/debugging/demangle_helper.hpp
```diff
@@ -8,103 +8,76 @@
...
+        std::unique_ptr<char, decltype(&std::free)> demangled_;
     };
-}    // namespace pika::debug::detail
-
-#else
 
-namespace pika::debug::detail {
     template <typename T>
-    using cxxabi_demangle_helper = demangle_helper<T>;
+    using cxx_type_id = demangle_helper<T, cxxabi_supported__>;
```
msimberg, line 56: Is there a good reason not to do something like this:
```
#if defined(PIKA_DEBUG_DETAIL_HAVE_CXXABI) // define this further up
    // if available : demangle an arbitrary c++ type using gnu utility
    template <typename T>
    struct cxx_type_id<T>
    {
        cxx_type_id()
          : demangled_{abi::__cxa_demangle(typeid(T).name(), nullptr, nullptr, nullptr), std::free}
        {
        }

        char const* type_id() const { return demangled_ ? demangled_.get() : typeid(T).name(); }

    private:
        std::unique_ptr<char, decltype(&std::free)> demangled_;
    };
#else
    // default : use built-in typeid to get the best info we can
    template <typename T>
    struct cxx_type_id
    {
        char const* type_id() const { return typeid(T).name(); }
    };
#endif
```

or even
```
    template <typename T>
    struct cxx_type_id<T>
    {
#if defined(PIKA_DEBUG_DETAIL_HAVE_CXXABI) // define this further up  
        // if available : demangle an arbitrary c++ type using gnu utility     
        cxx_type_id()
          : demangled_{abi::__cxa_demangle(typeid(T).name(), nullptr, nullptr, nullptr), std::free}
        {
        }

        char const* type_id() const { return demangled_ ? demangled_.get() : typeid(T).name(); }

    private:
        std::unique_ptr<char, decltype(&std::free)> demangled_;
#else
        // default : use built-in typeid to get the best info we can
        char const* type_id() const { return typeid(T).name(); }
#endif
    };
```
?

These would avoid potential issues with defining `__cxa_demangle`.

libs/pika/debugging/include/pika/debugging/demangle_helper.hpp
```diff
@@ -8,103 +8,76 @@
...
-    // usage : std::cout << print_type<args...>("separator")
-    // separator is appended if the number of types > 1
-    // --------------------------------------------------------------------
-    template <typename T = void>
+// --------------------------------------------------------------------
+// print type information
+// usage : std::cout << debug::print_type<args...>("separator")
+// separator is appended if the number of types > 1
+// --------------------------------------------------------------------
+namespace pika::debug {
```
msimberg, line 64: Is this meant to be user-facing? If not, we should keep it in `pika::debug::detail` (we can shorten it to `pika::detail` as well).

biddisco: yes, I wanted `pika::debug::print_xxx` to be in the pika::debug namespace as these are the user facing apis that people might use.
Just pushing a new version that I think resolves the abi namespace issue cleanly. Don't know why I didn't do it first time around, but that's why we have these annoying reviews ....

biddisco:
Added one clang format patch since I had to reformat my stuff and noticed there's a mismatch in in versions in scripts

msimberg (approved):

PR #1102, opened 2024-04-23
More MPI polling improvements

biddisco:
This PR contains a number of improvements to the mpi polling and continuation handling
1) If a request is not being handled inline, then it must be transferred to the polling pool - but if the polling pool has been disabled, then this extra transfer can be elided. We remove some unwanted transfers
2) If the polling pool is enabled and no other thread can participate in the handling of requests/continuations, then the polling loop can be simplified to remove some locks and queues as they are unnecessary
3) Remove all stream based operations (they will be reinstated in a later PR)
4) Consistently handle high priority boosts for tasks that are transferred/injected to the polling loop or back to the work queues.

### DIFF
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -68,18 +68,13 @@ namespace pika::mpi::experimental::detail {
     // return a scheduler on the mpi pool, with or without stack
     inline auto mpi_pool_scheduler(execution::thread_priority p, bool stack = true)
     {
-        if (stack)
+        ex::thread_pool_scheduler sched{&resource::get_thread_pool(get_pool_name())};
+        if (!stack)
         {
-            return ex::with_priority(
-                ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p);
+            sched = ex::with_stacksize(std::move(sched), execution::thread_stacksize::nostack);
         }
-        else
-        {
-            return ex::with_stacksize(
-                ex::with_priority(
-                    ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p),
-                execution::thread_stacksize::nostack);
-        }
+        sched = ex::with_priority(std::move(sched), p);
+        return sched;
     }
 
     // -----------------------------------------------------------------
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -66,16 +66,18 @@ namespace pika::mpi::experimental::detail {
 
     // -----------------------------------------------------------------
     // return a scheduler on the mpi pool, with or without stack
-    inline auto mpi_pool_scheduler(bool stack = true)
+    inline auto mpi_pool_scheduler(execution::thread_priority p, bool stack = true)
     {
         if (stack)
         {
-            return ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())};
+            return ex::with_priority(
+                ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p);
         }
         else
         {
             return ex::with_stacksize(
-                ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())},
+                ex::with_priority(
+                    ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p),
                 execution::thread_stacksize::nostack);
         }
     }
@@ -84,10 +86,6 @@ namespace pika::mpi::experimental::detail {
     // return a scheduler on the default pool with added priority if requested
     inline auto default_pool_scheduler(execution::thread_priority p)
     {
-        if (p == execution::thread_priority::normal)
-        {
-            return ex::thread_pool_scheduler{&resource::get_thread_pool("default")};
-        }
         return ex::with_priority(
             ex::thread_pool_scheduler{&resource::get_thread_pool("default")}, p);
     }
@@ -168,10 +166,11 @@ namespace pika::mpi::experimental::detail {
     // adds a request callback to the mpi polling code which will call
     // set_value/error on the receiver
     template <typename Receiver>
-    void schedule_task_callback(MPI_Request request, Receiver&& receiver)
+    void
+    schedule_task_callback(MPI_Request request, execution::thread_priority p, Receiver&& receiver)
     {
         detail::add_request_callback(
-            [receiver = PIKA_MOVE(receiver)](int status) mutable {
+            [receiver = PIKA_MOVE(receiver), p](int status) mutable {
                 PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("schedule_task_callback")));
                 if (status != MPI_SUCCESS)
                 {
@@ -181,8 +180,7 @@ namespace pika::mpi::experimental::detail {
                 else
                 {
                     // pass the result onto a new task and invoke the continuation
-                    auto snd0 = ex::just(status) |
-                        ex::transfer(default_pool_scheduler(execution::thread_priority::high)) |
+                    auto snd0 = ex::just(status) | ex::transfer(default_pool_scheduler(p)) |
                         ex::then([receiver = PIKA_MOVE(receiver)](int status) mutable {
                             PIKA_DETAIL_DP(
                                 mpi_tran<5>, debug(str<>("set_value_error_helper"), status));
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -145,7 +145,7 @@ namespace pika::mpi::experimental::detail {
     template <typename OperationState>
     void resume_request_callback(MPI_Request request, OperationState& op_state)
     {
-        op_state.completed = false;
+        PIKA_ASSERT(op_state.completed == false);
         detail::add_request_callback(
             [&op_state](int status) mutable {
                 PIKA_DETAIL_DP(mpi_tran<5>,
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -142,10 +142,6 @@ namespace pika::mpi::experimental {
             ///
             /// * continuation : the polling thread will call the continuation directly
             ///
-            /// * mpix_continuation : the continuation of the mpi_request is passed into the mpi
-            /// library using mpix extensions and is triggered by the polling thread directly
-            /// by the mpi library (during MPI_Test)
-            ///
             /// * unspecified : reserved for development purposes or for customization by an
             /// application using pika
             yield_while = 0x00,       // 00 ... 15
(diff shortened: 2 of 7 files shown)


libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -67,16 +66,18 @@ namespace pika::mpi::experimental::detail {
...
+                ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p);
         }
         else
         {
             return ex::with_stacksize(
-                ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())},
+                ex::with_priority(
+                    ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p),
                 execution::thread_stacksize::nostack);
         }
```
msimberg, line 82: Optional refactoring suggestion:
```
        ex::thread_pool_scheduler sched{&resource::get_thread_pool(get_pool_name()};
        if (!stack) {
            sched = ex::with_stacksize(std::move(sched), execution::thread_stacksize::nostack);
        }
        sched = ex::with_priority(std::move(sched), p);
        return sched;
```
or a some similar variation?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -64,40 +65,41 @@ namespace pika::mpi::experimental {
 
         const char* stream_name(stream_type s);
 
+        // -----------------------------------------------------------------
+        /// Adds an mpi request to the internal queues for polling/handling
         PIKA_EXPORT bool add_request_callback(request_callback_function_type&&, MPI_Request);
-        PIKA_EXPORT void register_polling(pika::threads::detail::thread_pool_base&);
-        PIKA_EXPORT void unregister_polling(pika::threads::detail::thread_pool_base&);
```
msimberg, line 69: This seems to have disappeared from the header. Just checking: is this because it's no longer needed elsewhere, only internally?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -115,57 +117,81 @@ namespace pika::mpi::experimental {
...
-            /// 2 bits control the handler method,
-            method_mask = 0x30,
-
-            /// the individual methods that are supported for dispatching continuations
-            yield_while = 0x00,
-            suspend_resume = 0x10,
-            new_task = 0x20,
-            continuation = 0x30,
+            /// 3 bits control the handler method,
+            method_mask = 0x70,
```
msimberg, line 128: Optional refactoring: use binary literals instead, which in my opinion would be easier to read:
```suggestion
            method_mask = 0b0111'0000,
```
Likewise for the other flags.

However, related to this I think at the very least for the user this bitmask interface is terrible. I think separating the different options into separate enums in the environment variables at minimum would be good, if not internally. Is the optimal packing of these flags important for performance? This is not a blocker, and I'd be happy to do this refactoring separately.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -115,57 +117,81 @@ namespace pika::mpi::experimental {
...
+            continuation = 0x30,      // 48 ... 63
+            unspecified = 0x40,       // 64
 
             /// Default flags are to invoke inline, but transfer completion using a dedicated pool
             default_mode = use_pool | request_inline | high_priority | new_task,
         };
 
-        // 2 bits define continuation mode
+        /// 2 bits define continuation mode
         inline handler_mode get_handler_mode(int flags)
```
msimberg, line 162:
```suggestion
        inline handler_mode get_handler_mode(std::underlying_type_t<handler_mode> flags)
```

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -167,7 +166,9 @@ namespace pika::mpi::experimental {
 
         struct initializer
         {
-            initializer() { init_throttling_default(); }
+            initializer()
+            { /*init_throttling_default();*/
```
msimberg, line 170: What's this change about?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -584,6 +587,9 @@ namespace pika::mpi::experimental {
                 polling_status::busy;
         }
 
+        // -------------------------------------------------------------
+        pika::threads::detail::polling_status poll() { return poll_multithreaded(); }
```
msimberg, line 591: Is this in preparation for having a `poll_singlethreaded` or similar variant? Or is this is a leftover change that can be reverted?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -742,85 +748,87 @@ namespace pika::mpi::experimental {
...
-            detail::register_polling(pika::resource::get_thread_pool(name));
-        }
-
         static bool defaults_set = false;
         if (!defaults_set)
         {
-            detail::init_throttling_default();
-            defaults_set = true;
+            // detail::init_throttling_default();
+            // defaults_set = true;
```
msimberg, line 825: What's this about?

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -279,7 +276,7 @@ struct message_receiver
                 // the recursive lambda will handle it
                 auto rx_snd2 = ex::just(buf, message_size, MPI_UNSIGNED_CHAR, prev_rank(rank, size),
                                    tag, MPI_COMM_WORLD) |
-                    mpi::transform_mpi(MPI_Irecv, mpi::stream_type::receive_2) |
+                    mpix::transform_mpi(MPI_Irecv /*, mpix::stream_type::receive_2*/) |
```
msimberg, line 279: Can the `stream_type`s be uncommented?

msimberg (commented):
Initial review. I'll look closer once #1099 is merged and the diff is a bit easier to look at.

msimberg:
@biddisco would you mind summarizing in the PR description what has been improved and changed?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -41,6 +41,7 @@ namespace pika::mpi::experimental {
     enum pool_create_mode
     {
         pika_decides = 0,
+        pika_decides_unoptimized,
```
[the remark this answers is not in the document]
biddisco: Aha. this flag is to disable other optimizations (like using MPI_INIT THREAD_SINGLE)  but they are not in the PR. I will remove that option until it is needed (I've seen a few others like that too, that I will remove)

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: correct. The user no longer explicitly creates an mpi pool, instead, during init, you call mpi::pool_create and it sets everything up, there is still a register polling function, but it moved and does not take a pool param since the mpi code knows which pool to use.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: I was actually in the process of changing to a binary bitmask, so I'm in agreement here. I also agree that other flags would be useful, but I'd rather do that in another PR. It's just too convenient to have an integer in testing loops to scan over all of them at the moment.

libs/pika/async_mpi/src/mpi_polling.cpp
biddisco: All the stream based throttling code is  essentially still in there, but I have not used it as it was causing too many lockups. I was convinced there was a bug in condition_variable somewhere and so all throttling has been disabled, but I do want to try using it again so the major part of the code is in place but parts commented out and api calls (as you saw in another comment) still have params that are commented out.
Option 1 - remove it all, then put it back in in another PS
Option 2 - allow some messy commented out bits in places

libs/pika/async_mpi/src/mpi_polling.cpp
biddisco: same as above. currently unused throttling code

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
biddisco: same as above. stream throttling disabled, so commented out.

Do I remove all throttling bits and put it back in a cleaned up separate PR?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
msimberg: Understandable. Do you want to at least do the hex to binary literal change still in this PR, and we can discuss how best to deal with the flags separately?

libs/pika/async_mpi/src/mpi_polling.cpp
msimberg: All right, let's keep the commented out stuff for now then.

msimberg: https://github.com/pika-org/pika/pull/1106

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -744,77 +710,81 @@ namespace pika::mpi::experimental {
...
-                    error(str<>("pika::mpi::experimental::init"), "init failed"));
-                PIKA_THROW_EXCEPTION(pika::error::invalid_status, "pika::mpi::experimental::init",
-                    "the MPI installation doesn't allow multiple threads");
+                // when a dedicated pool is used (1 thread), and all requests are transferred to the pool, we can use MPI_THREAD_SINGLE
+                // std::int32_t ranks = pika::detail::get_env_var_as<std::int32_t>("OMPI_WORLD_SIZE", -1);
+                // if the mpi thread pool is disabled, then we can't use single threaded
+                if (pool_mode == pool_create_mode::pika_decides)
+                {
+                    // required = MPI_THREAD_SINGLE;
+                    // PIKA_DETAIL_DP(detail::mpi_debug<0>, debug(str<>("MPI_THREAD_SINGLE")));
```
msimberg, line 753: Just checking: how much of this is intentionally commented out?

msimberg: See https://app.circleci.com/pipelines/github/pika-org/pika/10746/workflows/fd4bcefe-047c-4d49-9cd8-ee810cd66322/jobs/110842?invite=true#step-102-21.

biddisco: I should have deleted that section. It is harmless left in, but it is part of future optimizations to switch to init thread SINGLE if possible.

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -457,22 +452,14 @@ int pika_main(pika::program_options::variables_map& vm)
...
-        pika::mpi::experimental::detail::get_completion_mode_default();
-    }
-    //
+    // Don't create an MPI pool if the user disabled it
+    auto pool_mode = mpix::pool_create_mode::pika_decides;
+    if (vm["no-mpi-pool"].as<bool>()) { pool_mode = mpix::pool_create_mode::force_no_create; }
+
     msr_deb<2>.debug(str<>("init RP"), "create_pool");
-    mpi::create_pool(rp, "", mpi::pool_create_mode::pika_decides);
+    mpix::create_pool("", mpix::pool_create_mode::pika_decides);
```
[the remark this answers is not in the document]
biddisco: oops

msimberg (approved):

PR #1104, opened 2024-04-24
Move `to_underlying` helper to dedicated header in `type_support`

msimberg:


msimberg:
@biddisco this might conflict with your changes in #1102. This can go in after #1102.

PR #1105, opened 2024-04-24
Fix builds with newest stdexec

msimberg:
Opts in to the newer stdexec sender/receiver concepts (with `using sender_concept = stdexec::sender_t` etc.). Also fixes a number of compilation failures resulting from stricter checks etc. in stdexec.

### DIFF
--- a/.gitlab/includes/nvhpc23_9_pipeline.yml
+++ b/.gitlab/includes/nvhpc23_9_pipeline.yml
@@ -17,7 +17,7 @@ include:
     GPU_TARGET: '60'
     SPACK_SPEC: "pika@main arch=$SPACK_ARCH %${NVHPC_COMPILER} +stdexec +mpi +cuda cuda_arch=${GPU_TARGET} \
                  malloc=system cxxstd=$CXXSTD ^boost@1.78.0%${NVHPC_COMPILER} ^hwloc@2.7.0 \
-                 ^stdexec@git.b39a506096900de4074ee556867a3805bd8351bd=main ^cuda@11.8%${NVHPC_COMPILER} \
+                 ^stdexec@git.b39a506096900de4074ee556867a3805bd8351bd=main ^cuda@12.4%${NVHPC_COMPILER} \
                  ^whip%${NVHPC_COMPILER}"
     CMAKE_FLAGS: "-DPIKA_WITH_CXX_STANDARD=$CXXSTD -DPIKA_WITH_CUDA=ON -DPIKA_WITH_MALLOC=system \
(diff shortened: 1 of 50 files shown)


libs/pika/execution/tests/unit/algorithm_let_error.cpp
```diff
@@ -229,7 +229,7 @@ int main()
         PIKA_TEST(!let_error_callback_called);
     }
 
-    test_adl_isolation(ex::let_error(ex::just(), my_namespace::my_type{}));
+    test_adl_isolation(ex::let_error(ex::just_error(42), my_namespace::my_let_error_type{}));
```
msimberg, line 232: To do: we don't have `just_error` and `just_done` in pika's `std::execution` implementation.

msimberg: I updated this to use our internal `error_sender` intended for testing. I'll hold off adding `just_done` and `just_error` as an incentive to prefer `PIKA_WITH_STDEXEC=ON`.

.gitlab/includes/nvhpc23_9_pipeline.yml
```diff
@@ -12,12 +12,12 @@ include:
   variables:
     SPACK_ARCH: linux-ubuntu22.04-haswell
     COMPILER: gcc@12.1.0
-    NVHPC_COMPILER: nvhpc@23.9
+    NVHPC_COMPILER: nvhpc@24.3
```
msimberg, line 15: Todo: if we keep the version bump, change the pipeline name. Currently the version bump doesn't change anything (i.e. does not get rid of build failures), so I might simply revert this.

msimberg: NVHPC version changed back as it doesn't make a difference. I have bumped the CUDA version used in the NVHPC CI configuration to 12.4 though. It seems like NVHPC with a standalone CUDA version older than 12 fails to compile stdexec, and it seems related to how host/device code is detected. Specifically, https://github.com/NVIDIA/stdexec/blob/ab774d5f84ea4d4e9253e06e82b45ed4b8a8e016/include/stdexec/execution.hpp#L462 ends up expanding to incliding a `__trap` in host code (it's only available on the device) and compilation fails. The function in stdexec that contains `STDEXEC_TERMINATE` used to be a template function that doesn't get instantiatiated in regular use (it's only meant for debugging) so this never came up earlier. Now it's a non-template member function that fails to compile right away.

Since this isn't an issue with newer versions of CUDA I'm not going to make any special workarounds, besides bumping the CUDA version in CI. I'm still trying to make a small reproducer and will open an issue if I get one.

PR #1128, opened 2024-05-27
Add support for MPI continuations

biddisco:
Some cleanups to the main mpi polling routines and API - quite a lot of redundant use removed and some minor api changes to make things clearer.

And ...  support for mpi-continuations using the openmpi extension. 

we check for the extension using
```
#if __has_include(<mpi-ext.h>)
# include <mpi-ext.h>
# ifndef OMPI_HAVE_MPI_EXT_CONTINUE
#  pragma warning "MPIX Continuations not found in <mpi-ext.h>"
# endif
``` 
so code is only enabled if `OMPI_HAVE_MPI_EXT_CONTINUE` is found in `<mpi-ext.h>` 

Some optimizations are added to handle single/multiple threaded polling

### DIFF
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -163,16 +163,18 @@ namespace pika::mpi::experimental::detail {
     // handler_method::mpix_continuation
     /// typedef int (MPIX_Continue_cb_function)(int rc, void *cb_data);
     template <typename OperationState>
-    static int mpix_callback([[maybe_unused]] int rc, void* cb_data)
+    int mpix_callback([[maybe_unused]] int rc, void* cb_data)
     {
         PIKA_DETAIL_DP(mpi_tran<1>, debug(str<>("MPIX"), "callback triggered"));
         auto& op_state = *static_cast<OperationState*>(cb_data);
         // wake up the suspended thread
         {
             std::lock_guard lk(op_state.mutex);
+            op_state.status = rc;
             op_state.completed = true;
         }
         op_state.cond_var.notify_one();
+        // tell mpix that we handled it ok, error is passed into set_error in mpi_trigger
         return MPI_SUCCESS;
     }
 
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -35,7 +35,7 @@ namespace pika::mpi::experimental::detail {
     // -----------------------------------------------------------------
     // by convention the title is 7 chars (for alignment)
     template <int Level>
-    inline constexpr debug::detail::print_threshold<Level, 0> mpi_tran("MPITRAN");
+    inline constexpr debug::detail::print_threshold<Level, 9> mpi_tran("MPITRAN");
 
     namespace ex = pika::execution::experimental;
 
@@ -155,6 +155,23 @@ namespace pika::mpi::experimental::detail {
                 op_state.cond_var.notify_one();
             },
             request);
+    }
+
+    // -----------------------------------------------------------------
+    /// typedef int (MPIX_Continue_cb_function)(int rc, void *cb_data);
+    template <typename OperationState>
+    static int mpix_callback([[maybe_unused]] int rc, void* cb_data)
+    {
+        PIKA_DETAIL_DP(mpi_tran<1>, debug(str<>("MPIX"), "callback triggered"));
+        auto& op_state = *static_cast<OperationState*>(cb_data);
+        // wake up the suspended thread
+        {
+            std::lock_guard lk(op_state.mutex);
+            op_state.completed = true;
+            detail::complete_mpix();
+        }
+        op_state.cond_var.notify_one();
+        return MPI_SUCCESS;
     }
 
     // -----------------------------------------------------------------
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -35,7 +35,7 @@ namespace pika::mpi::experimental::detail {
     // -----------------------------------------------------------------
     // by convention the title is 7 chars (for alignment)
     template <int Level>
-    inline constexpr debug::detail::print_threshold<Level, 9> mpi_tran("MPITRAN");
+    inline constexpr debug::detail::print_threshold<Level, 0> mpi_tran("MPITRAN");
 
     namespace ex = pika::execution::experimental;
 
@@ -168,7 +168,6 @@ namespace pika::mpi::experimental::detail {
         {
             std::lock_guard lk(op_state.mutex);
             op_state.completed = true;
-            detail::complete_mpix();
         }
         op_state.cond_var.notify_one();
         return MPI_SUCCESS;
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -134,9 +134,9 @@ namespace pika::mpi::experimental::detail {
                 {
                     // pass the result onto a new task and invoke the continuation
                     auto snd0 = ex::transfer_just(default_pool_scheduler(p)) |
-                        ex::then([receiver = PIKA_MOVE(receiver)]() mutable {
+                        ex::then([receiver = PIKA_FORWARD(Receiver, receiver)]() mutable {
                             PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("set_value")));
-                            ex::set_value(PIKA_FORWARD(Receiver, receiver));
+                            ex::set_value(PIKA_MOVE(receiver));
                         });
                     ex::start_detached(PIKA_MOVE(snd0));
                 }
@@ -160,7 +160,7 @@ namespace pika::mpi::experimental::detail {
     }
 
     // -----------------------------------------------------------------
-    // handler_method::mpix_continuation
+    // handler_method::mpix_continuation - signature is
     /// typedef int (MPIX_Continue_cb_function)(int rc, void *cb_data);
     template <typename OperationState>
     int mpix_callback([[maybe_unused]] int rc, void* cb_data)
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -18,6 +18,7 @@
 #include <pika/execution/algorithms/just.hpp>
 #include <pika/execution/algorithms/then.hpp>
 #include <pika/execution/algorithms/transfer.hpp>
+#include <pika/execution/algorithms/transfer_just.hpp>
 #include <pika/execution_base/any_sender.hpp>
 #include <pika/execution_base/receiver.hpp>
 #include <pika/execution_base/sender.hpp>
@@ -37,20 +38,8 @@ namespace pika::mpi::experimental::detail {
     template <int Level>
     inline constexpr debug::detail::print_threshold<Level, 0> mpi_tran("MPITRAN");
 
-    namespace ex = pika::execution::experimental;
-
     // -----------------------------------------------------------------
-    template <typename T>
-    struct any_sender_helper
-    {
-        using type = ex::unique_any_sender<T>;
-    };
-
-    template <>
-    struct any_sender_helper<void>
-    {
-        using type = ex::unique_any_sender<>;
-    };
+    namespace ex = pika::execution::experimental;
 
     // -----------------------------------------------------------------
     // is func(Ts..., MPI_Request) invocable
@@ -103,49 +92,17 @@ namespace pika::mpi::experimental::detail {
     }
 
     // -----------------------------------------------------------------
-    // adds a request callback to the mpi polling code which will call
-    // the set_value/set_error helper using the void return signature
-    template <typename OperationState>
-    void set_value_request_callback_void(MPI_Request request, OperationState& op_state)
-    {
-        detail::add_request_callback(
-            [&op_state](int status) mutable {
-                PIKA_DETAIL_DP(mpi_tran<5>,
-                    debug(str<>(
-                        "callback_void") /*, "stream", detail::stream_name(op_state.stream)*/));
-                set_value_error_helper(status, PIKA_MOVE(op_state.receiver));
-            },
-            request);
-    }
-
-    // -----------------------------------------------------------------
-    // adds a request callback to the mpi polling code which will call
-    // the set_value/set_error helper with a valid return value
-    template <typename Result, typename OperationState>
-    void set_value_request_callback_non_void(MPI_Request request, OperationState& op_state)
-    {
-        detail::add_request_callback(
-            [&op_state](int status) mutable {
-                PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("callback_nonvoid")));
-                PIKA_ASSERT(std::holds_alternative<Result>(op_state.result));
-                set_value_error_helper(status, PIKA_MOVE(op_state.receiver),
-                    PIKA_MOVE(std::get<Result>(op_state.result)));
-            },
-            request);
-    }
-
-    // -----------------------------------------------------------------
+    // handler_method::suspend_resume
     // adds a request callback to the mpi polling code which will call
     // notify_one to wake up a suspended task
     template <typename OperationState>
-    void resume_request_callback(MPI_Request request, OperationState& op_state)
+    void add_suspend_resume_request_callback(MPI_Request request, OperationState& op_state)
     {
         PIKA_ASSERT(op_state.completed == false);
         detail::add_request_callback(
             [&op_state](int status) mutable {
-                PIKA_DETAIL_DP(mpi_tran<5>,
-                    debug(str<>("callback_void_suspend_resume"), "status", status
-                        /*, "stream", detail::stream_name(op_state.stream)*/));
+                PIKA_DETAIL_DP(
+                    mpi_tran<5>, debug(str<>("callback_void_suspend_resume"), "status", status));
                 // wake up the suspended thread
                 {
                     std::lock_guard lk(op_state.mutex);
@@ -158,6 +115,52 @@ namespace pika::mpi::experimental::detail {
     }
 
     // -----------------------------------------------------------------
+    // handler_method::new_task
+    // adds a request callback to the mpi polling code which will call
+    // set_value/error on the receiver
+    template <typename Receiver>
+    void add_new_task_request_callback(
+        MPI_Request request, execution::thread_priority p, Receiver&& receiver)
+    {
+        detail::add_request_callback(
+            [receiver = PIKA_MOVE(receiver), p](int status) mutable {
+                PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("schedule_task_callback")));
+                if (status != MPI_SUCCESS)
+                {
+                    ex::set_error(PIKA_FORWARD(Receiver, receiver),
+                        std::make_exception_ptr(mpi_exception(status)));
+                }
+                else
+                {
+                    // pass the result onto a new task and invoke the continuation
+                    auto snd0 = ex::transfer_just(default_pool_scheduler(p)) |
+                        ex::then([receiver = PIKA_MOVE(receiver)]() mutable {
+                            PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("set_value")));
+                            ex::set_value(PIKA_FORWARD(Receiver, receiver));
+                        });
+                    ex::start_detached(PIKA_MOVE(snd0));
+                }
+            },
+            request);
+    }
+
+    // -----------------------------------------------------------------
+    // handler_method::continuation
+    // adds a request callback to the mpi polling code which will call
+    // the set_value/set_error helper using the void return signature
+    template <typename OperationState>
+    void add_continuation_request_callback(MPI_Request request, OperationState& op_state)
+    {
+        detail::add_request_callback(
+            [&op_state](int status) mutable {
+                PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("callback_void")));
+                set_value_error_helper(status, PIKA_MOVE(op_state.receiver));
+            },
+            request);
+    }
+
+    // -----------------------------------------------------------------
+    // handler_method::mpix_continuation
     /// typedef int (MPIX_Continue_cb_function)(int rc, void *cb_data);
     template <typename OperationState>
     static int mpix_callback([[maybe_unused]] int rc, void* cb_data)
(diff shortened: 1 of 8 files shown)


libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -175,16 +133,47 @@ namespace pika::mpi::experimental::detail {
...
+                set_value_error_helper(status, PIKA_MOVE(op_state.receiver));
+            },
+            request);
+    }
+
+    // -----------------------------------------------------------------
+    // handler_method::mpix_continuation
+    /// typedef int (MPIX_Continue_cb_function)(int rc, void *cb_data);
+    template <typename OperationState>
+    static int mpix_callback([[maybe_unused]] int rc, void* cb_data)
```
msimberg, line 166:
```suggestion
    int mpix_callback([[maybe_unused]] int rc, void* cb_data)
```

msimberg, line 166: Can or should we do something with `rc`? I assume it's an error code, so if it's not "success" we should likely propagate that to receivers?

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -78,48 +76,28 @@ namespace pika::mpi::experimental {
             execution::thread_priority p = use_priority_boost(mode) ?
                 execution::thread_priority::boost :
                 execution::thread_priority::normal;
-            if (inline_req || !pool_exists())
+
+            auto completion_snd = [=](MPI_Request request) -> unique_any_sender<> {
+                if (!inline_com)
+                {
+                    if (request == MPI_REQUEST_NULL)
+                        return transfer_just(default_pool_scheduler(p));
```
msimberg, line 84:
```suggestion
                    {
                        return transfer_just(default_pool_scheduler(p));
                    }
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -78,48 +76,28 @@ namespace pika::mpi::experimental {
...
+                    return transfer_just(default_pool_scheduler(p), request) | trigger_mpi(mode);
+                }
+                if (request == MPI_REQUEST_NULL) return just();
```
msimberg, line 87:
```suggestion
                if (request == MPI_REQUEST_NULL) { return just() };
```

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -85,6 +85,8 @@ namespace pika::mpi::experimental::detail {
             bool completed{false};
             pika::detail::spinlock mutex;
             pika::condition_variable cond_var;
+            // MPIX
```
msimberg, line 88: General question: should we refer to this as "MPI continuations" rather than "MPIX"? If I understand correctly, MPIX generally refers to MPI "extensions", right? By this I include variable names/typedefs etc. as well.

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -186,6 +177,46 @@ namespace pika::mpi::experimental::detail {
...
+                                    r.op_state.cond_var.wait(
+                                        l, [&]() { return r.op_state.completed; });
+                                }
+
+                                PIKA_DETAIL_DP(mpi_tran<1>,
+                                    debug(str<>("MPIX"), "woken", ptr(r.op_state.request)));
+
+                                // call set_value/set_error depending on mpi return status
+                                set_value_error_helper(
+                                    /*r.op_state.status*/ MPI_SUCCESS,
```
msimberg, line 216: Always success?

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -186,6 +177,46 @@ namespace pika::mpi::experimental::detail {
...
+                                    debug(str<>("MPIX"), "register_mpix_continuation",
+                                        ptr(r.op_state.request), ptr(r.op_state.request)));
+
+                                MPIX_Continue_cb_function* func = &mpix_callback<operation_state>;
+                                register_mpix_continuation(&r.op_state.request, func, &r.op_state);
+                                {
+                                    PIKA_DETAIL_DP(mpi_tran<1>,
+                                        debug(str<>("MPIX"), "waiting", ptr(r.op_state.request)));
+                                    std::unique_lock l{r.op_state.mutex};
+                                    r.op_state.cond_var.wait(
```
msimberg, line 207: Do I understand it correctly that with MPI continuations this will only ever use the "suspend_resume" method, i.e. waiting on a condition variable? Is there an option to let MPI continuations directly call `set_value/error` instead of notifying a condition variable in the continuation.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -38,8 +38,22 @@
 #include <utility>
 #include <vector>
 
-namespace pika::mpi::experimental {
+#if __has_include(<mpi-ext.h>)
+# include <mpi-ext.h>
+# ifndef OMPI_HAVE_MPI_EXT_CONTINUE
+#  pragma warning "MPIX Continuations not found in <mpi-ext.h>"
```
msimberg, line 44: Can we make this only a message or even not print it? I wouldn't want a `-Werror` build to error out on this not being available.

Alternatively, make `PIKA_WITH_MPI_CONTINUATIONS` an explicit CMake option, and the build will fail MPI continuations aren't available?

In either case, we could probably remove some of the members if continuations are disabled?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -38,8 +38,22 @@
...
+# endif
+
+# define MPIX_RESULT_CHECK(func)                                                                   \
```
msimberg, line 47:
```suggestion
# define PIKA_MPI_CONTINUATION_RESULT_CHECK(func)                                                                   \
```

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -38,8 +38,22 @@
...
+     {                                                                                             \
+         int code = func;                                                                          \
+         if (code != MPI_SUCCESS)                                                                  \
+             throw pika::mpi::experimental::mpi_exception(code, "MPIX problem");                   \
+     }    //else { std::cout << "MPIX Success" << std::endl; }
```
msimberg, line 52:
```suggestion
     }
```

libs/pika/async_mpi/src/mpi_polling.cpp
msimberg: Actually, does this need to be a macro? `static inline` function might make more sense?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -573,11 +635,20 @@ namespace pika::mpi::experimental {
         // -------------------------------------------------------------
         pika::threads::detail::polling_status poll()
         {
-            if (detail::use_pool(detail::task_completion_flags_) &&
-                !detail::use_inline_request(detail::task_completion_flags_))
+            // get mpi completion mode settings
+            auto mode = get_completion_mode();
+            bool singlethreaded =
+                use_pool(mode) && !use_inline_request(mode) /*&& use_inline_completion(mode)*/;
```
msimberg, line 641: Is the commented-out code needed?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -622,6 +693,32 @@ namespace pika::mpi::experimental {
 
         int comm_world_size() { return detail::mpi_data_.size_; }
 
+#ifdef OMPI_HAVE_MPI_EXT_CONTINUE
+        void register_mpix_continuation(
+            MPI_Request* request, MPIX_Continue_cb_function* cb_func, void* op_state)
+        {
+            // @TODO we don't strictly need defer complete flag, but for debugging ...
```
msimberg, line 700: `MPIX_CONT_DEFER_COMPLETE` is already disabled below, remove comment (and commented out flag)?

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -186,6 +177,46 @@ namespace pika::mpi::experimental::detail {
...
+                                break;
+                            }
+                            case handler_method::continuation:
+                            {
+                                // The callback will call set_value/set_error
+                                // execution will continue on the callback thread
+                                add_continuation_request_callback<>(r.op_state.request, r.op_state);
+                                break;
+                            }
+                            case handler_method::mpix_continuation:
```
msimberg, line 195: What happens if MPI continuations are not enabled, but one requests MPI continuations?

How much of the MPI continuations stuff can we ifdef out if they're not enabled/available?

msimberg (changes requested):

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
biddisco: I removed the brackets because I was pasting code into the paper and trying to reduce space. I'll put them back

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
biddisco: I don't really mind what we call it, they are mpi continuations in the mpi extensions header, so I thought mpix_continuations was a good name. in the header they use `OMPI_HAVE_MPI_EXT_CONTINUE` so mpi_ext_continuations? The functions are `MPIX_Continue_cb_function` etc, so MPIX seemed like a goood prefix for me to use for now. I will replace some commmenrtrs and debug stuff with `MPI_EXT_CONTINUE`

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
biddisco: changed to use the rc return from the callback

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
biddisco: We could add that, but how can we stop people from putting big code in continuations that will run on the polling thread. the callback is invoked from inside the mpi library and may leave locks held or other stuff. We can add support for it, but it'd be better in another PR when we are more familiar with the extensions etc.

libs/pika/async_mpi/src/mpi_polling.cpp
biddisco: I use this just as a sanity check when building on various machines in case the openmpi is somehow wring. it is useful to see the message, but I can add a cmake option.

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
biddisco: at the moment, it's an error. the code will break, but that's why I put the `#pragma warning` in the other place. it's experimental and I didn't try to make it safe yet.

biddisco:
RE. cmake flag for mpi_ext_continuations 
If I add a cmake flag for it and then compile against an mpi that doesn't have it, I have to turn off the flag, then when I want to use it, turn it back on (I'm frequently switching between mpi libs etc), etc etc. I'd rather just have pika compile and if the extension is available, it is enabled, if not, it isn't. so instead of a cmake option, I will add  a check in `pika::mpi::init` that throws an exception if you select an mpi_ext completion mode, but it isn't available.
Comments?

biddisco:
I think I addressed all the comments on this. Did I miss anything?

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -175,16 +133,49 @@ namespace pika::mpi::experimental::detail {
...
+    {
+        PIKA_DETAIL_DP(mpi_tran<1>, debug(str<>("MPIX"), "callback triggered"));
+        auto& op_state = *static_cast<OperationState*>(cb_data);
+        // wake up the suspended thread
+        {
+            std::lock_guard lk(op_state.mutex);
+            op_state.status = rc;
+            op_state.completed = true;
+        }
+        op_state.cond_var.notify_one();
```
msimberg, line 176: I suspect that we need to protect this with the lock as well but details of lifetimes may mean that it's not necessary, please ignore if you think this is fine. The case I'm thinking about is this:

|  thread 1 | thread 2 |
| --- | --- |
| register continuation | |
| | continuation ready |
| | `mpix_callback` |
| | take lock |
| | set completed = true |
| | release lock |
| take lock | |
| wait with predicate | |
| completed is true, continue | |
| ??? | |
| release condition variable | |
| | notify condition variable |
| | :rotating_light: |

Is there anything stopping this from happening? This is obviously very unlikely but not impossible, and will happen sooner or later.

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -175,16 +133,49 @@ namespace pika::mpi::experimental::detail {
...
                     // pass the result onto a new task and invoke the continuation
-                    auto snd0 = ex::just(status) | ex::transfer(default_pool_scheduler(p)) |
-                        ex::then([receiver = PIKA_MOVE(receiver)](int status) mutable {
-                            PIKA_DETAIL_DP(
-                                mpi_tran<5>, debug(str<>("set_value_error_helper"), status));
-                            set_value_error_helper(status, PIKA_MOVE(receiver));
+                    auto snd0 = ex::transfer_just(default_pool_scheduler(p)) |
+                        ex::then([receiver = PIKA_MOVE(receiver)]() mutable {
+                            PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("set_value")));
+                            ex::set_value(PIKA_FORWARD(Receiver, receiver));
```
msimberg, line 139: Sorry, missed this earlier. The forward and move are in the wrong places:
```suggestion
                    auto snd0 = ex::transfer_just(default_pool_scheduler(p)) |
                        ex::then([receiver = PIKA_FORWARD(Receiver, receiver)]() mutable {
                            PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("set_value")));
                            ex::set_value(PIKA_MOVE(receiver));
```

msimberg (changes requested):
Two more comments, otherwise this looks ok to be merged.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -112,68 +112,69 @@ namespace pika::mpi::experimental {
...
+            case handler_method::unspecified: return "unspecified";
             default: return "invalid";
             }
         }
 
         /// utility : needed by static checks when debugging
         PIKA_EXPORT int comm_world_size();
 
+        /// mpix extensions in openmpi to support mpi continuations
+        typedef int(MPIX_Continue_cb_function)(int rc, void* cb_data);
```
msimberg, line 173: https://app.circleci.com/pipelines/github/pika-org/pika/10942/workflows/465a087b-682d-47b5-ae9e-f777dd1a2832/jobs/114444?invite=true#step-102-13

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
biddisco: currently this can't happen because I've used `MPIX_CONT_POLL_ONLY` which means that the continuation can only be called when we call `MPI_Test` and since that is protected by a lock in the multithreaded polling loop, and called on the thread that is setting the continuation in the single-threaded version, the continuatiob cannot be called until after we have set everything up correctly.
At some point I will upgrade the `mpix_continuation` version to allow the non `POLL_ONLY` callbacks, and also consider adding support for executing the sender continuation directly rather than using a suspend/resume approach. But to add this in, I will need a new PR

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
biddisco: The receiver needs to be moved into the lambda, I'm not sure forward is right there, but it should be moved into set_value as well. 2 x `PIKA_MOVE` I think

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: ok, fixed

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
msimberg: Your last change looks good to me. The first is a forward because `Receiver` is a template parameter of the with `Receiver&&` in the signature.

biddisco: I'm not sure I understand "Your last change looks good to me" contradicts "The first is a forward because Receiver is a template parameter of the with Receiver&& in the signature."
What are you trying to say? 
What I meant to say was yes - it is a template param that uses `&&`, but regardless of how it is passed into the callback registration function, it should be `moved` into the callback lambda, so I prefer move to forward - because then it really is moved. (probably the template should be changed to only accept a move....but it isn't worth the effort to be fair)

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
msimberg: Thanks, I see how this might not happen in the single-threaded version, but if there's no dedicated MPI pool and/or the MPI request is created inline (not on the MPI pool) then I don't see how the locks help. The lock in the polling loop is unrelated to the condition variable lock, no?

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
msimberg: What I meant is that your last change is what it should look like, followed by the reason why it should look like that, but you know that already. Hence I marked this as resolved.

Yes, in general we can start taking receivers by value, they have to be moved. But we can do that separately across all receivers, so let's leave it like this here.

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
biddisco: condition_variable.wait checks the predicate 
```
            while (!pred()) { wait(lock); }
```
before actually waiting, but in  mpi_helper, the `completed = true` is done under the lock, so if thereread A sets the callback, thread B triggers the callback, then when thread A tries to wait - it will already have completed = true and it will skip the wait and we won't miss a `notify_one`
Am I mistaken?  (but you are correct, my initial explanation was incorreect they are not the same locks)

msimberg: The issue is not about missing a notify. Indeed the `wait` correctly checks the predicate, so it'll continue only after the continuation has been triggered. The issue is about calling `notify` on a condition variable that has been released.

biddisco: calling notify on a release cond_var is not a problem. nothing happens. what is bothering you about it? I can move the set callback function inside the lock if it makes you happy.

msimberg: Exactly! Sorry, I updated "release" to "destroy" in the table in the first comment.

biddisco:
I just noticed that when I pushed the fix to the mpix registration inside lock, I also included a couple of other things that I didn't intend to push. Apologies, However, you could include them as they are "safe" (meaning the use of thread level optimization is by default disabled), so it does no harm to have it in.

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -512,13 +506,36 @@ int main(int argc, char* argv[])
...
 
+    // -----------------
+    // process command line options early so we can use them for mpi_init
+    namespace po = pika::program_options;
+    po::variables_map vm;
+    po::store(po::command_line_parser(argc, argv).options(cmdline).allow_unregistered().run(), vm);
+    po::notify(vm);
+
+    if (vm["mpi-optimizations"].as<bool>()) { mpix::enable_optimizations(true); }
+    else { mpix::enable_optimizations(false); }
```
msimberg, line 527:
```suggestion
    mpix::enable_optimizations(vm["mpi-optimizations"].as<bool>());
```

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
msimberg: Note, https://github.com/pika-org/pika/pull/1128/commits/53f257eee822cf9a874059eafe78d5f8407df9b2 does not fix the issue, and I would remove that commit. The registering of a continuation does not need to happen under a lock. The notification of the condition variable has to be done under a lock (the code I'm commenting on here).

biddisco: It does fix it surely, because notify cannot be called until the lock is released, and that only happens when the `cond_var` goes into wait. At this point the `cond_var` is guaranteed  to be alive.  
Previously, registering the continuation was outside the lock, it could be triggered and `cond_var` notified by thread 2 in your table above `before` the lock was taken by thread 1, and the `cond_var` might be destroyed before thread 2 calls notify.  

New table


|  thread 1 | thread 2 |
| --- | --- |
| take lock | |
| register continuation | |
| | continuation ready (earliest possible) |
| | `mpix_callback` |
| wait with predicate (+unlock)| |
| | take lock |
| | set completed = true |
| | release lock |
| | notify *destroyed* condition variable |
| wake - completed is true, continue | |
| destroy condition variable | |
| | :smile: : no segfaults possible |

I'm not keen on calling `notify` under lock, because that guarantees contention on the wakeup. Is my analysis faulty?

msimberg: I think the unlock of this internal will still be problematic: https://github.com/pika-org/pika/blob/a40b9b5fac72969c5fd99942a3ce169b92d299aa/libs/pika/synchronization/include/pika/synchronization/condition_variable.hpp#L68. The notify will wake up the waiting thread, the waiting thread will proceed to destroy the condition variable, and unlocking the internal mutex will be a write to freed memory.

Not directly related to this PR, but still relevant: I've started looking into running the unit tests with _all_ the completion modes. Once we have that we would also get runs with thread sanitizer on the MPI continuations mode, and we can then a bit more confidently try to relax constraints if it looks like it's needed and/or possible.

There may of course also be ways of rewriting the condition variable itself to avoid issues like that. However, until then I'd prefer to go with the safer option.

How does that sound?

biddisco: If the condition variable can only be woken up when the notifier holds the lock, then something is very wrong. 

The line you highlighted is where a lock is taken, which is moved into the `detail::cond_var` notify_one and from there, a task is taken from the queue and resume is called on it - if that task were now to destroy the `cond_var` immediately - the one thing that could go wrong is that the next operation - unlocking the lock as it goes out of scope - an  `atomic<bool>` stored in the cond_var_data mutex -- is set to false. If the data is already destroyed, that atomic set will fail. 
Is it possible to have the condition variable destroy itself whilst I'm still holding that lock with its internal mutex?
If tjhjere is nothing that checks that lock before the destructor is called, then there might be a problem. I looked through the code and can't see anything protecting it. 
Since the `cond_var` data is held with an `intrusive_ptr` then one could m ake it safe by adding an `inclusive_ptr_add_ref` inside an RAII wrapper inside the top level notify - then the data could not go out of scope if the woken thread deletes the cond_var.

This seems wrong. what have I misunderstood?

msimberg: > If the condition variable can only be woken up when the notifier holds the lock, then something is very wrong.

I kind of agree, but I'm not confident enough about the implementation to say if it has to be this way or if there's a way around it. There are cases where it'd be safe to do the notify without a lock, such as if both waiter and notifier hold a reference count on a heap-allocated condition variable.

> The line you highlighted is where a lock is taken, which is moved into the `detail::cond_var` notify_one and from there, a task is taken from the queue and resume is called on it - if that task were now to destroy the `cond_var` immediately - the one thing that could go wrong is that the next operation - unlocking the lock as it goes out of scope - an `atomic<bool>` stored in the cond_var_data mutex -- is set to false. If the data is already destroyed, that atomic set will fail. Is it possible to have the condition variable destroy itself whilst I'm still holding that lock with its internal mutex? If tjhjere is nothing that checks that lock before the destructor is called, then there might be a problem. I looked through the code and can't see anything protecting it.

Aha, I think I see now. Since the waiter has to re-take the internal lock after waking up, the notifier will release the lock before the condition variable is destroyed. With this I think this is indeed now safe...

Since the `cond_var` data is held with an `intrusive_ptr` then one could m ake it safe by adding an `inclusive_ptr_add_ref` inside an RAII wrapper inside the top level notify - then the data could not go out of scope if the woken thread deletes the cond_var.
> 
> This seems wrong. what have I misunderstood?

That could probably help, assuming the analysis above is wrong.

Sorry for being difficult about this. I think you're right that the latest version is ok as is. Unless I've now triggered you to figure out other ways that this can break, let's go with the current version.

msimberg (approved):
Let's get this in and improve later as needed.

Thank you @biddisco!

Issue #789, opened 2023-09-26
Try to use the CSCS spack build cache in CI 


aurianer:
Try to use the CSCS spack build cache in CI once #32 is done.

msimberg:
I would keep this open as it's sufficiently different than what was implemented in #962 and may still be useful to do at some point. However, it's lower priority now that #962 is merged.

msimberg:
> What about using https://cache.spack.io/tag/develop-2024-09-22/?stack=developer-tools-manylinux2014? I think it's pretty straightforward to add

Can you try reviving https://github.com/pika-org/pika/pull/1129 with a new cache? I don't even think you need to choose the stack, but it's ok to use the generic develop-yyyy-mm-dd cache. If I remember correctly the HIP build issue is now actually fixed, so we should be able to use the cache on all pipelines.

The above is still a read-only cache, while this issue is about a read-write cache. A read-only cache shouldn't hurt though so it's worth trying it again now assuming the HIP build issue is actually fixed.

PR #1129, opened 2024-05-28
Enable upstream spack binary cache in CI

msimberg:
Related to https://github.com/pika-org/pika/issues/789, but does not fully cover the issue. This may be enough though. This simply uses the upstream cache to download packages, but does not populate a cache.

This seems to work well for pulling dependencies to build the compilers, but does not pull dependencies built with the spack-built compiler. At least the compiler build step is slightly faster with the cache, so I think this is still an improvement.

There's a new variable `SPACK_BUILDCACHE` which is currently just set to `SPACK_COMMIT`. If `SPACK_COMMIT` is an arbitrary hash, there won't be a corresponding cache, and in that case one might want to set `SPACK_BUILDCACHE` to something else explicitly. I've also updated `SPACK_COMMIT` to a `develop-X-Y-Z` tag which does have a build cache, so they can be the same.

### DIFF
--- a/.gitlab/docker/Dockerfile.spack_base
+++ b/.gitlab/docker/Dockerfile.spack_base
@@ -53,4 +53,6 @@ RUN spack external find --exclude python --scope site \
     && spack compiler find --scope site \
     && ([[ "${USE_SPACK_BUILDCACHE}" != "true" ]] \
         || (spack mirror add ${SPACK_BUILDCACHE} https://binaries.spack.io/${SPACK_BUILDCACHE} \
-            && spack buildcache keys --install --trust))
+            && spack mirror add develop https://binaries.spack.io/develop \
+            && spack buildcache keys --install --trust \
+            && spack mirror rm develop))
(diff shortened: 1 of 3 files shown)


msimberg:
Update description with some more details. This seems to pull dependencies for the compiler step nicely, but does not (somewhat understandably) pull dependencies from the cache for the pika dependencies step. I'll keep #789 open since this doesn't really fix it fully.

msimberg:
The HIP build also broke for some reason. Needs further investigation.

msimberg:
I'm closing this for now. Using an older commit of spack does not seem to affect the HIP builds, and I don't want to dive deeper into this at the moment. We can come back to this in the future.

msimberg:
With a "working" cache (the keys are not available on the `develop-2024-09-22` cache, but one can get them from the `develop` cache) this is again broken in new ways. One failure is (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/7974294872#L781):
```
ModuleNotFoundError: No module named 'distutils'
```
Maybe fixable?

Another is (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/7974294884#L1225):
```
 /bin/sh: 1: /usr/bin/gcc-12: not found
```
Probably some package in the cache has a system `gcc-12` hardcoded somewhere...

I'll close this again. We can give this another go in 6 months...

PR #1130, opened 2024-05-29
Enable testing for gh200

aurianer:
We were only building for arm64 but not runnning anything. This enables 2 pipelines on santis.

### DIFF
--- a/.gitlab/includes/gcc12_cuda12_santis_pipeline.yml
+++ b/.gitlab/includes/gcc12_cuda12_santis_pipeline.yml
@@ -8,7 +8,7 @@ include:
   - local: '.gitlab/includes/common_pipeline.yml'
   - local: '.gitlab/includes/common_spack_pipeline.yml'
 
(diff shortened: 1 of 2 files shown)


msimberg:
Thanks for fixing the process mask regex! This looks good on the CI configuration side. However, the test failures remain. Have you had a chance to look into those?

msimberg:
We'd need to revive this quite soon as daint is going down. @aurianer do you think you will have time to continue this or should I have a look?

msimberg:
I've cherry-picked the thread binding regex into https://github.com/pika-org/pika/pull/1240.

msimberg:
> If I'm not mistaken, this is now covered by #1239 no?

Yep, you're right. I suppose the addition in this PR is that it builds without CUDA on santis/todi/daint. If we can get the CUDA testing working that should be sufficient to cover the non-CUDA case as well. I'll close this.

PR #1133, opened 2024-05-30
Fix `--no-tests=error` ctest flag in build step of CI

msimberg:


msimberg:
This is a draft since one of the pipelines (performance CI) actually doesn't run any tests and fails with the correct flag. I might try to move the performance tests to run on all pipelines instead to fix this, but I'll get back to this at a later point.

msimberg:
I've moved the performance tests to the generic test job step, and removed the dedicated performance test pipeline. CI seems to be stuck, so it remains to be seen if I've done it right...

msimberg:
Newer CMake versions actually seem to error out on the wrongly spelled `-no-tests=error`. I'll try to prioritize finishing this PR.

PR #1134, opened 2024-05-30
Use low CPU frequency in performance CI pipeline

msimberg:
Testing what effect this has on performance. The hope is to have less variance, but given that we don't pass/fail PRs based on these results, it may not be necessary to set this.


msimberg:
I'm going to close this without merging. I don't think we need this.

PR #1148, opened 2024-06-05
Add `CLUSTER_NAME` environment variable to CI job system metadata

### LINKED ISSUE #1062: Add CSCS cluster name to elastic metadata
When sending data to elastic from CI, we currently only send the slurm partition name and hostname, but not the cluster name. We should add that so that we can distinguish between the same partition names on different clusters.

msimberg:
This should fix #1062.

### DIFF
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -130,6 +130,8 @@ make_multiarch_image:
   extends:
     - .cmake_variables_common
   script:
+    - env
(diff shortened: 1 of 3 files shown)


msimberg:
Closing. `CLUSTER_NAME` is only set in login shells and slurm jobs are not run in login shells. I'll come back to this if we need this badly. Until then `SLURM_CLUSTER_NAME` works just fine (clariden doesn't exist right now, so #1062 is a non-issue at the moment).

Issue #1150, opened 2024-06-07
Test all MPI continuations completion modes in unit tests


msimberg:

msimberg:
MPI continuations are not yet tested. All other modes are now tested since #1151.

Issue #1144, opened 2024-06-03
Refactor `PIKA_MPI_COMPLETION_MODE` into independent flags


msimberg:
While for testing the current bitmask passed to `PIKA_MPI_COMPLETION_MODE` is convenient, the values are not very transparent to a user setting the values. It'd be nice to split up the flags into `PIKA_MPI_INLINE_CONTINUATION`, `PIKA_MPI_DEDICATED_POOL` etc. If still needed for testing we can consider keeping some way to override the mode with a bitmask as well.

PR #1151, opened 2024-06-07
Test `mpi_ring_async_sender_receiver` with all MPI completion modes

### LINKED ISSUE #1142: Test all MPI completion modes in unit tests
We should make sure all modes are tested to weed out bugs in the various alternative code paths.

msimberg:
Fixes #1142, except for testing the MPI continuations modes. Those will be enabled separately (https://github.com/pika-org/pika/issues/1150).

This:
- Adds a `--pika:mpi-completion-mode` flag for setting the mode via command line options. This isn't intended as the primary way to change the mode for users, but is mostly there for the unit test. CMake's `ENVIRONMENT` property on tests allows setting an environment variable. I originally intended to use this, but the variables are not exported so sub-processes don't see the value, and since the tests are run through `mpirun`/`mpiexec` the variable isn't applied.
- Decrements MPI counters before calling the user-provided callback. This fixes a race in the test checking for zero MPI requests in flight. The assertion is also changed to use `PIKA_TEST_EQ` so that it's run with all build types and prints the failing value if it fails.
- Adds a `--recv-before-send` flag to the unit test, and makes send-before-recv the default, to avoid deadlocks when requests are done inline. Sends will never block indefinitely, so doing the sends first avoids hangs.
- Some minor additional cleanup in the test (see individual commit messages for descriptions).

### DIFF
--- a/libs/pika/async_mpi/src/mpi_polling.cpp
+++ b/libs/pika/async_mpi/src/mpi_polling.cpp
@@ -604,11 +604,6 @@ namespace pika::mpi::experimental {
                         debug(
                             str<>("CB invoke"), ptr(mpi_data_.callbacks_[index].request_), status));
 
-                    // Invoke callback (PIKA_MOVE doesn't compile here)
-                    PIKA_INVOKE(std::move(mpi_data_.callbacks_[index].cb_), status);
-
-                    pika::threads::detail::decrement_global_activity_count();
-
                     // Remove the request from our vector to prevent retesting
                     mpi_data_.requests_[index] = MPI_REQUEST_NULL;
 
@@ -616,6 +611,11 @@ namespace pika::mpi::experimental {
                     // if invoked code checks in_flight value
                     --mpi_data_.all_in_flight_;
                     --mpi_data_.active_requests_size_;
+
+                    // Invoke callback (PIKA_MOVE doesn't compile here)
+                    PIKA_INVOKE(std::move(mpi_data_.callbacks_[index].cb_), status);
+
+                    pika::threads::detail::decrement_global_activity_count();
                 }
             } while (event_handled == true);
 
--- a/libs/pika/async_mpi/src/mpi_polling.cpp
+++ b/libs/pika/async_mpi/src/mpi_polling.cpp
@@ -742,6 +742,7 @@ namespace pika::mpi::experimental {
 
     // -----------------------------------------------------------------
     std::size_t get_completion_mode() { return detail::completion_flags_; }
+    void set_completion_mode(std::size_t mode) { detail::completion_flags_ = mode; }
 
     // -----------------------------------------------------------------
     bool create_pool(std::string const& pool_name, pool_create_mode mode)
--- a/libs/pika/async_mpi/src/mpi_polling.cpp
+++ b/libs/pika/async_mpi/src/mpi_polling.cpp
@@ -113,7 +113,7 @@ namespace pika::mpi::experimental {
             bool error_handler_initialized_ = false;
             int rank_ = -1;
             int size_ = -1;
-            std::size_t max_polling_requests = get_polling_default();
+            std::atomic<std::size_t> max_polling_requests{get_polling_default()};
 
             // The sum of messages in queue + vector
             std::atomic<std::uint32_t> all_in_flight_{0};
@@ -441,7 +441,7 @@ namespace pika::mpi::experimental {
 
                     int num_completed = 0;
                     // do we poll for N requests at a time, or just 1
-                    if (mpi_data_.max_polling_requests > 1)
+                    if (mpi_data_.max_polling_requests.load(std::memory_order_relaxed) > 1)
                     {
                         // it seems some MPI implementations choke when the request list is
                         // large, so we will use a max of max_poll_requests per test.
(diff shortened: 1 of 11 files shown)


libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -604,18 +604,18 @@ namespace pika::mpi::experimental {
...
 
                     // decrement before invoking callback to avoid race
                     // if invoked code checks in_flight value
                     --mpi_data_.all_in_flight_;
                     --mpi_data_.active_requests_size_;
+
+                    // Invoke callback (PIKA_MOVE doesn't compile here)
+                    PIKA_INVOKE(std::move(mpi_data_.callbacks_[index].cb_), status);
+
+                    pika::threads::detail::decrement_global_activity_count();
```
biddisco, line 618: The multithreaded poll also needs this change (in a couple of places)

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
```diff
@@ -261,59 +264,66 @@ struct message_receiver
...
+            if (recv_before_send)
+            {
+                f_recv();
+                f_send();
+            }
+            else
+            {
+                f_send();
+                f_recv();
             }
```
biddisco, line 322: much cleaner to read. thanks

libs/pika/command_line_handling/src/parse_command_line.cpp
```diff
@@ -338,6 +338,11 @@ namespace pika::detail {
                 "allowed values: 0 - no NUMA sensitivity, 1 - allow only for "
                 "boundary cores to steal across NUMA domains, 2 - "
                 "no cross boundary stealing is allowed (default value: 0)")
+#if defined(PIKA_HAVE_MPI)
+            ("pika:mpi-completion-mode", value<std::size_t>(),
+                "the pika MPI polling completion mode (only available if MPI "
+                "built with MPI support)")
+#endif
```
biddisco, line 345: Really would like the ability for modules to add options in a modular way - then I would not have used ENV vars at all.

biddisco (changes requested):

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
msimberg: I considered adding an enum instead of the current "override default behaviour flag", something like `--mpi-order=send-before-recv|recv-before-send`. Would that be nicer, or is the current flag ok?

libs/pika/command_line_handling/src/parse_command_line.cpp
msimberg: I understand that. I think for real-world use cases the environment variables are still the most useful way of setting options since command line options might never make it into pika.

Note, this is likely anyway a temporary change until we get something in the direction of https://github.com/pika-org/pika/issues/1144.

Ideally, users don't have to set the completion mode at all (because the default is good enough for most use cases).

libs/pika/async_mpi/src/mpi_polling.cpp
msimberg: To clarify, the multithreaded polling is already handling the counts this way as far as I can tell. That's why you don't see any changes to it here.

biddisco: ok. ignore my comment. I misread the code.

libs/pika/async_mpi/tests/unit/mpi_ring_async_sender_receiver.cpp
biddisco: Ideally we would not have a flag at all, just change to the version that works. Leave it as you've done it. In general, pre-posting recvs is a good thing, which is why I did it that way, but the fact that is locks up makes it bad. So merge this as is and in the next round of concolidation/cleaning I'll reassess it.

biddisco (approved):

msimberg:
Not to be merged yet. The issue supposedly fixed by https://github.com/pika-org/pika/pull/1151/commits/cf856d237c86990b442d74a8127d5140ac0a24b9 still happens occasionally. Thanks to @biddisco for noticing this.

msimberg:
@biddisco I cherry-picked https://github.com/pika-org/pika/pull/1180/commits/1986abe3789a682da203cdd1ee3955a817fb6802 from your PR (https://github.com/pika-org/pika/pull/1180). Thank you for looking into that! Was that the only commit required to fix the remaining failures?

biddisco:
> @biddisco I cherry-picked [1986abe](https://github.com/pika-org/pika/commit/1986abe3789a682da203cdd1ee3955a817fb6802) from your PR (#1180). Thank you for looking into that! Was that the only commit required to fix the remaining failures?

Should be all that is needed and safe to merge now

biddisco (approved):
ok. Fortunately, we're not really using the polling size now, but I'll make sure that stuff get's checked next time I change it.

msimberg:
> Is this PR good to go now?

Yes, should be. I was hoping to get CSCS CI back for testing before merging it.

PR #1152, opened 2024-06-07
Add task yield benchmark

msimberg:
The added benchmark simply times how long a yield and reschedule takes, i.e. two context switches and adding and removing the task from a queue.

This also adds the benchmark to CI with both `--pika:threads=1` and `--pika:threads=2`. The one-thread version runs significantly faster, probably due to less contention from other threads trying to steal tasks. Running with more than one worker thread and the static scheduler gives roughly the same timings as a single worker thread with the default scheduler.


biddisco (approved):
seems reasonable

PR #1155, opened 2024-06-10
Remove `shared_mutex`

msimberg:
`shared_mutex` is currently badly maintained, badly tested, and possibly buggy with no pressing need to fix it. Removing it until there is a need for it, hopefully with more extensive testing.


msimberg:
As this only removes functionality I'm hoping that merging without the full CI will be safe.

PR #1156, opened 2024-06-10
Enable LTO in CircleCI configuration

msimberg:


msimberg:
I'd like to try it out a bit more broadly first (e.g. in DLA-Future and in other configurations), but yeah, assuming it doesn't break anything it sounds like a good idea to enable it by default! This PR was mostly to see if it actually breaks anything as a first step. I opened https://github.com/pika-org/pika/issues/1163 as a follow up. I suppose a next step would be trying to enable it on all CI configurations. I do believe it has to be overridable to disable in case of future issues, but as long as that's possible I don't see why we couldn't enable it by default. In spack we could also consider simply adding an `lto` variant (or should that be handled with `cxxflags`, not sure...).

msimberg:
Since this only affects the CircleCI configuration (which is currently working) I'm going to go ahead and merge this.

PR #1159, opened 2024-06-11
Fix potential hang in `thread_rescheduling_test`

msimberg:


msimberg:
There is another hang, for which I think I've found a cause. If I manage to have a clean fix for it before CI is ready for this to be merged I'll add it here; otherwise I'll open a follow-up PR.

msimberg:
The second commit fixes the other issue. This one is not directly a race. The `set_thread_state` call that is supposed to tell the task to terminate simply doesn't apply if the task is already `pending`, so I've added a loop to retry as long as the previous state was `pending` (which means nothing changed).

PR #1169, opened 2024-06-12
Enable LTO one CI configuration

msimberg:


msimberg:
Since performance tests are now run on all CI configurations, I updated this to simply use IPO on one of the configurations. I've not changed the jenkins performance tests configuration.

PR #1170, opened 2024-06-12
Run coverage on PRs

msimberg:
We currently only run the coverage workflow on `main`. This enables it for pull requests and the `release-*` branches as well.

Since PRs don't have access to secrets this uses a `workflow_run` workflow, which does have access to secrets, to upload the coverage report. The PR itself simply generates the coverage report and stores it in GitHub artifacts and then the `workflow_run` workflow downloads the file and pushes it to codacy.

### DIFF
--- a/.github/workflows/linux_coverage.yml
+++ b/.github/workflows/linux_coverage.yml
@@ -7,9 +7,13 @@
 name: Linux CI (Coverage)
 
 on:
+  merge_group:
+  pull_request:
   push:
     branches:
+      # Development and release branches
       - main
+      - release**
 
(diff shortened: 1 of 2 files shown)


msimberg:
Given that our coverage isn't really being reported anyway on codacy at the moment I'm going to go ahead and merge this for testing. The `workflow_run` documentation says:

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

(presumably for security reasons) so I can't test this purely on the PR. I'll revert/modify depending on how the merged changes work on `main`.

PR #1173, opened 2024-06-13
Fix coverage report upload job

msimberg:
#1170 had an unnecessary change of working directory to a non-existent directory. This removes the change of working directory and uses the default.


msimberg:
The workflows are triggered, but again, this can't be tested in a PR because GitHub will use the workflow definition on `main` (again, for security reasons). Another blind attempt, but hopefully with better success.

PR #1180, opened 2024-06-18
Fix `transform_mpi` test

### LINKED ISSUE #612: Re-enable disabled MPI tests

biddisco:
Fixes #612 (mpi algorithm test). 

Note : Async storage test is still experimental and not essential to the testing framework - it should be moved to performance tests when working correctly.


libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -191,19 +188,18 @@ int pika_main()
...
+                        (e.get_error() == pika::error::invalid_status);
+                    PIKA_TEST_MSG(err_ok, "Returned error code was not in expected list");
+                    std::vector<std::string> err_msgs = {
+                        "null datatype", "Invalid datatype", "MPI_ERR_TYPE"};
+                    bool msg_ok = false;
+                    for (const auto& msg : err_msgs)
+                    {
+                        msg_ok |= (std::string(e.what()).find(msg) != std::string::npos);
+                    }
+                    PIKA_TEST_MSG(msg_ok, "Error message did not contain expected string");
```
msimberg, line 202: I kind of liked being explicit about which MPI implementation produces what error, but understand that it might be annoying to keep up to date. Could you summarize how the exceptions/error messages changed from before? I currently don't see any explicit uses of `invalid_status` in the `async_mpi` module, but do see `bad_function_call`. Do we still expect `invalid_status` to be thrown, and if so, when?

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -114,10 +112,9 @@ int pika_main()
                 // MPI function pointer
                 int data = 0, count = 1;
                 if (rank == 0) { data = 42; }
-                auto result = tt::sync_wait(
+                tt::sync_wait(
                     ex::just(&data, count, datatype, 0, comm) | mpi::transform_mpi(MPI_Ibcast));
                 if (rank != 0) { PIKA_TEST_EQ(data, 42); }
```
msimberg, line 117: For consistency?
```suggestion
                PIKA_TEST_EQ(data, 42);
```

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -114,10 +112,9 @@ int pika_main()
                 // MPI function pointer
                 int data = 0, count = 1;
                 if (rank == 0) { data = 42; }
-                auto result = tt::sync_wait(
+                tt::sync_wait(
```
msimberg, line 115: Do I understand correctly that functions that return `MPI_Status` are now special cased to turn that into an exception (if not success)? What about functions that don't return `MPI_Status` (nor void)?

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -4,8 +4,9 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-set(tests # algorithm_transform_mpi
-    mpi_ring_async_sender_receiver
+set(tests #
+    algorithm_transform_mpi #
+    mpi_ring_async_sender_receiver #
```
msimberg, line 9: Do we need the `#`s?

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
biddisco: no. I just left the `invalid status` there cos that's what you had in your version, but I mistakenly thought we converted some mpi error status into the exception type, but in fact we always use `bad function call`. I'll remove it

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
biddisco: we only handle mpi type functions that return mpi status and we set any error on the set_error. I can put back the arbitrary return types if necessary, but I'd prefer to do that in a PR when it is needed by something. Removing it simplified my work a bit initially.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: they stop cmake-format from sticking everything on one line. I don't like our cmake-format rules very much ;)

msimberg: I kind of guessed that was the reason. You know we can change the configuration, right? In particular, I think this one controls how this is wrapped: https://github.com/pika-org/pika/blob/0803db56fb5a32920b6563149a32f674a5bdc5d6/.cmake-format.py#L46. Setting it to `1` would put these elements all on separate lines.

Why don't revert this change and open a separate PR with the `cmake-format.py` changes you'd propose, then we can discuss there?

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -54,24 +55,22 @@ namespace pika::mpi::experimental::detail {
...
     // -----------------------------------------------------------------
-    // return a scheduler on the default pool with added priority if requested
-    inline auto default_pool_scheduler(execution::thread_priority p)
+    // return a scheduler on the mpi pool, with or without stack
+    inline auto mpi_pool_scheduler(execution::thread_priority p)
     {
+        if (!pool_exists()) return default_pool_scheduler(p);
         return ex::with_priority(
-            ex::thread_pool_scheduler{&resource::get_thread_pool("default")}, p);
+            ex::thread_pool_scheduler{&resource::get_thread_pool(get_pool_name())}, p);
```
msimberg, line 73: Just checking: why does this not care about stack/no stack anymore? Is that intentional?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -224,6 +224,7 @@ namespace pika::mpi::experimental {
           : pool_name_(pool_name)
         {
             mpi::experimental::init(false, init_errorhandler);
+            mpi::experimental::register_polling();
```
msimberg, line 227: What happened here? Were we not registering polling before in the RAII helper? Should it be unregistered again in the destructor?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -203,7 +203,7 @@ namespace pika::mpi::experimental {
         std::size_t get_completion_mode_default()
         {
             // inline continuations are default
```
msimberg, line 205: The comment isn't true, right?

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -234,19 +229,33 @@ int pika_main()
         }
     }
 
-    test_adl_isolation(mpi::transform_mpi(my_namespace::my_sender{}, [](MPI_Request) {}));
+    test_adl_isolation(mpi::transform_mpi(my_namespace::my_sender{}, [](MPI_Request*) {}));
```
msimberg, line 232: I think I missed this change in another PR: does this need to be passed by pointer? I'd prefer if we did it by reference.

libs/pika/command_line_handling/src/command_line_handling.cpp
```diff
@@ -398,7 +399,9 @@ namespace pika::detail {
     std::size_t handle_mpi_completion_mode(
         detail::manage_config& cfgmap, pika::program_options::variables_map& vm)
     {
-        std::size_t completion_mode = cfgmap.get_value<std::size_t>("pika.mpi.completion_mode", 0);
+        int def_val = pika::detail::get_env_var_as<std::size_t>("PIKA_MPI_COMPLETION_MODE", 0);
+        std::size_t completion_mode =
+            cfgmap.get_value<std::size_t>("pika.mpi.completion_mode", def_val);
```
msimberg, line 404: I think this should not be needed, but did you see a change in behaviour after you added this? Note the default value for `pika.mpi.completion_mode` is already set in https://github.com/pika-org/pika/blob/50c60c37ef513bce9091826853e9c7fa4b8b71b1/libs/pika/runtime_configuration/src/runtime_configuration.cpp#L342. The `0` as a default value is a bit misleading. It should really just be an error if there's no default, because there should always be a default in the configuration.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: AFAICR - the algorithm_transform_mpi test had to be changed to allow for pool creation etc and the polling RAII helper called init - but due to changes in the way we setup stuff, the polling was skipped, so I added it here - but when the RAII helper destructs, it calls `mpi::experimental::finalize(pool_name_);` which does unregister polling for us

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
biddisco: I removed the option to use `no_stack` because it wasn't implemented properly everywhere and it made things more complicated. I can put it back in, but where should we decide if a task is created `no_stack` or not. if we transfer a task to the mpi pool for dispatch, then probably yes `no_stack` makes sense - but when we trigger a continuation it might not. I will add back no_stack in another PR after thinking more. (I'll clean up the comments in this PR)

libs/pika/command_line_handling/src/command_line_handling.cpp
biddisco: The stuff you put in broke the existing logic because you overwrite the completion mode with a new default value, so if I set the completion mode using an ENV var, it was thrown out by the default value set by `cfgmap.get_value<std::size_t>("pika.mpi.completion_mode", def_val);` so I simply made sure that the default value set there was taken from the env var. If nothing is passed, the default value of zero is used (which can be changed to something else). But in the main mpi init, we set the default to zero, hence the use here.

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
biddisco: actuaally, dispatch passes a request to trigger, which returns nothing, so this is wrong, but when I remove the args from the lambda, compilation fails. not quite sure what the problem is.

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
msimberg: Thanks, that makes sense. Let's leave it out from here at least.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
msimberg: I see, thanks. Maybe we should separate those so that it's somewhat symmetric in the constructor and destructor (i.e. that one does explicit init+register and the other does explicit finalize+unregister). We don't need to change it here, but what do you think of this change?

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
msimberg: Well, you definitely  need _an_ argument to the lambda. The question is what type of request it should be. The lambdas to other `transform_mpi` calls also take the request by pointer, so it seems to be consistent. I'm just wondering why it needs to be a pointer (rather than reference, or even by value)?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
msimberg: Also, it looks like the registering does not take into account the case that there's no `pool_name` set, while unregistering does. That should probably also be updated to be consistent? Again, not for this PR.

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
msimberg: After discussion this is indeed the correct fix. The callables take a pointer to a request to mirror what MPI functions do. This test was not using the correct argument type, and it's possible that we were previously getting some implicit conversion or similar that hid this mistake.

Thanks @biddisco for fixing this!

libs/pika/command_line_handling/src/command_line_handling.cpp
msimberg: See https://github.com/biddisco/pika/pull/2.

libs/pika/async_mpi/CMakeLists.txt
```diff
@@ -11,6 +11,33 @@ if(NOT ${PIKA_WITH_MPI})
...
+if("${MPIX_COMPILE_RESULT}" AND ("${MPIX_RUN_RESULT}" EQUAL 0))
+  message(STATUS "MPIx Continuations detected")
+  set(MPIX_CONTINUATIONS_AVAILABLE True)
+  set(MPIX_CONTINUATIONS_LOOP_COUNT 79)
+else()
+  message(STATUS "MPIx Continuations not detected")
+  set(MPIX_CONTINUATIONS_AVAILABLE False)
+  set(MPIX_CONTINUATIONS_LOOP_COUNT 63)
+endif()
+
```
msimberg, line 40: Can we move this to all the other configuration tests please? This can use the same mechanisms (`pika_add_config_test` etc.). I think this can be a compile-only test, see other comment.

msimberg: Let's call the result `PIKA_WITH_MPIX_CONTINUATIONS`?

libs/pika/async_mpi/src/check_openmpi_continuations.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#if __has_include(<mpi-ext.h>)
+# include <mpi-ext.h>
+#endif
+
+#ifdef OMPI_HAVE_MPI_EXT_CONTINUE
+int main() { return 0; }
+#else
+int main() { return 1; }
+#endif
```
msimberg, line 20:
```suggestion
//  Copyright (c) 2024 ETH Zurich
//
//  SPDX-License-Identifier: BSL-1.0
//  Distributed under the Boost Software License, Version 1.0. (See accompanying
//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)

#include <mpi.h>
#include <mpi-ext.h>

#if !defined(OMPI_HAVE_MPI_EXT_CONTINUE)
static_assert(false);
#endif

// I forget if we need a main or not here, remove if not necessary.
int main() {}
```
If we don't have MPI this will fail already at the includes. If we don't have OpenMPI this will also fail at the `mpi-ext.h` include. The `static_assert` takes care of the rest. This way we don't have to run it and check the result. What do you think?

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -40,7 +37,7 @@ foreach(test ${tests})
     FOLDER "Tests/Unit/Modules/AsyncMPI"
   )
 
-  foreach(polling_mode RANGE 0 63 1)
+  foreach(polling_mode RANGE 0 ${MPIX_CONTINUATIONS_LOOP_COUNT} 1)
```
msimberg, line 40:
```suggestion
  foreach(polling_mode RANGE 0 ${PIKA_MPIX_CONTINUATIONS_NUM_METHODS} 1)
```
?

msimberg (changes requested):

libs/pika/async_mpi/src/check_openmpi_continuations.cpp
biddisco: I'm updated the checks to use that. Tested it locally and it is fine, thanks

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: oops didn't see this until now, but I already renamed the loop counter to `PIKA_MPI_MODES_LOOP_COUNT`

libs/pika/async_mpi/CMakeLists.txt
biddisco: oh dear. I didn't see this comment either. Personally, I prefer to have all the mpi related flags/checks/vars inside the mpi module and not in some remote place - where did you have in mind for it to go?

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg: Ok. I thought we can leave out `LOOP_COUNT` since that signals how it's used, rather than what it represents, but let's stick this for now then.

libs/pika/async_mpi/CMakeLists.txt
biddisco: ook. I moved it all to cmake/tests

msimberg: Thanks. I understand your motivation for keeping in the MPI module, but given all other feature checks and CMake options are defined currently with the "global" options, I think it's easier to manage if this test is also added to where everything else is.

cmake/tests/check_openmpi_continuations.cpp
```diff
@@ -0,0 +1,17 @@
+//  Copyright (c) 2021 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// simple compile test to see if mpix continuuations are present
+// returns 0 when available, 0 otherwise
```
msimberg, line 8:
```suggestion
// simple compile test to see if mpix continuations are present
```

libs/pika/async_mpi/CMakeLists.txt
```diff
@@ -41,3 +50,9 @@ pika_add_module(
     pika_config
   CMAKE_SUBDIRS examples tests
 )
+
+option(PIKA_WITH_MPI_GENERATE_SCRIPTS "Generate benchmarking scripts for mpi tests" OFF)
+mark_as_advanced(PIKA_WITH_MPI_GENERATE_SCRIPTS)
+if(PIKA_WITH_MPI_GENERATE_SCRIPTS)
+  add_subdirectory(scripts)
+endif()
```
msimberg, line 58: Can we leave this for another PR? The `scripts` subdirectory is not addeed here as far as I can tell.

biddisco: oops. that wasn't supposed to be committed. will remove and force push

.github/workflows/linux_asan_ubsan_lsan.yml
```diff
@@ -84,6 +84,7 @@ jobs:
             export UBSAN_OPTIONS=print_stacktrace=1:suppressions=$PWD/tools/ubsan.supp
             cd build
             ctest \
+              --repeat-until-fail 7 \
```
msimberg, line 87:
```suggestion
```

.github/workflows/linux_asan_ubsan_lsan.yml
```diff
@@ -99,6 +100,7 @@ jobs:
...
```
msimberg, line 103:
```suggestion
```

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -60,7 +60,9 @@ namespace pika::mpi::experimental::detail {
 #if defined(PIKA_HAVE_STDEXEC)
 
         using completion_signatures = pika::execution::experimental::completion_signatures<
-            pika::execution::experimental::set_value_t(MPI_Request)>;
+            pika::execution::experimental::set_value_t(MPI_Request),
+            pika::execution::experimental::set_error_t(std::exception_ptr),
```
msimberg, line 64: This is not quite accurate, as it could be forwarding another error type from previous senders. I can update this separately from your PR if you'd like, but wanted to point it out here.

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -95,7 +97,7 @@ namespace pika::mpi::experimental::detail {
 
                 template <typename Error>
                 friend constexpr void
-                tag_invoke(ex::set_error_t, dispatch_mpi_receiver r, Error&& error) noexcept
+                tag_invoke(ex::set_error_t, dispatch_mpi_receiver&& r, Error&& error) noexcept
```
msimberg, line 100: Just to check, was the `&&` required to get this to compile? The receiver should be (cheaply) move-constructible, so both should work.

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -98,7 +101,7 @@ namespace pika::mpi::experimental::detail {
 
                 template <typename Error>
                 friend constexpr void
-                tag_invoke(ex::set_error_t, trigger_mpi_receiver r, Error&& error) noexcept
+                tag_invoke(ex::set_error_t, trigger_mpi_receiver&& r, Error&& error) noexcept
```
msimberg, line 104: Same here: was the `&&` necessary?

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -59,7 +59,10 @@ namespace pika::mpi::experimental::detail {
 
 #if defined(PIKA_HAVE_STDEXEC)
         using completion_signatures = pika::execution::experimental::completion_signatures<
-            pika::execution::experimental::set_value_t()>;
+            pika::execution::experimental::set_value_t(),
+            pika::execution::experimental::set_error_t(std::exception_ptr),
```
msimberg, line 63: Also same comment here about forwarding other error types. I can take care of this separately.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -866,9 +853,10 @@ namespace pika::mpi::experimental {
         auto mode = get_completion_mode();
         if (mode >= pika::detail::to_underlying(detail::handler_method::unspecified))
         {
-            PIKA_THROW_EXCEPTION(
-                pika::error::invalid_status, "Bad completion flags", "invalid completion mode");
+            // PIKA_THROW_EXCEPTION(
+            //     pika::error::invalid_status, "Bad completion flags", "invalid completion mode");
```
msimberg, line 857: Why is this disabled?

libs/pika/async_mpi/tests/unit/algorithm_transform_mpi.cpp
```diff
@@ -6,31 +6,154 @@
...
 #include <mpi.h>
 #include <string>
 #include <utility>
+#include <vector>
+
+#if defined(__clang__) || defined(__GNUC__)
+# define ATTRIBUTE_NO_SANITIZE_ADDRESS __attribute__((no_sanitize_address))
+#else
+# define ATTRIBUTE_NO_SANITIZE_ADDRESS
+#endif
```
msimberg, line 25: Please use https://github.com/pika-org/pika/blob/42b27fa3d827b622bf8a0b990859ba29b5a125fa/libs/pika/config/include/pika/config/compiler_specific.hpp#L170 instead.

msimberg (changes requested):

msimberg:
Diff coverage failing is ok. The only uncovered lines are MPI continuations-related and we can't test those (easily) at the momeent in CI.

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
biddisco: is this the correct way to do it ?
```
        using completion_signatures =
            pika::execution::experimental::transform_completion_signatures_of<std::decay_t<Sender>,
                pika::execution::experimental::empty_env,
                pika::execution::experimental::completion_signatures<
                    pika::execution::experimental::set_error_t(std::exception_ptr)>,
                pika::execution::experimental::completion_signatures<
                    pika::execution::experimental::set_value_t(MPI_Request),
                    pika::execution::experimental::set_stopped_t()>>;
```

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
biddisco: no, it was fine but I changed that before finding the correct fix later ....

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
msimberg: Close, but not quite, I think.

This is right, you want to add `set_error_t(exception_ptr)`:
```
                pika::execution::experimental::completion_signatures<
                    pika::execution::experimental::set_error_t(std::exception_ptr)>,
```
You could/should move the `set_stopped_t()` signature into this `completion_signatures`.

The `set_value_t` is wrong, I think. 

You'd want something like this:
```
        template <typename...>
        using no_value_completion = completion_signatures<>;
        using completion_signatures =
            pika::execution::experimental::transform_completion_signatures_of<std::decay_t<Sender>,
                pika::execution::experimental::empty_env,
                pika::execution::experimental::completion_signatures<
                    pika::execution::experimental::set_value_t(MPI_Request),
                    pika::execution::experimental::set_error_t(std::exception_ptr),
                    pika::execution::experimental::set_stopped_t() // this also gets added by the default SetDone template parameter
                >,
                no_value_completion>;
```
You're essentially saying I definitely want these `set_value/error/done` completion signatures, and I don't want any of the value completions from the previous sender. Implicitly you're also saying I want additionally all the error completions from the previous sender (it's the default `SetError`: https://github.com/NVIDIA/stdexec/blob/cddcd5a17dbec59d8502c0c6108553a20e74d2da/include/stdexec/__detail/__transform_completion_signatures.hpp#L319-L320).

See https://github.com/NVIDIA/stdexec/blob/cddcd5a17dbec59d8502c0c6108553a20e74d2da/include/stdexec/__detail/__transform_completion_signatures.hpp#L308-L369 for details.

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
msimberg: Thanks, do you mind removing the `&&`. We've generally tried to move to taking the receivers by value since it's much simpler to reason about (and not any more expensive assuming the receiver just holds a pointer to the operation state).

libs/pika/async_mpi/src/mpi_polling.cpp
biddisco: That was me playing with dlaf and using the 'old' mode which can be removed and this put back

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -57,12 +57,17 @@ namespace pika::mpi::experimental::detail {
...
+        template <typename...>
+        using no_value_completion = completion_signatures<>;
+        using completion_signatures =
+            pika::execution::experimental::transform_completion_signatures_of<std::decay_t<Sender>,
+                pika::execution::experimental::empty_env,
+                pika::execution::experimental::completion_signatures<
+                    pika::execution::experimental::set_value_t(MPI_Request),
+                    pika::execution::experimental::set_error_t(std::exception_ptr),
+                    // this also gets added by the default SetDone template parameter
+                    pika::execution::experimental::set_stopped_t()>,
```
msimberg, line 69: The comment was mostly meant for you, not to reproduce in the code necessarily:
```suggestion
                    pika::execution::experimental::set_error_t(std::exception_ptr)>,
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -85,6 +85,8 @@ namespace pika::mpi::experimental {
                         return ex::schedule(default_pool_scheduler(p));
                     }
                     return just(request) | trigger_mpi(mode) | transfer(default_pool_scheduler(p));
+                    return just(request) | trigger_mpi(mode) |
+                        ex::continues_on(default_pool_scheduler(p));
```
msimberg, line 89: This looks suspicious.

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -57,11 +57,17 @@ namespace pika::mpi::experimental::detail {
...
+        template <typename...>
+        using no_value_completion = completion_signatures<>;
+        using completion_signatures =
+            pika::execution::experimental::transform_completion_signatures_of<std::decay_t<Sender>,
+                pika::execution::experimental::empty_env,
+                pika::execution::experimental::completion_signatures<
+                    pika::execution::experimental::set_value_t(),
+                    pika::execution::experimental::set_error_t(std::exception_ptr),
+                    // this also gets added by the default SetDone template parameter
+                    pika::execution::experimental::set_stopped_t()>,
```
msimberg, line 69:
```suggestion
                    pika::execution::experimental::set_error_t(std::exception_ptr)>,
```

libs/pika/executors/tests/unit/std_thread_scheduler.cpp
```diff
@@ -24,7 +24,7 @@ struct custom_type_non_default_constructible_non_copyable
     int x;
     custom_type_non_default_constructible_non_copyable() = delete;
     explicit custom_type_non_default_constructible_non_copyable(int x)
-      : x(x){};
+      : x(x) {};
```
msimberg, line 27: I think this will need reverting to satisfy clang-format. Same in `thread_pool_scheduler.cpp`.

msimberg (changes requested):

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -57,10 +57,17 @@ namespace pika::mpi::experimental::detail {
         std::decay_t<F> f;
 
 #if defined(PIKA_HAVE_STDEXEC)
-
-        using completion_signatures = pika::execution::experimental::completion_signatures<
-            pika::execution::experimental::set_value_t(MPI_Request)>;
-
+        template <typename...>
+        using no_value_completion = completion_signatures<>;
```
msimberg, line 61:
```suggestion
        using no_value_completion = pika::execution::experimental::completion_signatures<>;
```

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
biddisco: in what way?

msimberg: Sorry, should've marked more context. Look at line 87. The return statement is duplicated.

libs/pika/executors/tests/unit/std_thread_scheduler.cpp
biddisco: when I merged, I got clang format errors in code unrelated to mine and I clang-formatted it anyway. I guess I should have ignored it

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
biddisco: oops. Edited it in conflict editor, obviously saved it badly. fixed

msimberg (approved):

PR #1188, opened 2024-06-24
Add docs requirements file to install the dependencies directly through pip

aurianer:


msimberg (commented):
Thanks! Would you mind updating the docs CI workflow to use this to use `requirements.txt` instead so that we know it's up to date?

.github/workflows/docs.yml
```diff
@@ -34,7 +34,7 @@ jobs:
                 --allow-remove-essential \
                 --allow-change-held-packages \
                 doxygen python3 python3-pip
-            pip install breathe recommonmark sphinx sphinx-material
+            pip install $(cat $PWD/docs/requirements.txt)
```
msimberg, line 37: Is this more idiomatic:
```suggestion
            pip install --requirement $PWD/docs/requirements.txt
```
?

docs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+sphinx
```
msimberg, line 1:
```suggestion
# noinshpect
sphinx
```
? Not sure what `pip` thinks of the comment...

PR #1189, opened 2024-06-24
Add documentation for `split_tuple`

msimberg:
Adds `split_tuple` to the API reference under `pika/execution.hpp`. Includes the doxygen docstring and a small example program that uses `split_tuple`.

I'm starting to document the `pika::execution` functionality that is not available in P2300 or stdexec, since documentation for what is in P2300 can be looked up elsewhere.

### DIFF
--- a/examples/documentation/split_tuple_documentation.cpp
+++ b/examples/documentation/split_tuple_documentation.cpp
@@ -1,0 +1,47 @@
+//  Copyright (c) 2024 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <pika/execution.hpp>
+#include <pika/init.hpp>
+
+#include <fmt/printf.h>
+
+#include <chrono>
+#include <thread>
+#include <tuple>
+#include <utility>
+
+int main(int argc, char* argv[])
+{
+    namespace ex = pika::execution::experimental;
+    namespace tt = pika::this_thread::experimental;
+
+    pika::start(argc, argv);
+    ex::thread_pool_scheduler sched{};
+
+    // split_tuple can be used to process the result and its square through
+    // senders, without having to pass both around together
+    auto [snd, snd_squared] = ex::schedule(sched) | ex::then([]() { return 42; }) |
+        ex::then([](int x) { return std::tuple(x, x * x); }) | ex::split_tuple();
+
+    // snd and snd_squared will be ready at the same time, but can be used independently
+    auto snd_print =
+        std::move(snd) | ex::transfer(sched) | ex::then([](int x) { fmt::print("x is {}\n", x); });
+    auto snd_process = std::move(snd_squared) | ex::transfer(sched) | ex::then([](int x_squared) {
+        fmt::print("Performing expensive operations on x * x\n");
+        std::this_thread::sleep_for(std::chrono::milliseconds(300));
+        return x_squared / 2;
+    });
+
+    auto x_squared_processed =
+        tt::sync_wait(ex::when_all(std::move(snd_print), std::move(snd_process)));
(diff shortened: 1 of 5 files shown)


msimberg:
The current version of the documentation looks like this:
2024-06-24-175434_1098x1256_scrot

Questions in particular to the non-pika devs:
- Is it confusing that it's documented as a variable (it's a callable function object with variable name `split_tuple`)?
- Is it clear enough to call it "sender adaptor" to explain how it can be used?
- Do we need to add a "function signature", either manually, or simply document it as a function even though it's not one?
- Is the example too trivial? Or too complicated?

Whatever other feedback you have is of course also welcome.

examples/documentation/split_tuple_documentation.cpp
```diff
@@ -0,0 +1,46 @@
...
+    auto snd_print =
+        std::move(snd) | ex::transfer(sched) | ex::then([](int x) { fmt::print("x is {}\n", x); });
+    auto snd_process = std::move(snd_squared) | ex::transfer(sched) | ex::then([](int x) {
+        fmt::print("Performing expensive operations on x * x\n");
+        std::this_thread::sleep_for(std::chrono::milliseconds(300));
+        return x / 2;
+    });
+
+    auto x = tt::sync_wait(ex::when_all(std::move(snd_print), std::move(snd_process)));
+    fmt::print("The final result is {}\n", x);
```
RMeli, line 40:
```suggestion
    auto snd_process = std::move(snd_squared) | ex::transfer(sched) | ex::then([](int x2) {
        fmt::print("Performing expensive operations on x * x\n");
        std::this_thread::sleep_for(std::chrono::milliseconds(300));
        return x2 / 2;
    });

    auto x2_processed = tt::sync_wait(ex::when_all(std::move(snd_print), std::move(snd_process)));
    fmt::print("The final result is {}\n", x2_processed);
```

I think the example is very clear, but the name of the variables can be improved for readability instead of using `x` everywhere?

albestro:
> Is it confusing that it's documented as a variable (it's a callable function object with variable name split_tuple)?

Let me share my secret: when I look to complex things like this, which have a lot of internal details and the signature is generally not straightforward to look at, I always start from looking at the example/description.

> Is it clear enough to call it "sender adaptor" to explain how it can be used?

I don't know if it is a term used in P2300 or elsewhere, but I think it fits nicely and explains well its role.

> Do we need to add a "function signature", either manually, or simply document it as a function even though it's not one?

See above, IMHO it would not add anything.

> Is the example too trivial? Or too complicated?

I think the example is just perfect, simple and straight to the point.

examples/documentation/split_tuple_documentation.cpp
msimberg: I changed it to `x_squared` and `x_squared_processed` to be even more explicit. Does that look ok to you?

msimberg:
Thanks @albestro and @RMeli for the detailed feedback, it's very useful! In addition to trying to link to other resources in the documentation that explain the background, I think we can probably try to document the fact that the sender algorithms are implemented as CPOs (including how they're customized). This would hopefully make it clearer why these are variables, instead of functions. However, this is for a separate PR.

In addition, yes, "sender adaptor" is a term used in P2300 (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r9.html#design-sender-adaptors):

> A [sender adaptor](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r9.html#sender-adaptor) is an algorithm that takes one or more senders, which it may execution::connect, as parameters, and returns a sender, whose completion is related to the sender arguments it has received.A [sender adaptor](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r9.html#sender-adaptor) is an algorithm that takes one or more senders, which it may execution::connect, as parameters, and returns a sender, whose completion is related to the sender arguments it has received.

`split_tuple` strictly returns a tuple of one or more senders, instead of just a single sender, but the intent is close enough that I think using "sender adaptor" helps, at least if one has seen the term before. We could also do better to link to explanations of sender adaptors, factories, and consumers.

PR #1190, opened 2024-06-24
Make function signatures, class names etc. bold in API documentation

msimberg:


msimberg:
Makes the API documentation look like this:
2024-06-24-180342_1044x1106_scrot



instead of this:

2024-06-24-180111_1062x1109_scrot

I find this makes it a lot easier to distinguish where one documented item starts. What do you think?

msimberg:
Merging manually since this only modifies documentation.

PR #1192, opened 2024-06-25
Capture more operation states and receivers by reference in sender adaptors to use SBO of `unique_function`

msimberg:


libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -111,52 +111,54 @@ namespace pika::mpi::experimental::detail {
...
                             PIKA_DETAIL_DP(mpi_tran<5>, debug(str<>("set_value")));
-                            ex::set_value(PIKA_MOVE(receiver));
+                            ex::set_value(PIKA_MOVE(op_state.receiver));
                         });
                     ex::start_detached(PIKA_MOVE(snd0));
                 }
             },
-            request);
+            op_state.request);
     }
```
msimberg, line 147: @biddisco in #1128 I completely missed that the forwards here were inside the lambda already (I blindly looked at the function signature). This rearranges the whole function quite a bit anyway, so would you mind looking it over to see if it's fine for you?

PR #1202, opened 2024-07-11
Use static libraries on a CI configuration without APEX enabled

msimberg:
I suspect that something related to static libraries are causing problems on the APEX CI configuration: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/7312229099#L133. This PR moves `BUILD_SHARED_LIBS=OFF` to another CI configuration to see if that's the case.


msimberg:
The two affected pipelines are now behaving ok in CI, merging manually.

PR #1140, opened 2024-05-31
Use valgrind in CUDA CI configuration

msimberg:
Fixes #802.

PR #1205, opened 2024-07-24
Update stdexec commit used in CI to `e99073de4c4b924804f54d7f52bc4e9fbcf53b92`

msimberg:
stdexec now `static_assert`s if a scheduler has a `schedule` member that doesn't return a sender, so we can't check with `is_scheduler_v` those types of non-schedulers (as one test does).

`make_completion_signatures` is also buggy in the new commit, so I've moved uses of `make_completion_signatures` over to the new `transform_completion_signatures_of` (the former was renamed to the latter in P2300R8).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ setup_arm64_machine: &setup_arm64_machine
 
 docker_default: &docker_default
   docker:
-    - image: pikaorg/pika-ci-base:26
+    - image: pikaorg/pika-ci-base:27
 
 defaults: &defaults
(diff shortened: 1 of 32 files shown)


msimberg:
This and https://github.com/pika-org/pika/pull/1140 have issues building the HIP stack with the newer spack commit. Further investigation needed.

msimberg:
We still need https://github.com/NVIDIA/stdexec/pull/1398 to get things going with the newer commits of stdexec.

msimberg:
The nvhpc pipeline isn't happy now. This needs more time...

PR #1210, opened 2024-07-24
Separate x86 and ARM CSCS CI pipelines to be completely independent

msimberg:
Also sets `SPACK_REPO` as a build argument in the spack base image stage (the default in the Dockerfile was the same as the value we were setting in the configuration, so no change is expected from this change).


msimberg:
I'm going to go ahead with this as this allows us to slightly better ignore the unavailable ARM configurations. @aurianer would you mind having a quick look to see if this seems ok to you?

PR #1212, opened 2024-07-26
Increase curl and job timeouts and retries in CSCS CI

msimberg:
Also remove the hardcoded `NUM_PROCS` from the CI configuration.


msimberg:
Small update:
- I initially added `retry-all-errors` to the curl configuration, but this ends up retrying everything, including 404s. In some packages, spack will first try a few URLs which may have a cached tarball but which isn't guaranteed to contain anything. These failures are guaranteed to add n * `retry-max-time` delays to each package. I initially set the `retry-max-time` to quite a large value "to be on the safe side" (120), which resulted in ~20 * 2 * 120 s (~80 minutes) delays in package downloads in a single pipeline, which is a bit excessive. A 30 second timeout would reduce it to ~20 minutes of "unnecessary" waiting. Better, but that's still quite bad.
- Without `retry-all-errors` we still get occasional fetch errors. I'm waiting to see if an update in the proxy that is used on rosa will help improve the situation since I'd prefer not to use `retry-all-errors` if possible.

msimberg:
While this doesn't help directly with the most common `FetchError`s we see in spack pipelines (the http proxy on compute nodes still fails despite these retries), these changes shouldn't hurt and the non-retry changes are useful to get in in any case.

PR #1213, opened 2024-07-26
Enable `detect_stack_use_after_return` option in asan CI configuration

msimberg:
This was previously problematic. Testing to see what CI says.


msimberg:
This is as problematic as before, closing.

PR #1217, opened 2024-08-14
Add documentation for `when_all_vector`

msimberg:
I should note that writing this highlighted to me the design decision to make `when_all_vector` require that senders send exactly zero or one values. This need not be the case in general, and we can consider changing `when_all_vector` or adding a new algorithm to instead send a `vector<tuple<Ts...>>` or even `vector<variant<tuple<T1s...>, tuple<T2s...>, ...>>` as more generic versions of the current implementation. This is similar to how there's a `sync_wait` and `sync_wait_with_variant` in P2300. I will not change the behaviour here, just highlight the possibility for the future.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -80,3 +80,9 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
 .. literalinclude:: ../examples/documentation/split_tuple_documentation.cpp
    :language: c++
    :start-at: #include
+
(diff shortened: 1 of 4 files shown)


msimberg:
I initially changed `print("...\n")` to `println("...")` in the documentation examples, but changed it back since `println` was added to fmt only in version 10, and we currently still support version 9.

RMeli (approved):
LGTM, the examples are clear.

albestro (approved):
Looks clear to me.

PR #1218, opened 2024-08-14
Add GCC 12 and HIP 6 CI configuration

### LINKED ISSUE #1164: Add HIP 6 CI configuration
I think it makes sense to add a new CI configuration with HIP 6 and keep the existing 5.X configuration, so that we cover all supported major versions.

msimberg:
Fixes #1164. The newly added GCC 12/HIP 6 configuration is now run on pushes and replaces the old GCC 12/HIP 5 which is still run on merges.

GCC (libstdc++) 13 and newer are incompatible with HIP for now: https://github.com/spack/spack/pull/46291.

### DIFF
--- a/.gitlab/pipelines_on_push.yml
+++ b/.gitlab/pipelines_on_push.yml
@@ -7,5 +7,5 @@
 include:
   - local: '.gitlab/includes/gcc13_pipeline.yml'
   - local: '.gitlab/includes/clang14_cuda11_pipeline.yml'
-  - local: '.gitlab/includes/gcc12_hip5_pipeline.yml'
+  - local: '.gitlab/includes/gcc12_hip6_pipeline.yml'
   - local: '.gitlab/includes/sloc.yml'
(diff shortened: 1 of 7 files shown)


.gitlab/pipelines_on_push.yml
```diff
@@ -5,8 +5,10 @@
...
-  - local: '.gitlab/includes/performance_gcc13_pipeline.yml'
-  - local: '.gitlab/includes/gcc12_hip5_pipeline.yml'
-  - local: '.gitlab/includes/sloc.yml'
+  # - local: '.gitlab/includes/gcc13_pipeline.yml'
+  # - local: '.gitlab/includes/clang14_cuda11_pipeline.yml'
+  # - local: '.gitlab/includes/performance_gcc13_pipeline.yml'
+  # - local: '.gitlab/includes/gcc12_hip5_pipeline.yml'
+  # TODO: Only on merge
+  - local: '.gitlab/includes/gcc14_hip6_pipeline.yml'
+  # - local: '.gitlab/includes/sloc.yml'
```
msimberg, line 14: To do.

msimberg:
The GCC14/HIP conflict seems to be the same issue as reported here: https://github.com/llvm/llvm-project/issues/69288. That one is about clang in CUDA mode, but the error looks exactly the same otherwise. I will add some comments in the code and a conflict if GCC 12 seems to be working fine.

msimberg:
I've added some sort of conflict in upstream spack: https://github.com/spack/spack/pull/46291.

PR #1224, opened 2024-08-26
Use sockets, not NUMA nodes, when dealing with CPU masks

### LINKED ISSUE #1223: Out-of-bounds access on CPU mask on Grace CPU with hwloc 2.11
On Grace CPUs with hwloc 2.11 it seems like pika thinks there are 8 NUMA nodes each with 72 cores, which is false (there are only 4). This may be related to the following in the hwloc release notes (https://raw.githubusercontent.com/open-mpi/hwloc/v2.11/NEWS):
> Don't hide the GPU NUMA node on NVIDIA Grace Hopper.

though I'm not 100% sure yet.

msimberg:
Still a draft. Tested on Grace-Hopper system to give correct results. This is only the minimal change needed which changes uses of topology `*numa_node*` functionality to `*socket*`.

Fixes #1223.

When dealing with thread affinity, this now uses sockets instead of numa nodes. This also means that the following changes:
- the resource partitioner helper object `resource::numa_domain` is now called `resource::socket` instead, including all related variables
- the `--pika:bind` option `numa` is now called `socket`

The following hasn't changed:
- The `numa-balanced` mapping is still called `numa-balanced`. Should we rename it `socket-balanced`?
- The `numa-sensitive` command line/config option is still called `numa-sensitive`. I think this still makes sense (even though many schedulers don't really take this into account?).
- The shared priority queue scheduler still deals with numa domains.

In short, when talking about affinity and going from socket/numa to pu, we now use sockets exclusively because numa to pu may be many to one, unlike socket to pu. When going the other way around, i.e. pu to socket/numa, we can still talk about pu to numa to understand if two pus belong to the same numa domain(s).

### DIFF
--- a/libs/pika/affinity/src/affinity_data.cpp
+++ b/libs/pika/affinity/src/affinity_data.cpp
@@ -161,12 +161,12 @@ namespace pika::detail {
             // pu_num.
             return topo.get_core_affinity_mask(pu_num);
         }
-        if (0 == std::string("numa").find(affinity_domain_))
+        if (0 == std::string("socket").find(affinity_domain_))
         {
-            // The affinity domain is 'numa', return a bit mask corresponding
-            // to all processing units of the NUMA domain containing the
+            // The affinity domain is 'socket', return a bit mask corresponding
+            // to all processing units of the socket containing the
             // given pu_num.
-            return topo.get_numa_node_affinity_mask(pu_num);
+            return topo.get_socket_affinity_mask(pu_num);
         }
 
         // The affinity domain is 'machine', return a bit mask corresponding
--- a/libs/pika/affinity/src/parse_affinity_options.cpp
+++ b/libs/pika/affinity/src/parse_affinity_options.cpp
@@ -307,67 +307,67 @@ namespace pika::detail {
 
         num_pus.resize(num_threads);
 
-        // numa nodes
-        std::size_t num_numas = (std::max)(std::size_t(1), t.get_number_of_numa_nodes());
-        std::vector<std::size_t> num_cores_numa(num_numas, 0);
-        std::vector<std::size_t> num_pus_numa(num_numas, 0);
-        std::vector<std::size_t> num_threads_numa(num_numas, 0);
-        for (std::size_t n = 0; n < num_numas; ++n)
+        // sockets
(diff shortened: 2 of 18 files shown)


msimberg:
@aurianer, @biddisco could you please have a look at the changes here to see if they make sense to you? In particular, note the points in the PR description. Would you like to see more mentions of numa turn into socket?

msimberg:
> I'm not sure what you mean by more mentions but I would just mention it in the release notes of the next release

I just meant any place in the source code where `numa` appears in the source code, be it a variable name, class name, string etc. For example, `"numa-balanced"`.

Definitely mentioning this in the release notes.

biddisco:
I just looked through this and can see what is essentially a search/replace for numa with socket - but  what I don't see is how this actually changes the behaviour of the code - There must be at least one line where the hwloc description of the hardware is inspected for socket info - which is different from numa info -  where is it?

msimberg:
> I just looked through this and can see what is essentially a search/replace for numa with socket - but what I don't see is how this actually changes the behaviour of the code - There must be at least one line where the hwloc description of the hardware is inspected for socket info - which is different from numa info - where is it?

Yep, it's indeed more or less just a search and replace. The (first) place that breaks if we use numa domains instead of sockets is https://github.com/pika-org/pika/blob/89f533772b9e076410f92a666efdecde9ec05592/libs/pika/resource_partitioner/src/detail_partitioner.cpp#L240-L296. Specifically, https://github.com/pika-org/pika/blob/89f533772b9e076410f92a666efdecde9ec05592/libs/pika/resource_partitioner/src/detail_partitioner.cpp#L274 fails with the 289th `pid` (i.e. `pid = 288`) because the process mask for the full machine only has 288 cores. This happens when considering the 5th NUMA domain, which hwloc reports also has 72 associated cores.

The reason using sockets instead of NUMA domains is that we iterate through 4 sockets of 72 cores each instead of 8 NUMA domains of 72 cores each. It is possible that hwloc should report either 0 or 288 associated cores for the newly added 4 GPU NUMA domains. In the former case we might be fine, but it's still semantically wrong to assume that any cores associated with a NUMA domain is exclusively associated with a single NUMA domain. With a socket (or package as hwloc also calls it), a core always belongs to a single socket, so we don't get any double-counting.

Does this make more sense?

biddisco (approved):
LGTM

PR #1229, opened 2024-08-30
Adapt to `FindBoost` deprecation with CMake 3.30

### LINKED ISSUE #1222: Update `pika_setup_boost.cmake` to use Boost's built-in CMake configuration
CMake 3.30 deprecates CMake's FindBoost modules and attempts to use Boost's CMake configuration. We can fall back to the old policy temporarily, but long term we should prefer to use Boost's own configuration. See https://cmake.org/cmake/help/latest/module/FindBoost.html.

I briefly attempted to use Boost's own configuration in a spack install tree, but curiously I could not see a `BoostConfig.cmake` file in the install tree, so we'd need to check if it's just my installation that was missing it or if we need to take that into account more generally.

aurianer:
Fixes #1222

We require Boost version >= 1.71 in pika, `BoostConfig.cmake` is available in boost installations since 1.70, so only changing module to config was necessary.

### DIFF
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -8,7 +8,7 @@ include:
   - local: '.gitlab/includes/common_pipeline.yml'
 
 variables:
-  SPACK_COMMIT: b9e4e98f15b4a07814adf8b22da19e631f2749e4
+  SPACK_COMMIT: develop-2024-09-01
 
 .base_spack_image:
   timeout: 1 hours
--- a/.gitlab/includes/common_spack_pipeline.yml
+++ b/.gitlab/includes/common_spack_pipeline.yml
@@ -8,7 +8,7 @@ include:
   - local: '.gitlab/includes/common_pipeline.yml'
 
 variables:
-  SPACK_COMMIT: develop-2024-08-18
+  SPACK_COMMIT: b9e4e98f15b4a07814adf8b22da19e631f2749e4
 
 .base_spack_image:
   timeout: 1 hours
--- a/cmake/pika_setup_boost.cmake
+++ b/cmake/pika_setup_boost.cmake
@@ -32,7 +32,7 @@ if(NOT TARGET pika_dependencies_boost)
   set(Boost_NO_BOOST_CMAKE ON) # disable the search for boost-cmake
 
   # Find the headers and get the version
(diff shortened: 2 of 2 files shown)


msimberg (approved):
Thanks! We'll need a newer commit of spack which contains https://github.com/spack/spack/pull/46062, but otherwise this is ok to merge.

cmake/pika_setup_boost.cmake
```diff
@@ -46,7 +46,8 @@ if(NOT TARGET pika_dependencies_boost)
 
   list(REMOVE_DUPLICATES __boost_libraries)
 
-  find_package(Boost ${Boost_MINIMUM_VERSION} MODULE REQUIRED COMPONENTS ${__boost_libraries})
+  pika_set_cmake_policy(CMP0167 NEW)  # deprecates FindBoost
+  find_package(Boost ${Boost_MINIMUM_VERSION} CONFIG REQUIRED COMPONENTS ${__boost_libraries})
```
msimberg, line 50: I actually see we still have a `find_package(Boost)` without an explicit `MODULE` or `CONFIG` above this on line 35. Should we be explicit about `CONFIG` there as well, or remove `CONFIG` from both?

.gitlab/includes/common_spack_pipeline.yml
```diff
@@ -8,7 +8,7 @@ include:
   - local: '.gitlab/includes/common_pipeline.yml'
 
 variables:
-  SPACK_COMMIT: develop-2024-08-18
+  SPACK_COMMIT: b9e4e98f15b4a07814adf8b22da19e631f2749e4
```
msimberg, line 11: Updating to the latest tag with a date, so it's easier to see how old the tag is.

```suggestion
  SPACK_COMMIT: develop-2024-09-01
```

msimberg:
There seems to be a concretization issue on the nvhpc pipeline with the new spack version: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/7742931167. I'm going to risk merging this to get the next release out with the assumption that the error is simply some overconstraint in spack and that there's no issue in the pika build with that pipeline once we get it working.

msimberg:
The clang pipeline failure also needs https://github.com/spack/spack/pull/46200.

PR #1230, opened 2024-09-02
Simplify `pika_add_nvhpc_flags`

aurianer:


cmake/pika_setup_cuda.cmake
```diff
@@ -143,19 +143,15 @@ if(PIKA_WITH_CUDA AND NOT TARGET pika_internal::cuda)
...
-    set(source_compile_flags)
-  endif()
-
   set(source_gpu_cc_flags)
   foreach(cuda_arch ${CMAKE_CUDA_ARCHITECTURES})
     set(source_gpu_cc_flags "${source_gpu_cc_flags} -gpu=cc${cuda_arch}")
   endforeach()
 
+  get_source_file_property(source_compile_flags ${source} COMPILE_FLAGS)
+  string(REPLACE "NOTFOUND" "" source_compile_flags ${source_compile_flags})
```
msimberg, line 152: Highly unlikely, but in principle the flags could have something like `-DBLAH_NOTFOUND_123` and the `NOTFOUND` would get removed. Perhaps do a regex with `"^NOTFOUND$"` or simply keep the explicit `if` for this?

msimberg (approved):
Thanks, this looks good!

PR #1231, opened 2024-09-04
Bisect NVHPC pipeline concretization failure

msimberg:
Bisecting https://github.com/pika-org/pika/pull/1229#issuecomment-2329099051.


msimberg:
https://github.com/spack/spack/pull/45784 is the PR that broke the build. Essentially it stops from finding NVHPC as a compiler, and thus concretization fails later. It's not clear to me why it doesn't find NVHPC anymore with that change.

PR #1239, opened 2024-09-10
Move GPU CI pipelines from old daint to new daint

msimberg:
Moves the GPU CSCS CI pipelines from Daint XC to daint on Alps.

I've had to make some small changes to accommodate the new system:
- I've disabled valgrind testing on the CUDA pipeline. I've instead moved it to a CPU-only pipeline that runs on x86. Valgrind was reporting errors that I can't tell yet if they're simply false positives on ARM or not. #1367 tracks reenabling valgrind on CUDA builds.
- I've removed the minor version restriction on the CUDA 11 pipelines. At least with CUDA 11.5 and older, when the tests are run on GH200, the first test that uses a GPU takes significantly longer to run (100s of seconds). With 11.8 this is no longer a problem, so I now only require `cuda@11` in the specs which results in 11.8. This also means that we don't have to use `+allow-unsupported-compilers` for the clang+CUDA configuration.
- I install gfortran in the base image. On the nvhpc 24.9 pipeline we use gcc@13.3.0 as the base compiler. This happens to be the system compiler as well, and installing nvhpc requires a fortran compiler. This was not available unless we explicitly install the `gfortran` package.
- I've patched some incompatibilities with CUDA that happen only on ARM. Where the patch is applied, I've supplied a number of links to the issue. It seems like CUDA 12.6 no longer requires patching, but I think it's good if we're able to test with multiple versions of CUDA 12. Right now we only test with CUDA 12.0 on the plain CUDA configuration and CUDA 12.4 on the nvhpc pipelines. Once we drop the CUDA 11 pipelines we can test a few more different versions of CUDA 12.
- I've removed the `gcc13_santis` pipeline since we have enough other pipelines that run on ARM now.
- The GCC 12/CUDA 12 pipeline now also runs tests.
- Failures are no longer allowed on the ARM pipelines.


.gitlab/includes/clang14_cuda11_pipeline.yml
```diff
@@ -28,25 +28,25 @@ include:
...
     - .variables_clang14_cuda11_config
-    - .build_template_rosa
+    - .build_template_gh200
 
 .clang14_cuda11_test_common:
   needs: [clang14_cuda11_build]
   extends:
     - .variables_clang14_cuda11_config
-    - .test_common_gpu_daint_cuda
+    - .test_common_gh200
```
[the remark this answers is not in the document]
msimberg: Not 100% sure, since the cluster name might change and we don't really care which one it runs on as long as it has Grace-Hopper? We do need to mention the cluster name of course in the base templates `*_gh200`, but I'm not sure (yet) that the exact cluster is a relevant distinction for the different configurations. If we find out that it is worth running pipelines on different gh200 clusters we can introduce the distinction at that point? What do you think?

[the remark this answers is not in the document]
msimberg: I followed your advice in the end, since we call the other test template `test_common_eiger_mc`, i.e. we include the cluster name there. So I've included `daint` in the Grace-Hopper template as well.

.gitlab/pipelines_on_push.yml
```diff
@@ -9,3 +9,6 @@ include:
   - local: '.gitlab/includes/clang14_cuda11_pipeline.yml'
   - local: '.gitlab/includes/gcc12_hip6_pipeline.yml'
   - local: '.gitlab/includes/sloc.yml'
+  # TODO: move to on_merge before merging
```
msimberg, line 12: To do.

msimberg:
All right, we're making some progress:

- The GCC 12/CUDA 12 pipeline is working now.
- The clang/cuda one isn't working because it uses valgrind and valgrind doesn't seem to like some of the instructions used. I'll see if using a less specific arm instruction set may help valgrind here.
- The CUDA 11 pipeline isn't working. I would've expected it to be compatible despite the old CUDA version. I'll see if changing required driver versions etc. helps at all.

I may end up disabling the test steps for the latter two in this PR to reenable them in separate PRs.

.gitlab/includes/gcc13_gh200_pipeline.yml
```diff
@@ -8,7 +8,7 @@ include:
   - local: '.gitlab/includes/common_pipeline.yml'
```
msimberg, line 8: Remove pipeline?

msimberg:
The clang/cuda configuration with valgrind no longer complains about illegal instructions: good. It now reports many issues, which I don't know yet if they're real or not.

I'll aim to get the GCC 12/CUDA 12 pipeline running properly (still some tweaks needed on the CSCS CI side apparently) and then I'll attempt to revive the two other CUDA configurations separately, possibly introducing another valgrind configuration on x86.

.gitlab/docker/Dockerfile.spack_build
```diff
@@ -54,3 +54,7 @@ RUN CTEST_XML=$PWD/ctest.xml; \
             --no-compress-output \
             -R tests \
             --no-tests=error"
+
+RUN apt-get update && apt-get -yqq install --no-install-recommends \
+    gdb \
+  && rm -Rf /var/lib/apt/lists/*
```
msimberg, line 60: To do: move to separate PR.

.gitlab/includes/clang14_cuda11_pipeline.yml
```diff
@@ -10,43 +10,43 @@ include:
...
 .variables_clang14_cuda11_config:
   variables:
-    SPACK_ARCH: linux-ubuntu22.04-haswell
+    SPACK_ARCH: linux-ubuntu22.04-aarch64
     COMPILER: clang@14.0.6
     CXXSTD: 17
-    GPU_TARGET: '60'
+    GPU_TARGET: '80'
     SPACK_SPEC: "pika@main arch=$SPACK_ARCH %${COMPILER} +cuda cuda_arch=${GPU_TARGET} malloc=system
-      cxxstd=$CXXSTD +valgrind ^boost@1.79.0 ^cuda@11.5.0 +allow-unsupported-compilers ^hwloc@2.7.0 ^valgrind ~mpi"
```
msimberg, line 18: To do: open an issue about reenabling CUDA+valgrind in CI.

.gitlab/docker/Dockerfile.spack_build
msimberg: Done in https://github.com/pika-org/pika/pull/1357.

.gitlab/includes/common_pipeline.yml
```diff
@@ -36,19 +36,10 @@ variables:
   variables:
     SLURM_PARTITION: normal
 
-.test_common_gpu_daint_cuda:
+.test_common_daint_gh200:
   extends:
-    - .container-runner-daint-gpu
+    - .container-runner-todi-gh200 # TODO: daint? rename template?
```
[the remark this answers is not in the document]
msimberg: Replace with `.container.runner-daint-gh200` once it's available.

msimberg:
This is now ready for merging for my perspective. I've updated the PR description to outline the changes.

msimberg:
I think we might need longer timeouts on the new valgrind CI configuration. Will try adjusting that.

.gitlab/pipeline.yml
```diff
@@ -7,5 +7,5 @@
 include:
   - local: '.gitlab/pipelines_on_push.yml'
   - local: '.gitlab/pipelines_on_merge.yml'
-    rules:
-      - if: $CI_COMMIT_BRANCH == "main" || $CI_COMMIT_BRANCH =~ /^gh-readonly-queue/ || $CI_COMMIT_BRANCH =~ /^release/
+    # rules:
+    #   - if: $CI_COMMIT_BRANCH == "main" || $CI_COMMIT_BRANCH =~ /^gh-readonly-queue/ || $CI_COMMIT_BRANCH =~ /^release/
```
msimberg, line 11: Uncomment before merging.

msimberg:
Merging manually. The attempt at merging through the merge queue failed for unrelated reasons.

PR #1242, opened 2024-09-10
Don't use gperftools or jemalloc in apex when used together with pika

msimberg:
Investigating whether this may be causing issues in CI...


msimberg:
This doesn't seem to have any effect on the apex CI configuration, closing.

PR #1248, opened 2024-09-13
Clean up `mpi_base` module

biddisco:
Remove unused features/functions/variables.

Most of mpi_base was brought in from HPX which needs MPI for it's internal communication (active messages) and it must therefore guarantee initialization and certain setup defaults whenever it might be used in a multi node environment.

pika only uses MPI when the user explicitly requests MPI and therefore the setup/teardown is always directed by the user and we can remove some unnecessary code/checks.

Replace some duplicated mpi initialized/rank fetch calls in pika body with reused functions from mpi_environment.

### DIFF
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -24,7 +24,6 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
 
 #include <exception>
 #include <tuple>
@@ -151,9 +150,10 @@ namespace pika::mpi::experimental::detail {
                             {
                                 PIKA_DETAIL_DP(mpi_tran<5>,
                                     debug(str<>("set_error"), "status != MPI_SUCCESS",
-                                        detail::error_message(status)));
+                                        mpi::detail::error_message(status)));
                                 ex::set_error(PIKA_MOVE(r.op_state.receiver),
-                                    std::make_exception_ptr(mpi_exception(status)));
+                                    std::make_exception_ptr(
+                                        mpi::exception(status, "dispatch mpi")));
                                 return;
                             }
                             // early poll just in case the request completed immediately
--- a/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
@@ -24,6 +24,8 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
+#include <pika/mpi_base/mpi.hpp>
+#include <pika/mpi_base/mpi_exception.hpp>
 
 #include <exception>
 #include <tuple>
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -24,7 +24,6 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
 #include <pika/runtime/runtime.hpp>
 
 #include <exception>
@@ -85,7 +84,7 @@ namespace pika::mpi::experimental::detail {
         else
         {
             ex::set_error(PIKA_FORWARD(Receiver, receiver),
-                std::make_exception_ptr(mpi_exception(mpi_status)));
+                std::make_exception_ptr(mpi::exception(mpi_status, "set_error handler")));
         }
     }
 
@@ -125,7 +124,8 @@ namespace pika::mpi::experimental::detail {
                 if (status != MPI_SUCCESS)
                 {
                     ex::set_error(PIKA_MOVE(op_state.receiver),
-                        std::make_exception_ptr(mpi_exception(status)));
+                        std::make_exception_ptr(
+                            mpi::exception(status, "new_task_request_callback")));
                 }
                 else
                 {
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
@@ -24,6 +24,8 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
+#include <pika/mpi_base/mpi.hpp>
+#include <pika/mpi_base/mpi_exception.hpp>
 #include <pika/runtime/runtime.hpp>
 
 #include <exception>
--- a/libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
@@ -25,7 +25,6 @@
 #include <pika/execution_base/sender.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
 
 #include <exception>
 #include <tuple>
--- a/libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
@@ -25,6 +25,7 @@
 #include <pika/execution_base/sender.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
+#include <pika/mpi_base/mpi.hpp>
 
 #include <exception>
 #include <tuple>
--- a/libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
@@ -24,7 +24,7 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi_exception.hpp>
+#include <pika/mpi_base/mpi.hpp>
 #include <pika/synchronization/condition_variable.hpp>
 
 #include <exception>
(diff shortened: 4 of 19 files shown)


biddisco:
do not review this yet.

biddisco:
fixed a bunch of clang format things and removed commented out code

biddisco:
fixed a bunch of clang format things and removed commented out code

biddisco:
Apparently, I do not have permission to reopen the PR

biddisco:
clever. it has reopened, then I forced pushed the cleaned version

biddisco:
I also want to move mpi_exception stuff into the mpi_base module, but there's no point doing that until the other mpi stuff has been merged because [pause] ... conflicts

libs/pika/async_mpi/include/pika/async_mpi/dispatch_mpi.hpp
```diff
@@ -24,7 +24,6 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
```
msimberg, line 27: Why was this removed? This file still uses `MPI_*` names directly.

libs/pika/async_mpi/include/pika/async_mpi/mpi_helpers.hpp
```diff
@@ -24,7 +24,6 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
```
msimberg, line 27: Why removed?

libs/pika/async_mpi/include/pika/async_mpi/transform_mpi.hpp
```diff
@@ -25,7 +25,6 @@
 #include <pika/execution_base/sender.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
```
msimberg, line 28: Why removed?

libs/pika/async_mpi/include/pika/async_mpi/trigger_mpi.hpp
```diff
@@ -25,7 +24,7 @@
 #include <pika/executors/thread_pool_scheduler.hpp>
 #include <pika/functional/detail/tag_fallback_invoke.hpp>
 #include <pika/functional/invoke.hpp>
-#include <pika/mpi_base/mpi.hpp>
```
msimberg, line 28: Why removed?

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
```diff
@@ -19,59 +19,38 @@
...
-            static bool enabled_;
-            static bool has_called_init_;
-            static int provided_threading_flag_;
-            static MPI_Comm communicator_;
-
-            static int is_initialized_;
-        };
-}}    // namespace pika::util
+namespace pika::mpi {
+    struct PIKA_EXPORT environment
```
msimberg, line 23: If we're changing the namespace now, I think we should not count this as part of the public API. This is really an internal helper, right? Could you move this to `pika::mpi::detail`?

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
```diff
@@ -19,59 +19,38 @@
...
+    {
+        // This function checks a list of environment variables to see if commonly set
+        // ones such as mpich, openmpi, slurm, etc are present.
+        // The list of env vars can be set via pika command line params,
+        // or defaults used from cmake configure time.
+        // Although this function is not currently used in pika,
+        // it is left for potential future use
+        static bool check_environment(util::runtime_configuration const& cfg);
+
```
msimberg, line 32: Can we remove this and add this back if/when it's needed in the future instead?

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
```diff
@@ -81,12 +60,12 @@ namespace pika { namespace util {
...
-        {
-            static bool check_mpi_environment(runtime_configuration const& cfg);
-        };
-}}    // namespace pika::util
+namespace pika::mpi {
+    struct PIKA_EXPORT environment
+    {
+        static bool check_environment(runtime_configuration const& cfg);
+    };
+}    // namespace pika::mpi
```
msimberg, line 68: Do you know if we need this dummy definition? We may be ok with removing it completely when MPI is disabled, assuming calls to it are ifdefed out. Especially given the comment that `check_environment` is unused, we may be able to remove it.

libs/pika/mpi_base/src/mpi_environment.cpp
```diff
@@ -19,17 +20,16 @@
...
-        bool detect_mpi_environment(util::runtime_configuration const& cfg, char const* default_env)
+        bool detect_environment(util::runtime_configuration const& cfg, char const* default_env)
         {
 #if defined(__bgq__)
-            // If running on BG/Q, we can safely assume to always run in an
-            // MPI environment
+            // If running on BG/Q, we can safely assume to always run in an MPI environment
             return true;
-#else
+#endif
```
msimberg, line 32: Isn't changing this from `#else` to `#endif` going to lead to warnings about multiple returns from the function, with the second one being unused? I think semantically the `#else` was more appropriate.

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
```diff
@@ -19,59 +19,38 @@
...
+            int* argc, char*** argv, const int required, const int minimal, int& provided);
+
+        // finalize mpi, do not call unless init was previously called
+        static void finalize();
+
+        // returns true if mpi::environment::init has previously been called
+        static bool pika_called_init();
+
+        // convenience functions that retrieve mpi settings
+        static bool is_mpi_inititialized();
```
msimberg, line 45:
```suggestion
        static bool is_mpi_initialized();
```

libs/pika/mpi_base/src/mpi_exception.cpp
```diff
@@ -1,19 +1,20 @@
...
 
 #include <pika/config.hpp>
-#include <pika/async_mpi/mpi_exception.hpp>
 #include <pika/errors/exception.hpp>
-#include <pika/mpi_base/mpi.hpp>
+#include <pika/mpi_base/mpi_exception.hpp>
 
 #include <cstddef>
 #include <memory>
+#include <mpi.h>
```
msimberg, line 14: Please include `pika/mpi_base/mpi.hpp` instead. It suppresses warnings from `mpi.h` that we'd otherwise need to suppress everywhere that `mpi.h` is included.

msimberg (changes requested):

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
```diff
@@ -19,59 +19,38 @@
...
+        static int rank();
+        static int size();
+        static std::string get_processor_name();
+
+    private:
+        static bool mpi_init_pika_;
```
msimberg, line 51: Unused (reported by codacy)? Can it be removed?

libs/pika/runtime/src/runtime.cpp
```diff
@@ -905,15 +906,8 @@ namespace pika::detail {
...
-            if (MPI_Initialized(&mpi_initialized) == MPI_SUCCESS && mpi_initialized)
-            {
-                int rank = 0;
-                if (MPI_Comm_rank(MPI_COMM_WORLD, &rank) == MPI_SUCCESS)
-                {
-                    strm << "MPI rank: " << rank << '\n';
-                }
-            }
+            if (pika::mpi::environment::is_mpi_inititialized())
+                strm << "{mpi-rank}: " << pika::mpi::environment::rank() << '\n';
```
msimberg, line 910: The MPI rank is currently (on main) printed as:
```
*******************************************************************************
MPI rank: 0
   0: PU L#0(P#0), Core L#0(P#2), Socket L#0(P#56), on pool "default"
   1: PU L#1(P#1), Core L#1(P#4), Socket L#0(P#56), on pool "default"
   2: PU L#2(P#2), Core L#2(P#5), Socket L#0(P#56), on pool "default"
   3: PU L#3(P#3), Core L#3(P#6), Socket L#0(P#56), on pool "default"
```

With this change it'd be:
```
*******************************************************************************
{mpi-rank}: 0
   0: PU L#0(P#0), Core L#0(P#2), Socket L#0(P#56), on pool "default"
   1: PU L#1(P#1), Core L#1(P#4), Socket L#0(P#56), on pool "default"
   2: PU L#2(P#2), Core L#2(P#5), Socket L#0(P#56), on pool "default"
   3: PU L#3(P#3), Core L#3(P#6), Socket L#0(P#56), on pool "default"
```
which to me looks odd compared to the output of the thread bindings. Can we keep the old style or use something that's a bit more in line with the thread bindings style?

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
biddisco: it is used by `pika_called_init` if the user uses pika::mpi::init to init mpi instead of manually calling mpi_init themselves

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
```diff
@@ -81,12 +52,12 @@ namespace pika { namespace util {
...
-        {
-            static bool check_mpi_environment(runtime_configuration const& cfg);
-        };
-}}    // namespace pika::util
+namespace pika::mpi {
+    struct PIKA_EXPORT environment
+    {
+        static bool check_environment(runtime_configuration const& cfg);
+    };
+}    // namespace pika::mpi
```
msimberg, line 60: It does look like you should be able to remove the whole `#else` branch here, no? `check_environment` was removed from the other branch, and the namespace was changed to `pika::mpi::detail`, so if it compiles like this it seems like it'll probably compile with this removed as well.

libs/pika/mpi_base/src/mpi_exception.cpp
```diff
@@ -1,19 +1,21 @@
...
 
 #include <pika/config.hpp>
-#include <pika/async_mpi/mpi_exception.hpp>
 #include <pika/errors/exception.hpp>
 #include <pika/mpi_base/mpi.hpp>
+#include <pika/mpi_base/mpi_exception.hpp>
 
 #include <cstddef>
 #include <memory>
+#include <mpi.h>
```
msimberg, line 15: You can remove this include since `pika/mpi_base/mpi_exception.hpp` is already included above.

msimberg (changes requested):
Thanks for updating this @biddisco. Besides the two comments about further deletions, this looks good to me.

libs/pika/mpi_base/include/pika/mpi_base/mpi_environment.hpp
biddisco: Oops, that was a mistake, I meant to remove it all, but must have been distracted when editing ...

msimberg (approved):

PR #1249, opened 2024-09-16
Use C++20 in CUDA 12 CI configuration

msimberg:
Adds a configuration check to see if the GPU compiler supports trivial virtual destructors that can be used in constexpr contexts. nvcc at least up to version 12.6 does not support it (https://compiler-explorer.com/z/4hfvrnzcG).

### DIFF
--- a/.gitlab/includes/gcc12_cuda12_pipeline.yml
+++ b/.gitlab/includes/gcc12_cuda12_pipeline.yml
@@ -12,7 +12,7 @@ include:
   variables:
     SPACK_ARCH: linux-ubuntu22.04-haswell
     COMPILER: gcc@12.1.0
-    CXXSTD: 17
+    CXXSTD: 20
     GPU_TARGET: "60"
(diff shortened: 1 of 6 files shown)


msimberg:
At least CUDA 12.4.1 and GCC 13.3.0 seem to reproduce:
```
/pika/source/libs/pika/execution_base/include/pika/execution_base/any_sender.hpp(65): error: expression must have a constant value
      inline constexpr empty_vtable_t<T> empty_vtable{};
                                                     ^
```

msimberg:
I've attempted to disable the code triggering the error when using CUDA in the latest commit simply to check if the other path works. I think the issue is not C++20, but simply our configuration checks not checking if we have constexpr trivial destructors in both the host and device compilers. The following fails to compile with nvcc: https://compiler-explorer.com/z/36K3P1esK. GCC/clang in C++20 mode do compile it. This is the test we currently use for the host compiler: https://github.com/pika-org/pika/blob/main/cmake/tests/cxx20_trivial_virtual_destructor.cpp.

PR #1251, opened 2024-09-16
Initialize hwloc topology with disallowed nodes to accurately know available resources on system

msimberg:
When cgroups are used (e.g. by srun) to constrain allowed PUs for a process, hwloc will by default only report the PUs and other resources that are allowed to be used with the cgroups restrictions. This can lead to pika thinking that the system has only n PUs when in reality it has m > n PUs. In itself this isn't a problem, but currently on pika main we derive the number of available PUs from the resources that hwloc reports, and thus also the CPU mask sizes may be smaller than the whole system. If using cgroups and `--pika:process-mask`/`PIKA_PROCESS_MASK` with a mask that doesn't contain the first n PUs pika's initialization will currently access CPU masks out of bounds or report an error.

This PR proposes to initialize the hwloc topology with the flag `HWLOC_TOPOLOGY_FLAG_INCLUDE_DISALLOWED`, which will include resources that may not be usable because of cgroups. This allows us to accurately get the maximum size of CPU masks.

Even though we request all resources internally, pika will not be able to bind threads to PUs that are outside of the cgroups restrictions. A warning will be printed if a mask is set outside of the restrictions.

I've relaxed an error to a warning when pika's expected CPU mask for a thread does not match what is retrieved with hwloc. This happens if the thread affinity can't be set.

I've also cleaned up some of the logging in `scheduled_thread_pool_impl.hpp`.

One downside of this approach is that, when using cgroups, hwloc logical PU indices will not correspond to indices reported by e.g. `lstopo`. Physical PU indices are exactly the same as before. I think this is a small inconsistency that's worth the price.

A possible alternative approach is to create a separate topology object with `HWLOC_TOPOLOGY_FLAG_INCLUDE_DISALLOWED` only for counting the number of PUs etc. to correctly size CPU masks, but then use the a topology object without disallowed resources afterwards.

### DIFF
--- a/libs/pika/topology/src/topology.cpp
+++ b/libs/pika/topology/src/topology.cpp
@@ -209,7 +209,13 @@ namespace pika::threads::detail {
         // Including disallowed resources means that logical indices may not match what is reported
         // by the standalone hwloc-bind tool (since the latter will not include disallowed
         // resources). Physical indices will always be consistent.
-        err = hwloc_topology_set_flags(topo, HWLOC_TOPOLOGY_FLAG_INCLUDE_DISALLOWED);
+        err = hwloc_topology_set_flags(topo,
+#if HWLOC_API_VERSION < 0x0002'0100
+            HWLOC_TOPOLOGY_FLAG_WHOLE_SYSTEM
+#else
+            HWLOC_TOPOLOGY_FLAG_INCLUDE_DISALLOWED
+#endif
+        );
         if (err != 0)
         {
             PIKA_THROW_EXCEPTION(pika::error::no_success, "topology::topology",


msimberg:
> Do you think it's worth adding a comment in the code about logical PU indices not matching lstopo's ones?

Maybe, I'll have a look to see where it could fit. Note: The indices will match if no cgroups restrictions are in places. It's only when restrictions are in place that they may not match. Process masks (from hwloc-bind, taskset, slurm etc.) are anyway always using physical indices, so no risk of ambiguity there.

libs/pika/runtime/src/runtime.cpp
```diff
@@ -939,17 +939,17 @@ namespace pika::detail {
...
                 {
                     std::string boundcpu_str =
                         pika::threads::detail::to_string(boundcpu_never_empty);
                     std::string pu_mask_str = pika::threads::detail::to_string(pu_mask_never_empty);
-                    PIKA_THROW_EXCEPTION(pika::error::invalid_status, "handle_print_bind",
-                        "unexpected mismatch between binding reported from HWLOC({}) and "
-                        "pika({}) on thread {}",
+                    PIKA_LOG(warn,
+                        "unexpected mismatch between binding reported from hwloc ({}) and "
+                        "binding expected by pika ({}) on thread {}",
```
msimberg, line 952: Perhaps mention `--pika:print-bind` here to check the bindings?

msimberg: @RMeli what do you think about https://github.com/pika-org/pika/pull/1251/commits/11314692fa49d0fc934640e59e20cb26eae2c4bf?

msimberg:
> LGTM thanks! Do you think it's worth adding a comment in the code about logical PU indices not matching lstopo's ones?

@aurianer how does this look to you: https://github.com/pika-org/pika/pull/1251/commits/4f7199517d21f779c6d258115258d67ef5715afc?

libs/pika/runtime/src/runtime.cpp
RMeli: Looks great, thanks!

RMeli (approved):

PR #1267, opened 2024-10-11
Allow pika to automatically create MPI pool on startup

### LINKED ISSUE #1143: Change default MPI completion mode
We should set it to something that performs reasonably well, while providing a "safety" in terms of running continuations in new tasks etc.

biddisco:
Add a configuration option to enable/disable creation of MPI pool. This removes the need for the user to setup a polling pool manually.

Move many functions into `detail` namespace to designate them as being for internal use only and not for general user consumption.

Clean up some of the enum flags and options that are used to setup pools and polling modes etc.

Change the default mode to "continuation mode" + "inline completion" + "high priority". The MPI pool is disabled by default. Fix #1143.

### DIFF
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -14,6 +14,7 @@
 #include <pika/modules/resource_partitioner.hpp>
 #include <pika/modules/threading_base.hpp>
 #include <pika/mpi_base/mpi.hpp>
+#include <pika/program_options/variables_map.hpp>
 #include <pika/runtime/thread_pool_helpers.hpp>
 #include <pika/synchronization/counting_semaphore.hpp>
 #include <pika/type_support/to_underlying.hpp>
@@ -30,11 +31,11 @@
 
 namespace pika::mpi::experimental {
 
-    enum pool_create_mode
+    /// mode to control installation of pika mpi error -> exception handler
+    enum exception_mode
     {
-        pika_decides = 0,
-        force_create,
-        force_no_create,
+        no_handler = 0,
+        install_handler,
     };
 
     namespace detail {
@@ -174,6 +175,35 @@ namespace pika::mpi::experimental {
             MPI_Request*, MPIX_Continue_cb_function*, void*);
         /// called after each completed continuation to restart/re-enable continuation support
         PIKA_EXPORT void restart_mpix();
+
+        void init_resource_partitioner_handler(pika::resource::partitioner&,
+            pika::program_options::variables_map const& vm,
+            resource::polling_pool_creation_mode mode);
+
+        PIKA_EXPORT bool create_pool(pika::resource::partitioner& rp, std::string const& pool_name,
+            resource::polling_pool_creation_mode mode);
+
+        PIKA_EXPORT const std::string& get_pool_name();
+        PIKA_EXPORT void register_pool(const std::string& pool_name);
+
+        PIKA_EXPORT void register_polling();
+        PIKA_EXPORT void unregister_polling();
+
+        // returns false if no custom mpi pool has been created
+        PIKA_EXPORT bool pool_exists();
+
+        // -----------------------------------------------------------------
+        /// set the maximum number of MPI_Request completions to handle at each polling event
+        PIKA_EXPORT void set_max_polling_size(std::size_t);
+        PIKA_EXPORT std::size_t get_max_polling_size();
+
+        // initialize the pika::mpi background request handler
+        // All ranks should call this function (but only one thread per rank needs to do so)
+        PIKA_EXPORT void start(
+            exception_mode errorhandler = no_handler, std::string pool_name = "");
+
+        // -----------------------------------------------------------------
+        PIKA_EXPORT void stop();
     }    // namespace detail
 
     /// return the total number of mpi requests currently in queues
@@ -182,53 +212,27 @@ namespace pika::mpi::experimental {
     PIKA_EXPORT size_t get_work_count();
 
     // -----------------------------------------------------------------
-    /// set the maximum number of MPI_Request completions to handle at each polling event
-    PIKA_EXPORT void set_max_polling_size(std::size_t);
-    PIKA_EXPORT std::size_t get_max_polling_size();
-
-    // -----------------------------------------------------------------
-    /// Get the polling transfer mode for continuations.
+    /// Set/Get the polling and handler mode for continuations.
     PIKA_EXPORT std::size_t get_completion_mode();
     PIKA_EXPORT void set_completion_mode(std::size_t mode);
 
-    PIKA_EXPORT bool create_pool(
-        std::string const& = "", pool_create_mode = pool_create_mode::pika_decides);
-
-    PIKA_EXPORT const std::string& get_pool_name();
-    PIKA_EXPORT void set_pool_name(const std::string&);
-
-    // returns false if no custom mpi pool has been created
-    PIKA_EXPORT bool pool_exists();
-
-    PIKA_EXPORT void register_polling();
-
+    /// can be used to chooose between single/multi thread model when initializing MPI
     PIKA_EXPORT int get_preferred_thread_mode();
 
-    /// when true pika::mpi can disable the pool, or change the thread mode
+    /// when true pika::mpi can disable pool creation, or change the thread mode
     /// otherwise, the flags passed by the user to completion mode etc are honoured
     PIKA_EXPORT void enable_optimizations(bool enable);
 
-    // initialize the pika::mpi background request handler
-    // All ranks should call this function (but only one thread per rank needs to do so)
-    PIKA_EXPORT void init(bool init_mpi = false, bool init_errorhandler = false);
-
     // -----------------------------------------------------------------
-    PIKA_EXPORT void finalize(std::string const& pool_name = "");
-
-    // -----------------------------------------------------------------
-    // This RAII helper class assumes that MPI initialization/finalization is handled elsewhere
-    struct [[nodiscard]] enable_user_polling
+    // This RAII helper class ensures that MPI polling start/stop is handled correctly
+    struct [[nodiscard]] enable_polling
     {
-        enable_user_polling(std::string const& pool_name = "", bool init_errorhandler = false)
-          : pool_name_(pool_name)
+        /// an empty pool name tells pika to use the mpi pool if it exists
+        enable_polling(exception_mode errorhandler = no_handler, std::string const& pool_name = "")
         {
-            mpi::experimental::init(false, init_errorhandler);
-            mpi::experimental::register_polling();
+            mpi::experimental::detail::start(errorhandler, pool_name);
         }
 
-        ~enable_user_polling() { mpi::experimental::finalize(pool_name_); }
-
-    private:
-        std::string pool_name_;
+        ~enable_polling() { mpi::experimental::detail::stop(); }
     };
 }    // namespace pika::mpi::experimental
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -71,29 +71,26 @@ namespace pika::mpi::experimental {
         /// flags that control how mpi continuations are handled
         enum class handler_method : std::uint32_t
         {
-            /// enable the use of a dedicated pool for polling mpi messages.
-            use_pool = 0b0000'0001,    // 1
-
             /// this bit enables the inline invocation of the mpi request, when set
             /// the calling thread performs the mpi operation, when unset, a transfer
             /// is made so that the invocation happens on a new task that
             /// would normally be on a dedicated pool if/when it exists
-            request_inline = 0b0000'0010,    // 2
+            request_inline = 0b0000'0001,    // 2
 
             /// this bit enables the inline execution of the completion handler for the
             /// request, when unset a transfer is made to move the completion handler
             /// from the polling thread onto a new one
-            completion_inline = 0b0000'0100,    // 4
+            completion_inline = 0b0000'0010,    // 4
 
             /// this bit enables the use of a high priority task flag
             /// 1) requests are boosted to high priority if they are passed the the mpi-pool
             ///    to ensure they execute before other polling tasks (reduce latency)
             /// 2) completions are boosted to high priority when sent to the main thread pool
             ///    so that the continuation is executed as quickly as possible
-            high_priority = 0b0000'1000,    // 8
+            high_priority = 0b0000'0100,    // 8
 
             /// 3 bits control the handler method,
-            method_mask = 0b0111'0000,    // 70
+            method_mask = 0b0011'1000,    // 56
 
             /// Methods supported for dispatching continuations:
             ///
@@ -112,15 +109,11 @@ namespace pika::mpi::experimental {
             ///
             /// * unspecified : reserved for development purposes or for customization by an
             /// application using pika
-            yield_while = 0b0000'0000,               // 0x00, 00 ... 15
-            suspend_resume = 0b0001'0000,            // 0x10, 16 ... 31
-            new_task = 0b0010'0000,                  // 0x20, 32 ... 47
-            continuation = 0b0011'0000,              // 0x30, 48 ... 63
-            mpix_continuation = 0b0100'0000,         // 0x40, 64 ... 79
-            unspecified = mpix_continuation + 16,    // 0x50, ...
-
-            /// Default flags are to invoke inline, but transfer completion using a dedicated pool
-            default_mode = use_pool | request_inline | high_priority | new_task,
+            yield_while = 0b0000'0000,          // 0x00, 00 ... 7
+            suspend_resume = 0b0000'1000,       // 0x08, 08 ... 15
+            new_task = 0b0001'0000,             // 0x10, 16 ... 23
+            continuation = 0b0001'1000,         // 0x18, 24 ... 31
+            mpix_continuation = 0b0010'0000,    // 0x20, 32 ... 39
         };
 
         /// 3 bits define continuation mode
@@ -145,11 +138,6 @@ namespace pika::mpi::experimental {
         {
             return (mode & pika::detail::to_underlying(handler_method::request_inline)) != 0;
         }
-        /// 1 bit defines whether we use a pool or not
-        inline bool use_pool(int mode)
-        {
-            return (mode & pika::detail::to_underlying(handler_method::use_pool)) != 0;
-        }
 
         /// used for debugging to show mode type in messages, should be removed
         inline const char* mode_string(int flags)
@@ -161,7 +149,6 @@ namespace pika::mpi::experimental {
             case handler_method::continuation: return "continuation";
             case handler_method::suspend_resume: return "suspend_resume";
             case handler_method::mpix_continuation: return "mpix_continuation";
-            case handler_method::unspecified: return "unspecified";
             default: return "invalid";
             }
         }
@@ -204,15 +191,20 @@ namespace pika::mpi::experimental {
 
     // -----------------------------------------------------------------
     /// Set/Get the polling and handler mode for continuations.
+    PIKA_EXPORT bool get_enable_pool();
+    PIKA_EXPORT void set_enable_pool(bool);
+
+    // -----------------------------------------------------------------
+    /// Set/Get the polling and handler mode for continuations.
     PIKA_EXPORT std::size_t get_completion_mode();
-    PIKA_EXPORT void set_completion_mode(std::size_t mode);
+    PIKA_EXPORT void set_completion_mode(std::size_t);
 
     /// can be used to choose between single/multi thread model when initializing MPI
     PIKA_EXPORT int get_preferred_thread_mode();
 
     /// when true pika::mpi can disable pool creation, or change the thread mode
     /// otherwise, the flags passed by the user to completion mode etc are honoured
-    PIKA_EXPORT void enable_optimizations(bool enable);
+    PIKA_EXPORT void enable_optimizations(bool);
 
     /// return the name assigned to the mpi polling pool
     PIKA_EXPORT const std::string& get_pool_name();
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -183,7 +183,6 @@ namespace pika::mpi::experimental {
         PIKA_EXPORT bool create_pool(pika::resource::partitioner& rp, std::string const& pool_name,
             resource::polling_pool_creation_mode mode);
 
-        PIKA_EXPORT const std::string& get_pool_name();
         PIKA_EXPORT void register_pool(const std::string& pool_name);
 
         PIKA_EXPORT void register_polling();
@@ -214,6 +213,9 @@ namespace pika::mpi::experimental {
     /// when true pika::mpi can disable pool creation, or change the thread mode
     /// otherwise, the flags passed by the user to completion mode etc are honoured
     PIKA_EXPORT void enable_optimizations(bool enable);
+
+    /// return the name assigned to the mpi polling pool
+    PIKA_EXPORT const std::string& get_pool_name();
 
     // initialize the pika::mpi background request handler
     // All ranks should call this function (but only one thread per rank needs to do so)
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -71,11 +71,6 @@ namespace pika::mpi::experimental {
         /// on any error instead of the default behavior of program termination
         PIKA_EXPORT void set_error_handler();
 
-        // -----------------------------------------------------------------
-        /// Background progress function for MPI async operations
-        /// Checks for completed MPI_Requests and sets ready state in waiting receivers
-        PIKA_EXPORT pika::threads::detail::polling_status poll();
-
         /// utility function to avoid duplication in eager check locations
         PIKA_EXPORT bool poll_request(MPI_Request /*req*/);
 
--- a/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
+++ b/libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
@@ -195,17 +195,18 @@ namespace pika::mpi::experimental {
         /// set the maximum number of MPI_Request completions to handle at each polling event
         PIKA_EXPORT void set_max_polling_size(std::size_t);
         PIKA_EXPORT std::size_t get_max_polling_size();
+
+        // -----------------------------------------------------------------
+        /// Set/Get the pool_enabled flag
+        PIKA_EXPORT bool get_pool_enabled();
+        PIKA_EXPORT void set_pool_enabled(bool);
+
     }    // namespace detail
 
     /// return the total number of mpi requests currently in queues
     /// This number is an estimate as new work might be created/added during the call
     /// and so the returned value should be considered approximate (racy)
     PIKA_EXPORT size_t get_work_count();
-
-    // -----------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


libs/pika/resource_partitioner/include/pika/resource_partitioner/partitioner_fwd.hpp
```diff
@@ -50,6 +49,25 @@ namespace pika::resource {
...
+        mode_manual = 0,
+        /// if the completion mode requires it, a pool will be created at startup
+        mode_pika_decides = 1,
+        /// overrides command line flags/env vars - enables creation of a polling pool
+        mode_force_create = 2,
+        /// overrides command line flags/env vars - disables creation of a polling pool
+        mode_force_no_create = 3,
+    };
+#endif
+
```
msimberg, line 70: The enum doesn't mention anything about "MPI" but the docstrings keep mentioning "MPI". There's obviously the CUDA polling as well, which is not mentioned here. Either we:
- keep that out of this and if in the future we want a CUDA polling pool we duplicate this, but then the variables should be MPI-specific (i.e. have `mpi` in the name)
- or we leave the topic open and don't mention MPI here now, to leave it open so that CUDA polling could in the future be told to happen on the (one and only) polling pool

What do you think?

msimberg (changes requested):
Thanks @biddisco. I think this is already looking good.

I think the `polling_pool_creation_mode` should be a `runtime_configuration` option though, not a resource partitioner option. It belongs alongside `pika.mpi.completion_mode` in my opinion. As I've mentioned before I'd go one step further and remove the `pool` bit from the completion mode and have it in a separate configuration option.

Were there any practical concerns that made you put it in the resource partitioner at this point or do you think it'd be feasible to make it a regular configuration option?

Not for this PR, but I think if we enable polling by default with MPI enabled, then I think we should do the same for CUDA polling. I think last time I benchmarked it, the overhead of doing polling even when there are no requests/events was minimal. If that's a concern we could enable polling by default and have a configuration option that can be used to turn polling off explicitly.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -688,155 +693,182 @@ namespace pika::mpi::experimental {
...
-        if (init_mpi)
+        if (pool_name.empty())
         {
-            int provided, required = get_preferred_thread_mode();
-            mpi::detail::environment::init(nullptr, nullptr, required, required, provided);
+            pool_name = detail::pool_exists() ? get_pool_name() :
+                                                resource::get_partitioner().get_default_pool_name();
         }
-        else if (!mpi::detail::environment::is_mpi_initialized())
+        detail::register_pool(pool_name);
```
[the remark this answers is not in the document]
biddisco: Essentially yes. If we let pika do everything, then create pool calls register and all is good, but if the user manually creates a pool and then starts polling, we need to make sure internal stuff is set and so an extra register is needed - I can make the user call register themselves in that case, but I've done my best to ensure that it is safe so that this is harmless as is

libs/pika/resource_partitioner/include/pika/resource_partitioner/partitioner_fwd.hpp
biddisco: Tjhat's exactly why I was being cryptic - I left MPI out of the actual name/flags, but mentioned it in the docs, because right now we only use it for mpi, but the first polling changes also moved cuda to the pool (that was reverted a long time ago), but we might want to try it again, now that stuff is more mature.
I'll rewrite the doc to be a bit more explicit about current use cases ...

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -688,155 +693,182 @@ namespace pika::mpi::experimental {
...
-        std::string name = pool_name;
-        if (name.empty()) { name = get_pool_name(); }
-        else
+        // -----------------------------------------------------------------
+        void unregister_polling()
         {
-            // override mpi pool name with whatever we decided on
-            set_pool_name(name);
+            auto mode = get_completion_mode();
+            // enable pika polling on the mpi pool if the handling mode needs it
```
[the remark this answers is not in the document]
biddisco: The main issue is that we need to reserve a core before the main init is done, so adding a flag to init in place of the existing callback made sense, but I suppose I could simply add a direct call into pika::mpi to do the pool create during init and not have the user set a flag at all. 
Seems a bit intrusive to create pools without the user explicitly asking for one - likewise having polling enabled by default is ok if the pool is custom, but on the default pool - hmmm. Users might well compile pika with mpi support and use it for app1, but also app2 and might not use mpi much or at all and it's better to ask to enable - than ask to disable a feature like that.

I would agree that having 
`--pika:mpi:enable-pool` and `--pika:mpi:completion-mode=42` is the way to go - how about if I leave the pool init as it is for now and in a new pr, add command line option parsing on a module-by-module basis so that we can have explicit `pika:mpi:` options when mpi is enabled and not when it is not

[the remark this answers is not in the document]
biddisco: on second thoughts, I'll add a flag for `mpi-enable-pool` to complement `completion-mode` in this PR and think about module stuff later - NB the completion modes will need to change numbering once pool is removed from the bitmask.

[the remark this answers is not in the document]
biddisco: I'd prefer to remove modes in a separate PR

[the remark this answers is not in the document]
msimberg: Yeah, perfectly fine by me (I also prefer it that way). We should try to do it before the next release though.

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -887,15 +921,36 @@ namespace pika::mpi::experimental {
...
 
-        PIKA_DETAIL_DP(detail::mpi_debug<5>,
-            debug(str<>("Clearing mode"), detail::mpi_data_, "disable_user_polling"));
+    // -----------------------------------------------------------------
+    size_t get_work_count() { return detail::mpi_data_.all_in_flight_; }
 
-        if (pool_name.empty())
+    // -----------------------------------------------------------------
+    bool get_enable_pool() { return detail::enable_pool_; }
+    void set_enable_pool(bool mode) { detail::enable_pool_ = mode; }
```
msimberg, line 933: Naming and namespace:
- `get_enable_pool` -> `pool_enabled` or something like that?
- at least `set_enable_pool` into `detail`, if not both? They're only used internally as far as I can tell right now. We can always make them public later if needed.

I feel like the `set/get_enable_pool` leaks the internal details in the sense that you're just setting/getting an internal variable. If they move to `detail` I think the names are fine as they won't be user-facing anymore then.

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -182,53 +203,43 @@ namespace pika::mpi::experimental {
...
-            mpi::experimental::init(false, init_errorhandler);
-            mpi::experimental::register_polling();
+            start_polling(errorhandler, pool_name);
         }
 
-        ~enable_user_polling() { mpi::experimental::finalize(pool_name_); }
-
-    private:
-        std::string pool_name_;
+        ~enable_polling() { stop_polling(); }
```
msimberg, line 243: I think most/all of this can be in `detail`?

`start/stop/enable_polling` could arguably stay in non-detail I suppose. However, I still think we could enable polling by default _or_ start polling the first time `transform_mpi` is used. What do you think?

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -4,25 +4,42 @@
...
 
 set(algorithm_transform_mpi_PARAMETERS THREADS 2 RANKS 2 MPIWRAPPER)
 set(algorithm_transform_mpi_DEPENDENCIES pika_execution_test_utilities)
 
+set(pool_creation_PARAMETERS THREADS 2 RANKS 2 MPIWRAPPER)
+set(pool_creation_NO_POOL ON)
+# cmake-format: on
+
+# note that because we add (eg) "_0_5" to the test name, we explicitly link pika_testing because
+# pika_add_unit_test expects "${name}_test"
```
msimberg, line 42: I don't understand this comment. On main we currently name the executable (e.g.) `mpi_ring_async_sender_receiver_test` and the tests are called `mpi_ring_async_sender_receiver_mode_X_test` and it links just fine. `pika_add_test` will link any executable ending with `_test` to `pika_testing` (https://github.com/pika-org/pika/blob/34f044e33b242a87ca1b2a06a702e0763af11087/cmake/pika_add_test.cmake#L108-L110), and as far as I can tell you haven't changed the executable name.

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -514,8 +487,8 @@ int main(int argc, char* argv[])
     nws_deb<6>.debug(3, "Calling pika::init");
     pika::init_params init_args;
     init_args.desc_cmdline = cmdline;
-    // Set the callback to init thread_pools
-    init_args.rp_callback = &init_resource_partitioner_handler;
+    // Tell init to create an mpi thread pool if needed
+    init_args.resource::polling_pool_creation_mode = ::pika::resource::polling_pool_pika_decides;
```
msimberg, line 491: Shouldn't this be removed?

libs/pika/async_mpi/tests/unit/pool_creation.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+        // Operator| overload
+        {
+            // MPI function pointer
+            int data = 0, count = 1;
+            if (rank == 0) { data = 42; }
+            tt::sync_wait(
+                ex::just(&data, count, datatype, 0, comm) | mpi::transform_mpi(MPI_Ibcast));
+            PIKA_TEST_EQ(data, 42);
+        }
```
msimberg, line 66: I don't think you this test needs to use both the regular function call and the pipe version here, since it's tested elsewhere already. Feel free to only use only whichever version you find more natural to use in each case.

libs/pika/async_mpi/tests/unit/pool_creation.cpp
```diff
@@ -0,0 +1,152 @@
...
+int main(int argc, char* argv[])
+{
+    MPI_Init(&argc, &argv);
+
+    // ---------------------------------------
+    // Start runtime and collect runtime exit status, turn on auto pool creation
+    pika::init_params init_args1;
+    init_args1.cfg.emplace_back("pika.mpi.enable_pool=" + std::to_string(true));
+    auto result1 = pika::init([&]() { return pika_main(""); }, argc, argv, init_args1);
+    PIKA_TEST_EQ(result1, 0);
```
msimberg, line 132: If you don't want to increment the variable names with a number you can put braces around each test case. I think it makes it a bit easier to read as well, since each test case is clearly separated.

libs/pika/command_line_handling/src/parse_command_line.cpp
```diff
@@ -340,8 +340,11 @@ namespace pika::detail {
                 "no cross boundary stealing is allowed (default value: 0)")
 #if defined(PIKA_HAVE_MPI)
             ("pika:mpi-completion-mode", value<std::size_t>(),
-                "the pika MPI polling completion mode (only available if MPI "
-                "built with MPI support)")
+                "the pika MPI polling completion/handler mode "
+                "(only available if MPI built with MPI support)")
+            ("pika:mpi-enable-pool",
+                "enable the creation of a dedicated pool for mpi-polling "
+                "(only available if MPI built with MPI support and available threads>1).")
```
msimberg, line 347: Regarding "only available if ... available threads>1": will using this flag result in error or not?

libs/pika/resource_partitioner/include/pika/resource_partitioner/detail/partitioner.hpp
```diff
@@ -220,8 +220,7 @@ namespace pika::resource::detail {
         // initial_thread_pools
         std::vector<socket> sockets_;
 
-        // store policy flags determining the general behavior of the
-        // resource_partitioner
+        // store policy flags determining the general behavior of the resource_partitioner
```
msimberg, line 223: Not related to this PR anymore.

libs/pika/resource_partitioner/include/pika/resource_partitioner/partitioner_fwd.hpp
```diff
@@ -37,8 +37,7 @@ namespace pika::resource {
     /// Returns false otherwise.
     PIKA_EXPORT bool is_partitioner_valid();
 
-    /// This enumeration describes the modes available when creating a
-    /// resource partitioner.
+    /// This enumeration describes the modes available when creating a resource partitioner.
```
msimberg, line 40: Not related anymore.

msimberg (changes requested):

msimberg:
Also see the thread sanitizer failures. I think you will need to add some synchronization or atomics to ensure that we don't do anything bad during initialization.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: we pass in a name of (eg) `mpi_ring_async_sender_receiver_mode_0_0` and it has flags `EXECUTABLE ${test}_test` which is correct (the test executable is `mpi_ring_async_sender_receiver_test`, but the check you linked to tests
```
if(TARGET ${${name}_EXECUTABLE}_test 
```
which will expand to `mpi_ring_async_sender_receiver_test_test` which is not correct and I couldn't be arsed fixing it, so just manually linked the testing lib.
Now that I write it down, I could fix it by just passing the EXECUTABLE as `mpi_ring_async_sender_receiver` and then the `test` suffix will match. I will try that ...

biddisco: well, that does work, but it strikes me as being wrong. the test is added with 
```
  pika_add_executable(
    ${test}_test 
```
which has `_test` appended as a suffix, so the executable is called xxx_test, so passing ```EXECUTABLE xxx_test```  syntactically seems right and the add_unit_test adding `_test` again is unnecessary.  I hope this makes sense. it is a bit confusing, anyway, I can either link `pika_testing` as I'm doing or use the (now) modified version of
```
      pika_add_unit_test(
        "modules.async_mpi"
        ${full_name_}
        ${${test}_PARAMETERS}
        EXECUTABLE
        ${test}
        ARGS
        $<$<BOOL:${enable_pool}>:--pika:mpi-enable-pool>
        "--pika:mpi-completion-mode=${polling_mode}"
      )
```
which drops the test suffix and works.

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
biddisco: correct. I forgot to check the other tests properly it seems.

libs/pika/command_line_handling/src/parse_command_line.cpp
biddisco: it prints a warning, but I'd better switch it to use PIKA_LOG as you suggested for other messages - we can throw instead ...

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: (what I forgot to say last night is that, really add_unit_test, is just adding a wrapper around an executable to invoke it with the right params, and the linking to pika_testing should happen in the add_executable itself - which is where I added it in the PR originally). But the comment was completely wrong :(

msimberg: That's a good catch! However, given that it works here without the changes you've made could you remove these changes from here? I've opened https://github.com/pika-org/pika/pull/1271 as a starting point to fix it and/or change the behaviour.

I think I agree in one sense that adding a test should not link. On the other hand this is done to enforce that tests and only tests link to `pika_testing`. If it's a test it should always be linking to it (and fail if no tests were actually done with `PIKA_TEST` et al., since we fail a test if no tests were run). If it's not a test it shouldn't link to `pika_testing`. And if it's using `PIKA_TEST` but pretending it's not a test, it is a test.

I'd like to avoid that we have to explicitly link to `pika_testing` everywhere. A half-way explicit option could be to add a `TESTING` flag to `pika_add_executable`. Would you mind commenting on https://github.com/pika-org/pika/pull/1271 if you have a preference or have other ideas on how we could do it differently?

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
biddisco: I've moved everything that isn't used by DLAF or by a unit test (for an explicit feature) into detail

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -174,61 +175,75 @@ namespace pika::mpi::experimental {
...
 
-    PIKA_EXPORT void register_polling();
+    }    // namespace detail
+
+    /// when true pika::mpi can disable pool creation, or change the thread mode
+    /// otherwise, the flags passed by the user to completion mode etc are honoured
+    PIKA_EXPORT void enable_optimizations(bool);
 
+    // -----------------------------------------------------------------
+    /// Set/Get the polling and handler mode for continuations.
```
msimberg, line 215:
```suggestion
    /// Get the polling and handler mode for continuations.
```

libs/pika/async_mpi/tests/unit/CMakeLists.txt
msimberg, line 42:
```suggestion
```

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -37,18 +54,27 @@ foreach(test ${tests})
...
+    foreach(polling_mode RANGE 0 ${PIKA_MPI_MODES_LOOP_COUNT} 1)
+      set(full_name_ ${test}_mode_${enable_pool}_${polling_mode})
+      pika_add_pseudo_target(${full_name_}_test)
+      pika_add_pseudo_dependencies(${full_name_}_test ${test}_test)
+      pika_add_unit_test(
+        "modules.async_mpi"
+        ${full_name_}
+        ${${test}_PARAMETERS}
+        EXECUTABLE
+        ${test}
```
msimberg, line 73: This should still have the `_test` suffix. The bug is in `pika_add_test`, not here.
```suggestion
        ${test}_test
```

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -37,18 +54,27 @@ foreach(test ${tests})
...
+        ARGS
+        $<$<BOOL:${enable_pool}>:--pika:mpi-enable-pool>
+        "--pika:mpi-completion-mode=${polling_mode}"
+      )
+    endforeach()
   endforeach()
-  pika_add_unit_test("modules.async_mpi" ${test} ${${test}_PARAMETERS})
```
msimberg, line 53: I would probably keep this one. It's good to know how the default case behaves.

(I see now that you had to add the explicit linking to `pika_testing` because this case was removed.)

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -373,7 +370,7 @@ int pika_main(pika::program_options::variables_map& vm)
     options.warmup = false;
     options.final = false;
 
-    mpi::set_max_requests_in_flight(options.in_flight_limit);
+    // mpi::set_max_requests_in_flight(options.in_flight_limit);
```
msimberg, line 373: Remove or uncomment.

libs/pika/async_mpi/tests/unit/mpi_async_storage.cpp
```diff
@@ -490,13 +463,10 @@ int main(int argc, char* argv[])
...
-        "no-mpi-pool", pika::program_options::bool_switch(), "Disable the MPI pool.");
-
-    cmdline.add_options()("in-flight-limit",
-        pika::program_options::value<std::uint32_t>()->default_value(
-            mpi::get_max_requests_in_flight()),
-        "Apply a limit to the number of messages in flight.");
+    // cmdline.add_options()("in-flight-limit",
+    //     pika::program_options::value<std::uint32_t>()->default_value(
+    //         mpi::get_max_requests_in_flight()),
+    //     "Apply a limit to the number of messages in flight.");
```
msimberg, line 469: Remove or uncomment (or open an issue to add it back in together with remove).

libs/pika/async_mpi/tests/unit/pool_creation.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    int result = 0;
+    {
+        // ---------------------------------------
+        // Start runtime and collect runtime exit status, turn on auto pool creation
+        pika::init_params init_args;
+        init_args.cfg.emplace_back("pika.mpi.enable_pool=" + std::to_string(true));
+        result |= pika::init([&]() { return pika_main(""); }, argc, argv, init_args);
+        PIKA_TEST_EQ(result, 0);
+    }
```
msimberg, line 114: Just for you to know: you don't have to accumulate and return the `result` from `main`. If any of the `PIKA_TEST_EQ(result, 0)` fail, the test will exit with a non-zero code anyway. Since you're testing each `result` returned from `pika::init`, I think you can remove the accumulation and just define `int result = 0` in each scope.

libs/pika/async_mpi/tests/unit/CMakeLists.txt
biddisco: An executable either needs to link to a library or not - whether you actually use it as a unit test or not should not affect the link rules (unless there is some kind of plugin mechanism at play). I did not look at what `pika_testing` actually provides but I will add this to my todo list

msimberg: I half-agree.

Forcing the linking of `pika_testing` is to ensure that a unit test actually does some testing. A unit test that doesn't use `PIKA_TEST` is either not a unit test or someone forgot to use `PIKA_TEST`. Conversely, a test that pretends not to be a unit test but uses `PIKA_TEST` is a unit test, whether it wants it or not. In other words, these all imply each other "link to `pika_testing`" <=> "uses `PIKA_TEST`" <=> "is a unit test".

No plugin mechanism at play.

We do this at the exit of the test fixture: https://github.com/pika-org/pika/blob/4bb041ca52bdf61ab541f49830604e7291f04c41/testing/testing/src/testing.cpp#L37, in case you want to dig into it further.

I'll mark this as resolved though since this part is now unchanged on this PR.

msimberg (approved):

msimberg:
We'll have to wait a bit for CSCS CI to be alive again...

libs/pika/async_mpi/tests/unit/CMakeLists.txt
```diff
@@ -4,25 +4,40 @@
...
+      "--rounds=5"
+      "--iterations=25"
+      "--message-bytes=64"
+      "--pika:ignore-process-mask"
+    THREADS 4
+    RANKS 2
+    MPIWRAPPER
 )
 
 set(mpi_async_storage_PARAMETERS
```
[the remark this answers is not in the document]
biddisco: it's not working, but in previous review comments you asked me to remove api refs that had been changed, so I slightly cleaned up the test in readiness for me having a go at re-enabling it

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -688,168 +736,189 @@ namespace pika::mpi::experimental {
...
+            // needed, in order to decide if pool can be disabled on 1 rank
+            mpi_data_.rank_ = mpi::detail::environment::rank();
+            mpi_data_.size_ = mpi::detail::environment::size();
+
+            bool do_create = true;
+            if (mode == polling_pool_creation_mode::mode_pika_decides)
+            {
+                if (mpi_data_.size_ == 1)
+                {
+                    PIKA_LOG(warn, "MPI Pool creation disabled, running on a single rank");
```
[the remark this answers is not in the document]
biddisco: I'm undecided on this. Usually you develop a test or app and run it, then to debug something, you might try running it on one rank to see if a problem still exists - with all other options unchanged - now pika will turn off the pool (if you had been using it) and you might not have expected that - it might substantially change the scheduling of tasks etc compared to what you had been testing with previously - is that a warning or an info?

libs/pika/async_mpi/src/mpi_polling.cpp
```diff
@@ -599,34 +610,33 @@ namespace pika::mpi::experimental {
...
-            bool single_threaded = singlethreaded(mode);
-
-#ifdef OMPI_HAVE_MPI_EXT_CONTINUE
-            // if mpi continuations are available, poll here and bypass main routine
-            if (get_handler_method(mode) == handler_method::mpix_continuation)
-            {
-                return try_mpix_polling(single_threaded);
-            }
+            return try_mpix_polling(false);
+        }
```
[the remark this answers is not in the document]
biddisco: aha good idea

libs/pika/async_mpi/include/pika/async_mpi/mpi_polling.hpp
```diff
@@ -117,11 +117,12 @@ namespace pika::mpi::experimental {
...
-            suspend_resume = 0b0000'1000,       // 0x08, 08 ... 15
-            new_task = 0b0001'0000,             // 0x10, 16 ... 23
-            continuation = 0b0001'1000,         // 0x18, 24 ... 31
-            mpix_continuation = 0b0010'0000,    // 0x20, 32 ... 39
+            yield_while = 0b0000'0000,                                          // 0x00, 00 -> 7
+            suspend_resume = 0b0000'1000,                                       // 0x08, 08 -> 15
+            new_task = 0b0001'0000,                                             // 0x10, 16 -> 23
+            continuation = 0b0001'1000,                                         // 0x18, 24 -> 31
+            mpix_continuation = 0b0010'0000,                                    // 0x20, 32 -> 39
+            default_mode = continuation + completion_inline + high_priority,    // 24 + 2 + 4 = 30
```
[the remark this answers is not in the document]
biddisco: yes, mode 30, should be mode 60 in old money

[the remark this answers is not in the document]
biddisco: and by default, pool is already off

libs/pika/async_mpi/src/mpi_polling.cpp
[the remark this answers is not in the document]
msimberg: Yeah, I agree it's a borderline case. My thinking is that it's expected behaviour that it gets disabled and there's not really anything the user can do at the moment (at runtime) to either force the creation of the pool or disable the warning (except by disabling _all_ warnings). It's (IMO) more "FYI" (i.e. `info`) rather than "you probably did something wrong, please check" (i.e. `warn`).

With a single rank there shouldn't be any (important) MPI work being done anyway.=

> it might substantially change the scheduling of tasks etc

How significant do you expect this to be? Not doing MPI calls on a single rank would already change scheduling of tasks. In the ideal case having or not having the MPI pool disabled with a single rank should mean you have one thread less, which should have minimal impact. (Keyword: should).

That said, I get your point and I'm not set on the above.

Most importantly: the behaviour should be documented on https://pikacpp.org/usage.html.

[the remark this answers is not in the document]
msimberg: Addition: logging currently prints warnings and worse by default. Info isn't printed by default.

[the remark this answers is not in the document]
biddisco: I don't think we have a command-line way of forcing the creation of the pool on 1 rank. :(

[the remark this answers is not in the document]
msimberg: Indeed, I don't think so either, but I also don't think it really makes sense to have that (except possibly as a debug/internal option; I would definitely not advertise that option if it existed)?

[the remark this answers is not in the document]
msimberg: Let's keep it as info for now then and see if there's any feedback once merged (likely none).

PR #1268, opened 2024-10-17
Update macOS CI workflow to use macos-15

msimberg:
`macos-12` will be removed in December 2024.

### DIFF
--- a/.github/workflows/macos_debug.yml
+++ b/.github/workflows/macos_debug.yml
@@ -19,7 +19,7 @@ jobs:
   build_and_test:
     name: github/macos/debug
     if: ${{ github.event_name == 'merge_group' }}
-    runs-on: macos-12
+    runs-on: macos-15
 
     steps:
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


msimberg:
@albestro, @rasolca, @RMeli do you know what the right way to install libraries is on macOS with CMake? We previously used the `macos-12` runner, but since it's being removed I've updated this to use `macos-15`. After the update running an installed `hello_world` ends up failing to run because it doesn't find libpika (https://github.com/pika-org/pika/actions/runs/11382602682/job/31666447087#step:9:5):
```
dyld[13137]: Library not loaded: @rpath/libpikad.0.dylib
  Referenced from: <D911E273-1415-3E24-8C39-55C58B480673> /usr/local/bin/hello_world
  Reason: no LC_RPATH's found
/Users/runner/work/_temp/cc5db1f8-e3f8-4c4f-89ac-7c53f2ba6403.sh: line 1: 13137 Abort trap: 6           hello_world
```
If I set `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` and `CMAKE_INSTALL_RPATH` (through `PIKA_WITH_FULL_RPATH=ON`) the executable runs fine. I'm ok with this, but I was curious if you have more experience with how things _should_ be set up? For some reason (likely historical) we always set rpath for linux, but disable it for macOS and Windows (by default).

PR #1274, opened 2024-10-18
Use `MPI_Init_thread` in `algorithm_transform_mpi` test

msimberg:
It's not really clear if we've just gotten lucky or if something else was going on in that test, but we should be initializing MPI with `MPI_THREAD_MULTIPLE` to be on the safe side.

Thank you @RMeli for noticing this.

### DIFF
--- a/.github/workflows/linux_asan_ubsan_lsan.yml
+++ b/.github/workflows/linux_asan_ubsan_lsan.yml
@@ -33,13 +33,6 @@ jobs:
         uses: hendrikmuhs/ccache-action@v1.2
         with:
           key: ccache-linux-sanitizers-address-undefined-leak
-      - name: Switch OpenMPI to MPICH
-        shell: bash
-        run: |
-            apt-get update
-            apt-get remove --yes mpi-default-dev
-            apt-get autoremove --yes
-            apt-get install --no-install-recommends --yes libmpich-dev
       - name: Configure
         shell: bash
         run: |
@@ -80,6 +73,11 @@ jobs:
             # https://github.com/actions/runner-images/issues/9491
             sysctl --write vm.mmap_rnd_bits=28
 
+            # We are certain that we want to run mpiexec as root despite warnings as that is the
+            # only user available in the container. Mistakes will only affect the current step.
+            export OMPI_ALLOW_RUN_AS_ROOT=1
+            export OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+
             export ASAN_OPTIONS=fast_unwind_on_malloc=0:strict_string_checks=1:detect_leaks=1:detect_stack_use_after_return=0:check_initialization_order=1:strict_init_order=1:suppressions=$PWD/tools/asan.supp
             export UBSAN_OPTIONS=print_stacktrace=1:suppressions=$PWD/tools/ubsan.supp
             cd build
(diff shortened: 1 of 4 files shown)


msimberg:
@biddisco it occurs to me that `get_preferred_thread_mode` is near impossible to use in a meaningful way. Ideally MPI should be initialized before pika, but if one uses `get_preferred_thread_mode` before pika has been initialized, it won't return the correct value (configuration options and command line options are not taken into account yet at that point) except by chance. If one tries to use it after pika has been initialized, it's too late since MPI should already have been initialized for pika by that point. Is this accurate?

biddisco:
> @biddisco it occurs to me that `get_preferred_thread_mode` is near impossible to use in a meaningful way. Ideally MPI should be initialized before pika, but if one uses `get_preferred_thread_mode` before pika has been initialized, it won't return the correct value (configuration options and command line options are not taken into account yet at that point) except by chance. If one tries to use it after pika has been initialized, it's too late since MPI should already have been initialized for pika by that point. Is this accurate?

I'm sure I replied to this already somewhere ... 
Correct. If ENV vars are used to set completion mode, pool status etc, then there is no requirement to init pika before the internal flags are set correctly. This is an unfortunate consequence of us adding support for command-line vars that I had forgotten about.

Also, it would be nice to use our error checking init in mpi:environment
```
    int provided;
    pika::mpi::detail::environment::init(
        &argc, &argv, MPI_THREAD_MULTIPLE, MPI_THREAD_MULTIPLE, provided);
``` 
but that has been moved into detail namespace so should not be in public facing code.

biddisco:
```
    int provided, preferred = mpix::get_preferred_thread_mode();
    pika::mpi::detail::environment::init(
        &argc, &argv, preferred, preferred, provided);
```
would be the original way of doing it using ENV vars.

biddisco:
I will include this fix in #1267 and this can be closed.

msimberg:
This needs further investigation. The test which now uses `MPI_Init_thread` now fails/times out depending on the MPI mode.

biddisco:
from slack 
```Does it look like whitelisting may be enough? = naive suggestion for "not sure what to do", i.e. can we ignore the warnings if they are warnings from inside mpich and we're not doing anything wrong?```

If we do not register an error handler, mpich does not hang on exit, if we do register an error handler, but don't trigger any errors, mpich doesn't hang on exit, if we register the handler, do trigger errors, but don't use exceptions, then it still hangs.
I therefore conclude that the problem lies inside mpich handling of errors when an error handler is registered.
This issue https://github.com/pmodels/mpich/issues/268 in mpich appears related, but was "fixed" many years ago.  

I suggest we mark this test as problematic in CI and file an issue with mpich with a reproducer (which I will attempt to create right away)

msimberg:
> I suggest we mark this test as problematic in CI and file an issue with mpich with a reproducer (which I will attempt to create right away)

I will have a few gos at suppressing the warnings/errors as `algorithm_transform_mpi` is one of the main tests for the MPI functionality and completely ignoring it for sanitizer testing would be a shame. Worst case if we can't suppress the warnings is that I separate the failing parts to a different test that can be ignored on its own so that we don't completely lose sanitizer testing on this test.

biddisco:
I've filed an issue with mpich with a reproducer https://github.com/pmodels/mpich/issues/7187 which shows there is a problem, though the reproducer does not hang on exit as our tests do. I will wait for comments on the reproducer before investigating further.

msimberg:
> I was expecting to see changes to the thread sanitizer workflow - but here there are changes to the ubsan github workflow. I don't really understand all the syntx of the yaml workfow files, so I trust this is as expected ...

There are no changes expected to the thread sanitizer workflow. It stays with MPICH since OpenMPI completely fails to initialize with thread sanitizer loaded.

The ubsan workflow you're referring to (which includes address and leak sanitizer) was changed from MPICH to OpenMPI, so that we can keep the error handling tests enabled there (which would otherwise be disabled with MPICH now, as in the thread sanitizer workflow).

biddisco:
hmmm. ok, it was the thread sanitizer check that was failing for me, so I assumed it needed to be switched from openmpi to mpich, but since you removed a bunch of commits, maybe this was actually the original default and now it is as it used to be

msimberg:
> hmmm. ok, it was the thread sanitizer check that was failing for me, so I assumed it needed to be switched from openmpi to mpich, but since you removed a bunch of commits, maybe this was actually the original default and now it is as it used to be

Exactly, MPICH is already being used for both tsan and asan on main (so I removed the change in this PR to change from MPICH to OpenMPI which wasn't useful for tsan).

biddisco (approved):
LGTM, lets see if dashboard passes tests

PR #1277, opened 2024-10-22
Update acknowledgements in documentation

msimberg:
- Updates acknowledgement to HPX.
- Adds ETH Zurich/CSCS/PASC acknowledgement for funding.


msimberg:
The result looks like this:
2024-10-22-150646_844x573_scrot

PR #1280, opened 2024-10-22
Remove uses of `aligned_storage_t`

msimberg:
`aligned_storage(_t)` is deprecated starting with C++23. Replace uses with `alignas(...) unsigned char[...]`.


msimberg:
I've moved out the reenabling of SBO in `any_sender` to #1281.

PR #1281, opened 2024-10-23
Reenable SBO in `any_sender` and friends

msimberg:
Attempting to reenable SBO without any UB, but this needs more testing. We disabled SBO in the past because it was causing issues at least on LUMI, and I haven't re-tested it with these changes yet.

Based on #1280.


msimberg:
I've rebased this on top of #1379, which should be merged soon.

PR #1286, opened 2024-10-23
Add documentation for the `unpack` sender adaptor

msimberg:


biddisco (approved):
LGTM

RMeli (approved):

libs/pika/execution/include/pika/execution/algorithms/unpack.hpp
```diff
@@ -211,5 +211,16 @@ namespace pika::execution::experimental {
...
             return detail::partial_algorithm<unpack_t>{};
         }
-    } unpack{};
+    };
+
+    /// \brief Transforms a sender of tuples into a sender of the elements of the tuples.
+    ///
+    /// Sender adaptor that takes a sender of tuple-likes and returns a sender where each tuple-like
+    /// has been unpacked into its elements, similarly to `std::apply`. Each completion signature
+    /// must send exactly one tuple-like, not zero or more than one. The adaptor does not unpack
```
albestro, line 220: Just checking my understanding: does it mean that it is able to unpack multiple tuple-likes at once and the completion signature should have the union of all the elements of all tuple-likes given?

As far as I know `std::apply` just accept a single tuple-like, so probably the "similarly to" sentence means (if answer to previous question is yes) that the transformation that `unpack` applies to each tuple-like is the same as the one done by `std::apply`, but it is just "similar" and not the same because `std::apply` works with just a single tuple-like.

But then it comes *"Each completion signature must send exactly one tuple-like"*, and I'm not sure anymore, because it sounds to me that `unpack` is "limited" just like `std::apply` to a single tuple-like in input, and if that's true, I might have misunderstood the first sentence above *"takes a sender of tuple-likes and returns a sender where each tuple-like has been unpacked ..."*.

Thanks for any clarification 😉

msimberg: Excellent question, thanks!

Your understanding about `std::apply` and a single tuple is the correct interpretation of the docs. I think I've struggled a bit with the distinction of sending multiple tuples in one completion signature and a single tuple in multiple different completion signatures. In

> Sender adaptor that takes a sender of tuple-likes and returns a sender where each tuple-like has been unpacked into its elements

I was trying to convey that the adaptor also accepts senders that have more than one `set_value` completion signature, hence the plural. Given your comment this was definitely not clear though. What do you think about using the singular in that sentence, i.e.

> Sender adaptor that takes a sender of a tuple-like and returns a sender where the tuple-like has been unpacked into its elements

but the add the following clarification:

> The predecessor sender can have any number of `set_value` completion signatures, each sending a single tuple
?

albestro: IMHO your new suggestion is super clear! 

> Sender adaptor that takes a sender of a tuple-like and returns a sender where the tuple-like has been unpacked into its elements.

This is the information that is generally needed to anyone using it, while the more "advanced" (at least for me) bit of information is left for the next sentence.

> The predecessor sender can have any number of set_value completion signatures, each sending a single tuple.

Which now is clear to me too, even if I don't fully grasp all the internals.

Thanks for the clarification!

msimberg: https://github.com/pika-org/pika/compare/43e57f0ed0e284126292d977b0c54935eaec1a7e..feba118ee0c6af608bab89fb14b224a72d40ac2b

PR #1288, opened 2024-10-24
Convert const West to East const format

biddisco:
Convert all files and add simple script for making conversion using

sed -i -E 's/const ([a-zA-Z0-9_<>,:]+)&/\1 const\&/g'

NB. the character set includes alphanumeric as well as "_<>,:" to allow const std::vector<std::vector, allocator> expressions to be correctly detected. Other might be needed for more complex parameter lists.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -114,6 +114,7 @@ PenaltyReturnTypeOnItsOwnLine: 20
 PointerAlignment: Left
 PPIndentWidth: 1
 ReflowComments: false
+QualifierAlignment: Right
 RequiresClausePosition: OwnLine
 RequiresExpressionIndentation: OuterScope
 SortIncludes:    true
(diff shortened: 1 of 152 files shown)


biddisco:
if approved, this could also be added to git hooks - perhaps clang-format supports this already ...

biddisco:
> if approved, this could also be added to git hooks - perhaps clang-format supports this already ...

from clang-format-14 onwards
`QualifierAlignment: Right`

biddisco:
closing because it would be easier to add a clang-format and let nature take its course y reformatting files each time they are changed/committed

biddisco:
reopening with clang format change instead of script

msimberg (changes requested):
Very nice, thank you!

Can you change the signature for `pika::start` in the docs to match east-const: https://github.com/pika-org/pika/blob/main/docs/api.rst#pikainithpp? The documentation build is failing because the docs signature doesn't the updated signature in the code anymore.

msimberg (approved):

PR #1292, opened 2024-10-25
Add documentation for `drop_operation_state`

aurianer:
Part of #49

### DIFF
--- a/examples/documentation/drop_operation_state_documentation.cpp
+++ b/examples/documentation/drop_operation_state_documentation.cpp
@@ -1,0 +1,44 @@
+//  Copyright (c) 2024 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <pika/execution.hpp>
+#include <pika/init.hpp>
+
+#include <cassert>
+#include <memory>
+#include <utility>
+
+int main(int argc, char* argv[])
+{
+    namespace ex = pika::execution::experimental;
+    namespace tt = pika::this_thread::experimental;
+
+    pika::start(argc, argv);
+    ex::thread_pool_scheduler sched{};
+
+    auto sp = std::make_shared<int>(42);
+    std::weak_ptr<int> sp_weak = sp;
+
+    auto s = ex::just(std::move(sp)) | ex::then([&](auto&&) { assert(sp_weak.use_count() == 1); }) |
+        // Even though the shared_ptr is no longer in use, it may be kept alive by the operation state
+        ex::then([&]() {
+            assert(sp_weak.use_count() == 1);
+            return 42;
+        }) |
+        ex::drop_operation_state() |
+        // Once drop_operation_state has been used, the shared_ptr is guaranteed to be released.
+        // Values are passed through the adaptor.
+        ex::then([&]([[maybe_unused]] int x) {
+            assert(sp_weak.use_count() == 0);
+            assert(x == 42);
+        });
+    tt::sync_wait(std::move(s));
+
+    pika::finalize();
+    pika::stop();
+
+    return 0;
+}
--- a/libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
@@ -236,7 +236,8 @@ namespace pika::drop_op_state_detail {
 }    // namespace pika::drop_op_state_detail
 
 namespace pika::execution::experimental {
-    inline constexpr struct drop_operation_state_t final
+
+    struct drop_operation_state_t final
     {
         template <typename Sender, PIKA_CONCEPT_REQUIRES_(is_sender_v<Sender>)>
         constexpr PIKA_FORCEINLINE auto PIKA_STATIC_CALL_OPERATOR(Sender&& sender)
@@ -248,5 +249,20 @@ namespace pika::execution::experimental {
         {
             return detail::partial_algorithm<drop_operation_state_t>{};
         }
-    } drop_operation_state{};
+    };
+
+    /// \brief Releases the operation state of the adaptor before signaling a connected receiver.
+    ///
+    /// Sender adaptor that takes any sender and returns a sender. Values received as references
+    /// from the predecessor sender will be copied before being passed on to successor senders.
+    /// Other values are passed on unchanged.
+    ///
+    /// The operation state of previous senders can hold on to allocated memory or values longer
+    /// than necessary which can prevent other algorithms from using those resources.
+    /// @p drop_operation_state can be used to explicitly release the operation state, and thus
+    /// associated resources, of previous senders.
+    ///
+    /// Added in 0.19.0.
+    inline constexpr drop_operation_state_t drop_operation_state{};
+
 }    // namespace pika::execution::experimental
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -81,6 +81,12 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
    :language: c++
    :start-at: #include
 
+.. doxygenvariable:: pika::execution::experimental::drop_operation_state
+
+.. literalinclude:: ../examples/documentation/drop_operation_state_documentation.cpp
+   :language: c++
+   :start-at: #include
+
 .. doxygenvariable:: pika::execution::experimental::split_tuple
 
 .. literalinclude:: ../examples/documentation/split_tuple_documentation.cpp
--- a/examples/documentation/CMakeLists.txt
+++ b/examples/documentation/CMakeLists.txt
@@ -4,8 +4,14 @@
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-set(example_programs drop_value_documentation hello_world_documentation init_hpp_documentation
-                     split_tuple_documentation unpack_documentation when_all_vector_documentation
+set(example_programs
+    drop_operation_state_documentation
+    drop_value_documentation
+    hello_world_documentation
+    init_hpp_documentation
+    split_tuple_documentation
+    unpack_documentation
+    when_all_vector_documentation
 )
 
 foreach(example_program ${example_programs})


examples/documentation/drop_operation_state_documentation.cpp
```diff
@@ -0,0 +1,34 @@
+//  Copyright (c) 2024 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <pika/execution.hpp>
+#include <pika/init.hpp>
+
+#include <assert.h>
```
msimberg, line 10:
```suggestion
#include <cassert>
```

examples/documentation/drop_operation_state_documentation.cpp
```diff
@@ -0,0 +1,34 @@
...
+    pika::start(argc, argv);
+    ex::thread_pool_scheduler sched{};
+
+    auto sp = std::make_shared<int>(42);
+    std::weak_ptr<int> sp_weak = sp;
+
+    auto s = ex::just(std::move(sp)) | ex::then([&](auto&&) { assert(sp_weak.use_count() == 1); }) |
+        ex::drop_operation_state() |
+        // If drop_operation state was not used, the shared pointer would stay alive
+        ex::then([&] { assert(sp_weak.use_count() == 0); });
```
msimberg, line 27: I realize the original test where this comes from is missing at least one more check, and I think this would be useful to add at least to the unit test, but I'll suggest it here:
```suggestion
    auto s = ex::just(std::move(sp)) |
        ex::then([&](auto&&) { assert(sp_weak.use_count() == 1); }) |
        // Even though the shared_ptr is no longer in use, it may be kept alive by the operation state
        ex::then([&]() { assert(sp_weak.use_count() == 1); }) |
        ex::drop_operation_state() |
        // Once drop_operation_state has been used, the shared_ptr is guaranteed to be released
        ex::then([&] { assert(sp_weak.use_count() == 0); });
```

I don't know if this is better, so please comment what you think: the example may also be rewritten with comments and asserts replaced by print statements saying something equivalent. I'm not sure which is easier for a user to understand. The shared_ptr may also be better demonstrated with a custom class that e.g. prints something in the destructor, but again, I'm not sure what's easiest for a new user to follow. At least shared_ptr is something that's in the standard and its behaviour is well documented.

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
```diff
@@ -248,5 +249,14 @@ namespace pika::execution::experimental {
         {
             return detail::partial_algorithm<drop_operation_state_t>{};
         }
-    } drop_operation_state{};
+    };
+
+    /// \brief Release any operation state coming before the adaptor before it signals a connected
+    ///  receiver.
```
msimberg, line 255: Possibly shorter:
```suggestion
    /// \brief Releases the operation state of the adaptor before signaling a connected receiver
```
?

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
```diff
@@ -248,5 +249,14 @@ namespace pika::execution::experimental {
...
+    ///
+    /// Sender adaptor that takes any sender, release any operation state coming before this adaptor
+    /// and returns a new sender.
```
msimberg, line 258:
```suggestion
    /// and returns a new sender.

```

msimberg, line 258: If one knows something about operation states one knows the following can't happen, but with the current wording this seems to imply an order:

- take any sender
- release the operation state
- return a new sender

which is not true (since there's no operation state yet at the point when the new sender is returned).

Perhaps you can separately describe what the sender adaptor does to completion signatures etc. (i.e. _almost_ nothing, except change lvalue references to rvalue references), and then separately describe what the adaptor does when it's signalled and before it signals a receiver?

msimberg (changes requested):
Thank you! Some suggestions for changes in comments.

examples/documentation/drop_operation_state_documentation.cpp
msimberg: I separated the unit test change into #1293.

msimberg: I'm also in-between. I do feel like showing the absence of `drop_operation_state` helps demonstrate what it's trying to achieve, because it's not always obvious that the operation state/values that are no longer used are _not_ released early.

This is very related: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3373r0.pdf. But I'm not sure including it in the documentation helps explain the current situation, since that proposal may or may not end up changing the standard eventually.

Really I'd like to hear from @albestro, @rasolca, and @RMeli as users what they find explains `drop_operation_state` best.

examples/documentation/drop_operation_state_documentation.cpp
```diff
@@ -0,0 +1,35 @@
...
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <pika/execution.hpp>
+#include <pika/init.hpp>
+
+#include <cassert>
+#include <utility>
+
```
RMeli, line 12:
```suggestion
#include <memory>

```
?

RMeli (approved):

examples/documentation/drop_operation_state_documentation.cpp
RMeli: I think the suggested change makes thing clearer, at least to me.

albestro (approved):
LGTM

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
```diff
@@ -248,5 +249,16 @@ namespace pika::execution::experimental {
...
         }
-    } drop_operation_state{};
+    };
+
+    /// \brief Releases the operation state of the adaptor before signaling a connected receiver.
+    /// The operation_state of previous senders can potentially hold on to a allocated memory longer
+    /// than necessary which would prevent other algorithm from accessing this resource.
+    ///
+    /// Sender adaptor that takes any sender and returns a sender. Values sent as references will be
+    /// copied.
```
albestro, line 259: Just checking: it just manipulate the operation state, but it proxies the sender that receives somehow? What I mean is, can I have something like this?

```cpp
ex::just(26) | ex::drop_operation_state() | ex::then([](int){});
```

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
```diff
@@ -248,5 +249,16 @@ namespace pika::execution::experimental {
         {
             return detail::partial_algorithm<drop_operation_state_t>{};
         }
-    } drop_operation_state{};
+    };
+
+    /// \brief Releases the operation state of the adaptor before signaling a connected receiver.
+    /// The operation_state of previous senders can potentially hold on to a allocated memory longer
+    /// than necessary which would prevent other algorithm from accessing this resource.
```
msimberg, line 256:
```suggestion
    /// The operation state of previous senders can hold on to allocated memory or values longer
    /// than necessary which can prevent other algorithms from using those resources.
```
?

msimberg: And to make it explicit, perhaps add this afterwards:
```suggestion
    /// drop_operation_state (not sure about the markup here to make it formatted correctly) can be used to explicitly release the operation state, and thus associated resources, of previous senders.
```
?

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
msimberg, line 259: I would maybe still keep this as the first thing in the detailed documentation (i.e. after the `\brief`) and then move the motivation for it at the end?

msimberg: Yep, the above is possible. The

> Values sent as references will be copied.

sentence is meant to explain this aspect. Do you think that's clear enough? We can also consider adding another example demonstrating this.

msimberg (changes requested):

libs/pika/init_runtime/include/pika/init_runtime/init_runtime.hpp
```diff
@@ -156,7 +156,7 @@ namespace pika {
     ///
     /// Until @ref pika::finalize has been called, @ref pika::stop will not return. This function
     /// exists to distinguish between the runtime being idle but still expecting work to be
-    /// scheduled on it and the runtime being idle and ready to be shutdown. Unlike @pika::stop,
+    /// scheduled on it and the runtime being idle and ready to be shutdown. Unlike @ref pika::stop,
```
msimberg, line 159: Thanks for spotting this! Do you mind removing this change here, since it looks like this isn't working the way it should (my fault). In a few cases we've used `@p` to make it monospace, but no link. In most other places we've used `@ref` but it doesn't make it monospace and it doesn't seem to link it correctly as well. I can look at this outside of this PR.

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
```diff
@@ -248,5 +249,19 @@ namespace pika::execution::experimental {
...
+    };
+
+    /// \brief Releases the operation state of the adaptor before signaling a connected receiver.
+    ///
+    /// Sender adaptor that takes any sender and returns a sender. Values sent as references will be
+    /// copied.
+    ///
+    /// The operation state of previous senders can hold on to allocated memory or values longer
+    /// than necessary which can prevent other algorithms from using those resources.
+    /// @a drop_operation_state can be used to explicitly release the operation state, and thus
```
msimberg, line 261: It's not right, but can you make this
```suggestion
    /// @ref drop_operation_state can be used to explicitly release the operation state, and thus
```
for now so it's consistent with other references. I'll look at this separately. I don't think we want this to be in italics.

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
albestro: Probably being a bit more explicit won't harm.

> Sender adaptor that takes any sender and returns a sender with the same values received (except for values sent as references which will be copied).

Not sure it is correct, but I think you might get my point about it.

For what concerns the example, if you feel it would not distract too much, you might think about passing around a sender of something instead of a sender of void in the already existing example. At least for me it would help distinguish between the operation state and the sender and make clear that one is filtered and the other pass through.

examples/documentation/drop_operation_state_documentation.cpp
```diff
@@ -0,0 +1,36 @@
...
+    pika::start(argc, argv);
+    ex::thread_pool_scheduler sched{};
+
+    auto sp = std::make_shared<int>(42);
+    std::weak_ptr<int> sp_weak = sp;
+
+    auto s = ex::just(std::move(sp)) | ex::then([&](auto&&) { assert(sp_weak.use_count() == 1); }) |
+        // Even though the shared_ptr is no longer in use, it may be kept alive by the operation state
+        ex::then([&]() { assert(sp_weak.use_count() == 1); }) | ex::drop_operation_state() |
```
albestro, line 27:
```suggestion
        ex::then([&] { assert(sp_weak.use_count() == 1); }) | ex::drop_operation_state() |
```

one or the other just for consistency.

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
albestro: That's exactly what I was thinking! Thanks!

I wouldn't care about the "special case" of references. What I was targeting was exactly what you did, i.e. differentiating the operation state from the values sent through.

libs/pika/init_runtime/include/pika/init_runtime/init_runtime.hpp
msimberg: https://github.com/pika-org/pika/pull/1298 fixes the above separately.

libs/pika/execution/include/pika/execution/algorithms/drop_operation_state.hpp
msimberg: I think removing the anchor would be good. `@ref` should work again after https://github.com/pika-org/pika/pull/1298, but `@p` is also still ok.

msimberg (approved):
Thank you!

PR #1294, opened 2024-10-28
Use shared pool of CUDA streams instead of thread-local pools

msimberg:
The idea is that the thread-local streams (three by default) may limit concurrency of kernels within a worker thread. Using a single but larger shared pool of streams should lead to more possible concurrency, especially if many independent kernels are submitted quickly after each other from the same worker thread. At the same time, having e.g. three thread-local streams per thread on a Grace CPU is fake concurrency: CUDA will not actually allow allow a concurrency of 3×72=216. Using a shared pool limits the number of created streams to something a bit more manageable, and does not scale unnecessarily with the number of worker threads. Finally, using the same streams concurrently from different threads should not create thread-safety issues.

This change does not noticeably change the performance on P100, MI250, or H100 GPUs.


msimberg:
On old Piz Daint (with P100s) this change does not seem to have a big impact on performance, which is good. I still want to try on H100s before I mark this as ready for review.

PR #1297, opened 2024-10-30
Add documentation for `require_started`

aurianer:
Part of https://github.com/pika-org/pika/issues/49

### DIFF
--- a/examples/documentation/CMakeLists.txt
+++ b/examples/documentation/CMakeLists.txt
@@ -9,6 +9,7 @@ set(example_programs
     drop_value_documentation
     hello_world_documentation
     init_hpp_documentation
+    require_started_documentation
     split_tuple_documentation
     unpack_documentation
     when_all_vector_documentation
@@ -32,3 +33,12 @@ foreach(example_program ${example_programs})
     pika_add_example_test("documentation" ${example_program} ${${example_program}_PARAMETERS})
   endif()
 endforeach()
+
+if(PIKA_WITH_TESTS AND PIKA_WITH_TESTS_EXAMPLES)
+  set_tests_properties(
+    tests.examples.documentation.require_started_documentation
+    PROPERTIES
+      PASS_REGULAR_EXPRESSION
+      "pika::execution::experimental::~require_started_sender: A require_started sender was never started"
+  )
+endif()
--- a/examples/documentation/require_started_documentation.cpp
+++ b/examples/documentation/require_started_documentation.cpp
@@ -1,0 +1,48 @@
+//  Copyright (c) 2024 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <pika/execution.hpp>
+#include <pika/init.hpp>
+
+#include <fmt/printf.h>
+
+#include <cassert>
+#include <cstdlib>
+#include <exception>
+#include <utility>
+
+int main(int argc, char* argv[])
+{
+    namespace ex = pika::execution::experimental;
+    namespace tt = pika::this_thread::experimental;
+
+    pika::start(argc, argv);
+    ex::thread_pool_scheduler sched{};
+
+    {
+        // require_started forwards values received from the predecessor sender
+        auto s = ex::just(42) | ex::require_started() |
+            ex::then([]([[maybe_unused]] auto&& i) { assert(i == 42); });
+        tt::sync_wait(std::move(s));
+    }
+
+    {
+        // The termination is ignored with discard, the sender is from the user's perspective
+        // rightfully not used
+        auto s = ex::just() | ex::require_started();
+        s.discard();
+    }
+
+    {
+        // The require_started sender terminates on destruction if it has not been used
+        auto s = ex::just() | ex::require_started();
+    }
+    assert(false);
+
+    pika::finalize();
+    pika::stop();
+    return 0;
+}
--- a/libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
+++ b/libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
@@ -439,7 +439,7 @@ namespace pika {
 #endif
 
     namespace execution::experimental {
-        inline constexpr struct require_started_t final
+        struct require_started_t final
         {
             template <typename Sender, PIKA_CONCEPT_REQUIRES_(is_sender_v<Sender>)>
             constexpr PIKA_FORCEINLINE auto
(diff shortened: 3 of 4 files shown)


libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
```diff
@@ -453,7 +453,22 @@ namespace pika {
...
+        /// \anchor require_started
+        /// \brief Diagnose if a sender has not been started and terminates on destruction. It
+        /// forwards the values of the predecessor sender.
+        ///
+        /// Sender adaptor that takes any sender and returns a sender.
+        ///
+        /// The destructor is by default terminating if the sender has not been connected or if the
+        /// operation state has not been started.
+        /// The operation state of a @ref require_started sender is allowed to not be started, but it
+        /// has to be explicitly asked for with discard.
```
msimberg, line 467:
```suggestion
        /// The destructor of the sender terminates it has not been connected. The destructor of the corresponding operation state also terminates if it has not been started.
        /// The @ref require_started sender sender is allowed to not be started if it has been explicitly requested with the @p discard member function.
```

Side note: I'm really not sure how we best document the `discard` member function. The sender type is not public... but we'd still like to document it somehow. Do you think it's enough the way you've mentioned it above?

msimberg (changes requested):

tools/pre-commit
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 # Copyright (c) 2019 The STE||AR-Group
+# Copyright (c) 2024 ETH Zurich
```
msimberg, line 4: Could you put this in a separate PR please?

examples/documentation/require_started_documentation.cpp
```diff
@@ -0,0 +1,54 @@
...
+            pika::stop();
+            std::exit(1);
+        });
+
+        auto rs3 = ex::just() | ex::require_started();
+    }
+    assert(false);
+
+    return 0;
+}
```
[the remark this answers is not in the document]
msimberg: Yes and no. I would actually put them at the end of main (i.e. after `assert(false)` and before `return 0`) where they belong under normal circumstances (I missed this). But I would remove them from the terminate call since at that point we just want to exit the program.

examples/documentation/require_started_documentation.cpp
```diff
@@ -0,0 +1,53 @@
...
+        // The termination is ignored with discard, the sender is from the user's perspective
+        // rightfully not used
+        auto s = ex::just() | ex::require_started();
+        s.discard();
+    }
+
+    {
+        // The require_started sender terminates on destruction if it has not been used
+        std::set_terminate([] {
+            fmt::print("std::terminate called: rs3 never connected or started\n");
```
msimberg, line 42: Hmm, should we actually remove this completely?
```suggestion
```

The comment already says this, and the log will print:
```
[2024-11-01 10:16:18.695] [pika] [error] [require_started.hpp:353] pika::execution::experimental::~require_started_sender: A require_started sender was never started
```

Note: the `algorithm_require_started_terminate` test prints `std::terminate called` so that there's something to check for in the output to make sure the termination handler was actually called.

I don't feel very strongly about this.

examples/documentation/require_started_documentation.cpp
```diff
@@ -0,0 +1,53 @@
...
+            pika::finalize();
+            pika::stop();
+            std::exit(1);
+        });
+
+        auto s = ex::just() | ex::require_started();
+    }
+    assert(false);
+
+    return 0;
```
msimberg, line 52:
```suggestion
            std::exit(1);
        });

        auto s = ex::just() | ex::require_started();
    }
    assert(false);
    
    pika::finalize();
    pika::stop();

    return 0;
```

libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
```diff
@@ -453,7 +453,22 @@ namespace pika {
             {
                 return detail::partial_algorithm<require_started_t>{};
             }
-        } require_started{};
+        };
+
+        /// \anchor require_started
+        /// \brief Diagnose if a sender has not been started and terminates on destruction. It
+        /// forwards the values of the predecessor sender.
```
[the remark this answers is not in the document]
msimberg: https://github.com/pika-org/pika/issues/1308

libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
```diff
@@ -453,7 +453,22 @@ namespace pika {
...
-        } require_started{};
+        };
+
+        /// \brief Diagnose if a sender has not been started and terminates on destruction. It
+        /// forwards the values of the predecessor sender.
+        ///
+        /// Sender adaptor that takes any sender and returns a new sender that sends the same values
+        /// as the predecessor sender.
+        ///
+        /// The destructor is by default terminating if the sender has not been connected or if the
```
msimberg, line 464:
```suggestion
        /// The destructor terminates if the sender has not been connected or if the
```
I would not mention the "default" since the throwing version isn't something we can do with `stdexec`.

libs/pika/execution/include/pika/execution/algorithms/require_started.hpp
msimberg: Yeah, I think the example is good, and no ideas at the moment about how to document the member function of an unspecified class...

examples/documentation/CMakeLists.txt
```diff
@@ -32,3 +33,8 @@ foreach(example_program ${example_programs})
     pika_add_example_test("documentation" ${example_program} ${${example_program}_PARAMETERS})
   endif()
 endforeach()
+
+set_tests_properties(
+  tests.examples.documentation.require_started_documentation PROPERTIES PASS_REGULAR_EXPRESSION
+                                                                        "terminate called"
+)
```
msimberg, line 40: I think this needs to be wrapped in `if (PIKA_WITH_TESTS AND PIKA_WITH_TESTS_EXAMPLES)` (https://app.circleci.com/pipelines/github/pika-org/pika/11893/workflows/05afdcaf-c1be-4eed-bca5-48369a22b124/jobs/131879).

msimberg (approved):
Thank you!

examples/documentation/CMakeLists.txt
```diff
@@ -32,3 +33,10 @@ foreach(example_program ${example_programs})
     pika_add_example_test("documentation" ${example_program} ${${example_program}_PARAMETERS})
   endif()
 endforeach()
+
+if(PIKA_WITH_TESTS AND PIKA_WITH_TESTS_EXAMPLES)
+  set_tests_properties(
+    tests.examples.documentation.require_started_documentation PROPERTIES PASS_REGULAR_EXPRESSION
+                                                                          "terminate called"
```
msimberg, line 40: libc++ prints a different message: https://github.com/pika-org/pika/actions/runs/11779938152/job/32809450059#step:7:410. Perhaps change this to match ` pika::execution::experimental::~require_started_sender: A require_started sender was never started` instead, since that's a message we control?

PR #1303, opened 2024-10-31
Relax regex for the `require_started_terminated` test

msimberg:
With APEX enabled, APEX's output may be interleaved between the logging messages and the terminate message, so allow for output to come in between the two.

Also allows any output at the end of the output.

Another follow-up to #1291 and #1296.


msimberg:
This one I will try to get properly through the merge queue to make sure I haven't missed any more CI configurations.

msimberg:
The APEX test failures are now fixed, merging manually.

PR #1304, opened 2024-10-31
Expand `execution` documentation

msimberg:
- Add more links to other implementations
- Mention that sender adaptors are CPOs
- Mention the namespace used by pika and the forwarding of stdexec functionality
- Mention that standardized adaptors are documented elsewhere

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -74,6 +74,19 @@ The ``pika/init.hpp`` header provides functionality to manage the pika runtime.
 ======================
 
 The ``pika/execution.hpp`` header provides functionality related to ``std::execution``.
+``std::execution`` functionality, including extensions provided by pika, is defined in the
+``pika::execution::experimental`` namespace. When the CMake option ``PIKA_WITH_STDEXEC`` is enabled,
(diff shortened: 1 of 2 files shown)


albestro (approved):

docs/api.rst
```diff
@@ -74,6 +74,19 @@ The ``pika/init.hpp`` header provides functionality to manage the pika runtime.
 ======================
 
 The ``pika/execution.hpp`` header provides functionality related to ``std::execution``.
+``std::execution`` functionality is defined in the ``pika::execution::experimental`` namespace. When
```
RMeli, line 77: It might be worth making it explicit here that the same namespace also provides additional functionality not provided by the standard?

RMeli (approved):

docs/api.rst
msimberg: Good idea. Do you think this is explicit enough: https://github.com/pika-org/pika/compare/5ca735f90ff3bb88c51c486fcd35230ad49df756..5d47c715712df0bdf522b855844681ac68f3813d?

RMeli: Yes, thanks.

PR #1307, opened 2024-11-01
Update pre-commit to use CI base image

aurianer:
- Simplifies the pre-commit script
- Use pika-ci-base image to have the updated clang-format and cmake-format version

The previous pre-commit script supposed that we had clang-format and cmake-format installed on the system. It also supposed that we update it each time we decide to upgrade the versions.
This PR is using the docker image used in CI, pulls it if it's not available locally. The requirement here is podman, but the user can also remove podman from the pre-commit script if he prefers, the simplified commands run within podman can be run locally

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-# Copyright (c) 2019 The STE||AR-Group
+# Copyright (c) 2024 ETH Zurich
 #
 # SPDX-License-Identifier: BSL-1.0
 # Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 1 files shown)


tools/pre-commit
```diff
@@ -11,67 +12,39 @@
...
+    exit_code=1
+}
+
+echo "Run clang-format"
+podman run -w /pika/source -v=$(git rev-parse --show-toplevel):/pika/source docker.io/pikaorg/pika-ci-base:latest \
+    git-clang-format -f --extensions cpp,hpp
+result_clang_format=$?
+
+echo "Run cmake-format"
+checksum=$(git diff --cached --name-only | grep -E '(CMakeLists.txt|\.cmake$)' | xargs -I {} sh -c 'md5sum {}')
```
msimberg, line 37:
```suggestion
checksum=$(git diff --cached --name-only | grep -E '(^CMakeLists.txt|\.cmake)(\.in)?$' | xargs -I {} sh -c 'md5sum {}')
```
? Not 100% sure about my regex-fu.

tools/pre-commit
```diff
@@ -11,67 +12,39 @@
...
+podman run -w /pika/source -v=$(git rev-parse --show-toplevel):/pika/source docker.io/pikaorg/pika-ci-base:latest \
+    bash -c "git diff --cached --name-only | grep -E '(CMakeLists.txt|\.cmake$)' | xargs cmake-format -i"
```
msimberg, line 39:
```suggestion
    bash -c "git diff --cached --name-only | grep -E '(^CMakeLists.txt|\.cmake)(\.in)?$' | xargs cmake-format -i"
```

tools/pre-commit
```diff
@@ -11,67 +12,39 @@
...
+checksum_after_format=$(git diff --cached --name-only | grep -E '(CMakeLists.txt|\.cmake$)' | xargs -I {} sh -c 'md5sum {}')
```
msimberg, line 40:
```suggestion
checksum_after_format=$(git diff --cached --name-only | grep -E '(^ CMakeLists.txt|\.cmake)(\.in)?$' | xargs -I {} sh -c 'md5sum {}')
```

tools/pre-commit
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
```
msimberg, line 1: Do you mind changing this while you're at it?
```suggestion
#!/usr/bin/env bash
```

tools/pre-commit
```diff
@@ -11,67 +12,39 @@
...
-    returncode=1
+function error_msg() {
+    echo "$1 formatted files. run git add -u and git commit again."
+    echo "run git commit --no-verify to bypass."
+    git status
+    exit_code=1
+}
+
+echo "Run clang-format"
+podman run -w /pika/source -v=$(git rev-parse --show-toplevel):/pika/source docker.io/pikaorg/pika-ci-base:latest \
```
msimberg, line 32: Possibly:
```suggestion
podman run -w /pika/source -v=$(git rev-parse --show-toplevel):/pika/source docker.io/pikaorg/pika-ci-base:27 \
```

It'll be right 99% of the time with `:latest` as well, but we have to update all the other `:27`s to `:28` anyway when bumping the image tag, so not the end of the world to update it here as well? On the other hand, CI will anyway reject the PR in the 1% cases. I'm fine either way.

I would maybe move the image name out into a variable though, just to make it more visible which image we're using and not having to repeat it.

msimberg (changes requested):
Some minor comments, looks good to me otherwise.

@biddisco might want to have a look as well since I think he's also using the pre-commit hook. If there's a preference for either we can also consider having both the old and the new version as alternatives (though only option is better of course).

tools/pre-commit
```diff
@@ -11,67 +12,39 @@
...
+    git-clang-format -f --extensions cpp,hpp
```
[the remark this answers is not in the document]
msimberg: Yep, looks like you're right. We still check for it in CI, so I don't think it's worth worrying too much about this corner case then. Manually formatting is not an improvement, so feel free to keep it this way. Please do add the CUDA extensions though.

tools/pre-commit
msimberg: Good point, that's fine by me.

msimberg (approved):

biddisco:
This is horrific - why would you require podman to be run to use a commit hook?

PR #1309, opened 2024-11-01
Expand and add more CUDA/HIP documentation

msimberg:
Adds documentation for classes that weren't yet documented and updates the existing documentation. Adds a section for `pika/cuda.hpp` to the API documentation.

Early stages, far from complete, but may require some discussion before continuing.


docs/api.rst
```diff
@@ -5,6 +5,8 @@
     Distributed under the Boost Software License, Version 1.0. (See accompanying
     file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+:tocdepth: 3
+
```
msimberg, line 9: I'm setting this explicitly to avoid having _everything_ show up in the table of contents on the right hand side.

- level 1 is basically only the page title "API reference"
- level 2 adds the header sections (`pika/execution.hpp` etc.)
- level 3 adds the functions/classes in each header
- level 4 adds member functions of classes (or parameters of functions, only with sphinx-immaterial)
- level 5 adds pre/post conditions of member functions (only with sphinx-immaterial)

I'm leaning towards 3 or 4 with a slight preference for 3, which is what I've currently chosen.
-

docs/api.rst
```diff
@@ -104,3 +105,98 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
...
+   to avoid using ``cuda`` even when HIP is enabled. If you have feedback on a rename or just want
+   to follow along, please see that issue.
+
+.. warning::
+   At the moment, ``nvcc`` can not compile stdexec headers. Of the CUDA compilers, only ``nvc++`` is
+   able to compile stdexec headers. If you have stdexec support enabled in pika, either ensure that
+   ``.cu`` files do not include stdexec headers, or use ``nvc++`` to compile your application.
+   However, ``nvc++`` does not officially support compiling device code. Use at your own risk.
+
+   For HIP there are no known restrictions.
```
msimberg, line 131: Do you think these are useful notes or unnecessary noise?

docs/api.rst
```diff
@@ -104,3 +105,98 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
...
+   worker thread and thus block progress of other work.
+
+The following example gives an overview of using the above CUDA functionalities in pika:
+
+.. literalinclude:: ../examples/documentation/cuda_overview_documentation.cu
+   :language: c++
+   :start-at: #include
+
+TODO: Note that while cuda_pool gives direct access to streams and handles, the intended usage is to
+access them via the scheduler and sender adaptors.
```
msimberg, line 152: To do.

docs/api.rst
```diff
@@ -104,3 +105,98 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
...
+
+.. doxygenclass:: pika::cuda::experimental::cuda_pool
+.. doxygenclass:: pika::cuda::experimental::cuda_stream
+.. doxygenclass:: pika::cuda::experimental::cublas_handle
+.. doxygenclass:: pika::cuda::experimental::locked_cublas_handle
+.. doxygenclass:: pika::cuda::experimental::cusolver_handle
+.. doxygenclass:: pika::cuda::experimental::locked_cusolver_handle
```
msimberg, line 159: I'm at the moment not planning to provide examples for these individually for two reasons:
- `cuda_pool` and friends are covered quite well in the overview example (i.e. create it and pass it to a `cuda_scheduler`)
- most of these should generally not be used directly, but the streams and handles should be accessed with the sender adaptors below

Do you think this is ok (leaving out examples here)? If yes, perhaps I should move them to the bottom of the section?

docs/api.rst
```diff
@@ -104,3 +105,98 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
...
+      :start-at: #include
+
+.. doxygenclass:: pika::cuda::experimental::cuda_scheduler
+
+..
+   .. literalinclude:: ../examples/documentation/cuda_scheduler_documentation.cpp
+      :language: c++
+      :start-at: #include
+
+.. doxygenvariable:: pika::cuda::experimental::then_with_stream
```
msimberg, line 173: This one is a bit weirdly documented at the moment, since we document the variable, but not the call operators of the `then_with_stream_t` class. I'm between having an example, which is usually clear enough on its own, and just documenting the call operators explicitly. Currently I refer e.g. to `\p f` as the callable passed to the adaptor, but the user has no information on how that is passed to the adaptor (not much of a problem for `then_with_stream` but a bigger problem for `then_with_cublas` since that also takes a pointer mode).

Should I include both an example and document the call operators? If we document the call operators should we do the same for `drop_value`, `drop_operation_state`, etc.?

examples/documentation/CMakeLists.txt
```diff
@@ -5,11 +5,13 @@
 # file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 set(example_programs
+    # cuda_overview_documentation # TODO
     drop_operation_state_documentation
     drop_value_documentation
     hello_world_documentation
     init_hpp_documentation
     split_tuple_documentation
+    # then_with_stream_documentation # TODO
```
msimberg, line 14: To do. These currently don't compile.

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
```diff
@@ -52,15 +64,24 @@ namespace pika::cuda::experimental {
...
     ///
-    /// The pool initializes a set of CUDA (thread-local) streams on
-    /// construction and provides access to the streams in a round-robin
-    /// fashion. The pool is movable and copyable with reference semantics.
-    /// Copies of a pool still refer to the original pool of streams.
+    /// The pool is movable and copyable with reference semantics.  Copies of a pool still refer to
+    /// the original pool of streams. A moved-from pool can't be used, except to check if it is
+    /// valid with \ref valid().
+    ///
+    /// The pool is equality comparable and formattable.
```
msimberg, line 84: Should we keep it simple and document this like above, or should we explicitly add `operator==` etc. to the documentation? I'm leaning towards keep it simple, but curious to hear what others think.

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
```diff
@@ -135,25 +156,64 @@ namespace pika::cuda::experimental {
...
+        /// \brief Move constructor.
         PIKA_NVCC_PRAGMA_HD_WARNING_DISABLE
         cuda_pool(cuda_pool&&) = default;
+        /// \brief Copy constructor.
         PIKA_NVCC_PRAGMA_HD_WARNING_DISABLE
         cuda_pool(cuda_pool const&) = default;
+        /// \brief Move assignment operator.
         PIKA_NVCC_PRAGMA_HD_WARNING_DISABLE
         cuda_pool& operator=(cuda_pool&&) = default;
+        /// \brief Copy assignment operator.
```
msimberg, line 183: I'm not sure there's any value in documenting these like above. Should we just leave the docstrings out and say it's copyable/movable etc. and describe the semantics in the class docstring?

libs/pika/async_cuda_base/include/pika/async_cuda_base/cublas_handle.hpp
```diff
@@ -27,6 +32,7 @@ namespace pika::cuda::experimental {
         static PIKA_EXPORT cublasHandle_t create_handle(int device, whip::stream_t stream);
 
     public:
+        /// TODO: How to best document constructor and other special member functions.
         PIKA_EXPORT cublas_handle();
```
msimberg, line 36: To do. Document that this is default constructible and what the state of a default-constructed handle is.

libs/pika/async_cuda/include/pika/async_cuda/then_on_host.hpp
```diff
@@ -245,5 +235,16 @@ namespace pika::cuda::experimental {
...
+    /// NOTE: This is not a customization of pika::execution::experimental::then.
+    /// It retains the cuda_scheduler execution context from the predecessor
+    /// sender, but does not run the continuation on a CUDA device. Instead, it
+    /// runs the continuation in the polling thread used by the cuda_scheduler on
+    /// the CPU. The continuation is run only after synchronizing all previous
+    /// events scheduled on the cuda_scheduler. Blocking in the callable given to
+    /// then_on_host blocks other work scheduled on cuda_scheduler from
+    /// completing. Heavier work should be transferred to a host scheduler as
+    /// soon as possible.
+    inline constexpr then_on_host_t then_on_host{};
```
msimberg, line 249: I'm considering removing this function as we've so far not had a use for it, and the safer option is to explicitly transfer to a new task if one wants to run something on the host. What do you think about removing it?

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
```diff
@@ -633,17 +626,20 @@ namespace pika::cuda::experimental {
...
-    /// has been used the last time by f.
-    inline constexpr struct then_with_cublas_t final
+    /// When the predecessor sender completes, calls \p f with a CUDA stream as the last argument
+    /// after other values sent by the predecessor sender. This adaptor can only be used when the
+    /// completion scheduler is a \ref cuda_scheduler. Other work may be scheduled concurrently on
+    /// the stream passed to \p f. Values sent by the predecessor sender are passed as references to
+    /// \p f and kept alive until the work submitted by \p f to the stream is completed. \p f may
+    /// return as soon as work has been submitted, and a connected receiver will be signaled only
+    /// once the kernels submitted to the stream have completed.
+    inline constexpr then_with_stream_t then_with_stream{};
```
msimberg, line 640: See https://github.com/pika-org/pika/pull/1309/files#r1825762317.

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
```diff
@@ -661,17 +657,19 @@ namespace pika::cuda::experimental {
...
-    /// thread-local and f may not yield a pika thread until after the handle
-    /// has been used the last time by f.
-    inline constexpr struct then_with_cusolver_t final
+    /// This sender is intended to be used to submit work using a cuBLAS handle. The stream
+    /// associated to the handle may also be used to submit work. The handle is accessed through a
+    /// \ref locked_cublas_handle and \p f should return as quickly as possible to avoid blocking
+    /// other work from using the handle.
+    ///
+    /// The behaviour of synchronization and lifetimes are the same as for \ref then_with_stream.
+    inline constexpr then_with_cublas_t then_with_cublas{};
```
msimberg, line 670: See https://github.com/pika-org/pika/pull/1309/files#r1825762317.

libs/pika/async_cuda_base/include/pika/async_cuda_base/cuda_stream.hpp
```diff
@@ -49,9 +60,25 @@ namespace pika::cuda::experimental {
...
         PIKA_EXPORT whip::stream_t get() const noexcept;
+
+        /// \brief Get the device of the stream.
+        ///
+        /// \return the device of the stream.
         PIKA_EXPORT int get_device() const noexcept;
+
+        /// \brief Get the priority of the stream.
+        ///
+        /// \return the priority of the stream.
```
msimberg, line 76: Should we just do:
```suggestion
        /// \brief Get the priority of the stream.
```
for simple functions like these?

docs/api.rst
```diff
@@ -104,3 +105,98 @@ The ``pika/execution.hpp`` header provides functionality related to ``std::execu
 .. literalinclude:: ../examples/documentation/when_all_vector_documentation.cpp
    :language: c++
    :start-at: #include
+
```
msimberg, line 108: The API reference page is starting to get quite long. We could consider splitting it up to have one header per page. I'm still slightly in favour of keeping it on one page for Ctrl-F-ability and clicking on references not changing the page, but wouldn't object to splitting it up either.

docs/api.rst
msimberg: Stick to level 3, do not add too much information in table of contents.

docs/api.rst
msimberg: Useful information, keep.

docs/api.rst
msimberg: Explore separately from this PR what documentation would look like with e.g. documentation for call operator on CPO type.

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
msimberg: Keep it simple. Unless there's special behaviour for special member functions, just mention here that they exist/don't exist.

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
msimberg: Remove these, unless there's special behaviour to document.

libs/pika/async_cuda_base/include/pika/async_cuda_base/cublas_handle.hpp
msimberg: Document this saying it's an invalid handle.

libs/pika/async_cuda/include/pika/async_cuda/then_on_host.hpp
msimberg: Remove `then_on_host`.

libs/pika/async_cuda_base/include/pika/async_cuda_base/cuda_stream.hpp
msimberg: Simplify for getters and setters.

libs/pika/async_cuda/include/pika/async_cuda/then_on_host.hpp
msimberg: https://github.com/pika-org/pika/pull/1319

docs/api.rst
msimberg: I ended up trying to document the types of the CPOs as well, and ended up with something that looks like this:

2024-11-07-153409_971x952_scrot

i.e. first type, then CPO itself, and then the example. I tried to keep it quite short, referring to the CPO documentation for more details. The important part for me was to get the signature of the CPO visible, so I've documented the two `operator()`s and their parameters.

@albestro, @aurianer, @rasolca, @RMeli thoughts?

msimberg:
I welcome more feedback on this PR now. I feel like some of the examples I've created are a bit forced/artificial in the name of showing off what's possible with the adaptors, so if you have suggestions on how to improve change them that's particularly welcome. I've also now documented the types of the CUDA/HIP sender adaptors so that we can document the parameters of them. I think it looks decent enough, but interested in hearing feedback from others.

examples/documentation/cuda_overview_documentation.cu
```diff
@@ -0,0 +1,90 @@
...
+
+        constexpr std::size_t n = 2048;
+        double* a = nullptr;
+        double* b = nullptr;
+        double* c = nullptr;
+        double alpha = 1.0;
+        double beta = 1.0;
+        whip::malloc(&a, sizeof(double) * n * n);
+        whip::malloc(&b, sizeof(double) * n * n);
+        whip::malloc(&c, sizeof(double) * n * n);
```
msimberg, line 59: I should probably initialize these with something reasonable... same in the other examples.

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
```diff
@@ -654,37 +664,59 @@ namespace pika::cuda::experimental {
...
+
+    /// \brief Sender adaptor which calls \p f with a cuSOLVER handle.
+    ///
+    /// This sender is intended to be used to submit work using a cuSOLVER handle. The stream
+    /// associated to the handle may also be used to submit work. The handle is accessed through a
+    /// \ref locked_cusolver_handle and \p f should return as quickly as possible to avoid blocking
+    /// other work from using the handle.
+    ///
+    /// The behaviour of synchronization and lifetimes are the same as for \ref then_with_stream,
+    /// except the the handle is passed as the first argument to match the typical function
```
albestro, line 719:
```suggestion
    /// except that the handle is passed as the first argument to match the typical function
```

libs/pika/async_cuda_base/include/pika/async_cuda_base/cublas_handle.hpp
```diff
@@ -27,19 +38,41 @@ namespace pika::cuda::experimental {
...
         PIKA_EXPORT explicit cublas_handle(cuda_stream const& stream);
         PIKA_EXPORT ~cublas_handle();
         PIKA_EXPORT cublas_handle(cublas_handle&&) noexcept;
         PIKA_EXPORT cublas_handle& operator=(cublas_handle&&) noexcept;
         PIKA_EXPORT cublas_handle(cublas_handle const&);
         PIKA_EXPORT cublas_handle& operator=(cublas_handle const&);
 
+        /// \brief Check if the handle is valid.
+        ///
+        /// \return true if the pool refers to a valid handle, false otherwise (e.g. if the handle
```
albestro, line 54: ?

```suggestion
        /// \return true if the handle refers to a valid handle, false otherwise (e.g. if the handle
```

libs/pika/async_cuda_base/include/pika/async_cuda_base/cusolver_handle.hpp
```diff
@@ -28,18 +39,38 @@ namespace pika::cuda::experimental {
...
         PIKA_EXPORT explicit cusolver_handle(cuda_stream const& stream);
         PIKA_EXPORT ~cusolver_handle();
         PIKA_EXPORT cusolver_handle(cusolver_handle&&) noexcept;
         PIKA_EXPORT cusolver_handle& operator=(cusolver_handle&&) noexcept;
         PIKA_EXPORT cusolver_handle(cusolver_handle const&);
         PIKA_EXPORT cusolver_handle& operator=(cusolver_handle const&);
 
+        /// \brief Check if the handle is valid.
+        ///
+        /// \return true if the pool refers to a valid handle, false otherwise (e.g. if the handle
```
albestro, line 55: ?
```suggestion
        /// \return true if the handle refers to a valid handle, false otherwise (e.g. if the handle
```

libs/pika/async_cuda/include/pika/async_cuda/then_with_stream.hpp
```diff
@@ -654,37 +664,59 @@ namespace pika::cuda::experimental {
...
-    /// thread-local and f may not yield a pika thread until after the handle
-    /// has been used the last time by f.
-    inline constexpr struct then_with_cusolver_t final
+    /// This sender is intended to be used to submit work using a cuBLAS handle. The stream
+    /// associated to the handle may also be used to submit work. The handle is accessed through a
+    /// \ref locked_cublas_handle and \p f should return as quickly as possible to avoid blocking
+    /// other work from using the handle.
+    ///
+    /// The behaviour of synchronization and lifetimes are the same as for \ref then_with_stream,
+    /// except the the handle is passed as the first argument to match the typical function
```
albestro, line 685: ?

```suggestion
    /// except that the handle is passed as the first argument to match the typical function
```

libs/pika/async_cuda_base/include/pika/async_cuda_base/cublas_handle.hpp
msimberg: Thanks for the detailed look! Fixed in the latest push.

examples/documentation/then_with_cublas_documentation.cu
```diff
@@ -0,0 +1,120 @@
...
+        constexpr std::size_t n = 2048;
+        gpu_data a{n * n};
+        gpu_data b{n * n};
+        gpu_data c{n * n};
+        double alpha = 1.0;
+        double beta = 1.0;
+
+        auto s = ex::just(std::move(a), std::move(b), std::move(c)) | ex::continues_on(cuda_sched) |
+            cu::then_with_stream([](auto& a, auto& b, auto& c, whip::stream_t stream) {
+                init<<<256, n * n / 256, 0, stream>>>(a.get(), b.get(), c.get(), n * n);
```
RMeli, line 100:
```suggestion
                init<<<n * n / 256, 256, 0, stream>>>(a.get(), b.get(), c.get(), n * n);
```
?

libs/pika/async_cuda_base/include/pika/async_cuda_base/cuda_stream.hpp
```diff
@@ -15,12 +15,19 @@
...
-    /// and destroys it on destruction. Is movable and copiable. A moved-from
-    /// stream holds the default stream. A copied stream uses the properties
-    /// from the given stream and creates a new stream.
+    /// An RAII wrapper for a CUDA stream which creates a stream on construction and destroys it on
+    /// destruction. Is movable and copyable. A moved-from stream holds the default stream. A copied
+    /// stream uses the properties from the given stream and creates a new stream.
+    ///
+    /// Equality comparable and formattable.
+    ///
+    /// When accessing the underlying stream, https://github.com/eth-cscs/whip is used for
```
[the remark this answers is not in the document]
msimberg: This will already be rendered as a clickable link in the final documentation, but do you mean something like `[whip](https://github.com/eth-cscs/whip)`?

docs/api.rst
```diff
@@ -116,3 +117,83 @@ All sender adaptors are `customization point objects (CPOs)
...
+.. _header_pika_cuda:
+
+``pika/cuda.hpp``
+=================
```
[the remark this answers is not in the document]
msimberg: I agree this might be a good idea. Really this is a symptom of us still calling it `pika/cuda.hpp` etc., but that won't be fixed now... Independently I've toyed with the idea of adding some extra context to the header sections, e.g.

- Runtime initialization etc. (`pika/init.hpp`)
- `std::execution` support (`pika/execution.hpp`)
- CUDA/HIP support (`pika/cuda.hpp`)
- etc.

So I think you've probably at least confirmed that adding some extra context to the headings is useful. I'll think about moving the detailed explanation/note further up as well (though I'm leaning towards keeping it where it is at the moment).

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
```diff
@@ -135,10 +165,21 @@ namespace pika::cuda::experimental {
...
+        /// \brief Construct a pool of CUDA streams and handles.
+        ///
+        /// \param device the CUDA device used for scheduling work
+        /// \param num_normal_priority_streams_per_thread the number of normal priority streams per
+        /// thread
+        /// \param num_high_priority_streams_per_thread the number of high priority streams per
+        /// thread
+        /// \param flags flags used to construct CUDA streams
+        /// \param num_cublas_handles the number of cuBLAS handles to create for the whole pool
+        /// \param num_cusolver_handles the number of cuSOLVER handles to create for the whole pool
```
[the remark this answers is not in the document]
msimberg: I'll try :upside_down_face: They are "arbitrary"/decent average value that seem to not perform too badly with DLA-Future.

libs/pika/async_cuda_base/include/pika/async_cuda_base/cuda_stream.hpp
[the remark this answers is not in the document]
RMeli: Either is fine, I forgot links are automatically clickable.

docs/api.rst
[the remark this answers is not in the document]
RMeli: I think adding some extra context to the header sections is a good idea. If you end up doing that, then there is no need to move the note further up, since one would see the context instead of just `pika/cuda.hpp`.

docs/api.rst
```diff
@@ -116,3 +117,83 @@ All sender adaptors are `customization point objects (CPOs)
...
+
+.. literalinclude:: ../examples/documentation/cuda_overview_documentation.cu
+   :language: c++
+   :start-at: #include
+
+.. note::
+   pika uses `whip <https://github.com/eth-cscs/whip>`__ internally for portability between CUDA and
+   HIP. However, users of pika are not forced to use whip as whip only creates aliases for CUDA/HIP
+   types and enumerations. whip is thus compatible with directly using the types and enumerations
+   provided by CUDA/HIP.
```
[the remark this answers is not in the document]
msimberg: Maybe also mention that we don't use anything like whip for {cu,roc}{BLAS,SOLVER}...

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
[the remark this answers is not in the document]
RMeli: I think that would be enough. It's good for people interested in pika but without any knowledge of DLA-Future to know these values are based on a production project.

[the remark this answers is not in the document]
msimberg: I've updated the documentation for the parameters in https://github.com/pika-org/pika/pull/1309/commits/f267193fb0721a22b235785b48fe49c8692c0f0b. I'm anticipating #1294 with the documentation here. I'm still documenting the per-thread parameters here, but I'd like to get #1294 merged first.

Do you find the warnings sufficient? We're essentially making a silent API change with #1294, and I'm worried that it's too silent (especially if one doesn't read documentation or release notes)... The main alternative I can think of is to make the parameters strong types instead of plain integers, thus requiring a explicit conversion at the callsite.

docs/api.rst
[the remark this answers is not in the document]
msimberg: I've moved the note, and mentioned {cu,roc}{BLAS,SOLVER} in https://github.com/pika-org/pika/pull/1309/commits/109ca791425cc2fbbf4d87d19aa16dce4e42786d.

docs/api.rst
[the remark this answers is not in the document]
msimberg: I've added some more descriptive headings in https://github.com/pika-org/pika/pull/1309/commits/47a0686ecba86d49268aa113ebe0f798f76e0249.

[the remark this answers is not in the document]
RMeli: LGTM, thanks!

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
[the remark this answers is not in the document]
msimberg: Ok, thanks, good to hear you're not too concerned. This kind of stuff is indeed why we've stuck to 0.X and semver, so I suppose I should exercise that right... it's just an uneasy feeling knowing that everything will compile just fine, but possibly perform differently.

> too overkill for the reasons above.

Yeah, I'd like to avoid that if possible.

However, I am planning on having a warning about this in DLA-Future (and basically introducing a new configuration option), so there the change should definitely not be silent.

libs/pika/async_cuda_base/include/pika/async_cuda_base/cuda_stream.hpp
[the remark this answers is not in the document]
msimberg: I changed this to a markdown link by the way. It looks nicer in the text.

docs/api.rst
msimberg: I'm going to keep this with the struct documentation. Feel free to suggest improvements in the future.

libs/pika/async_cuda/include/pika/async_cuda/cuda_pool.hpp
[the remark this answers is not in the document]
msimberg: I'll keep this as is (#1294 has also been merged).

PR #1311, opened 2024-11-01
Update documentation to use sphinx-immaterial theme

msimberg:
The current theme, sphinx-material, isn't very actively developed: https://github.com/bashtage/sphinx-material/issues/141. The author even suggests using sphinx-immaterial: https://github.com/jbms/sphinx-immaterial. This PR updates the documentation to use sphinx-immaterial, but relies on https://github.com/jbms/sphinx-immaterial/pull/338 at the moment for some newer features.

A few improvements:
- automatic dark mode (as well as manual override)
- interactive search
- (nicer) version-added roles (available in vanilla sphinx/rst, but the look better with sphinx-immaterial)
- generally slightly more polished style and responsive

I've also changed all the logos in the Acknowledgements to pngs, optimized them, and added a white background to the ETHZ logo (it was previously an svg with transparent background, leading to black text on black background in dark mode).


docs/conf.py
```diff
@@ -33,43 +34,103 @@
...
+        # "search.highlight"
+
+        # Share button for a search.
+        "search.share"
+
+        # "When navigation integration for the table of contents is enabled, it is always rendered
+        # as part of the navigation sidebar on the left." Does not seem to work? See
+        # https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-integration.
+        # "toc.integrate",
+    ],
```
msimberg, line 105: To do: remove features we don't need/use.

docs/conf.py
```diff
@@ -33,43 +34,103 @@
...
+        },
+        {
+            "media": "(prefers-color-scheme: dark)",
+            "scheme": "slate",
+            "primary": "black",
+            "accent": "light-blue",
+            "toggle": {
+                "icon": "material/lightbulb-outline",
+                "name": "Switch to system preference",
+            },
```
msimberg, line 133: To do: decide on colors. I've gone for a slightly simpler "white top bar in light mode" and "black top bar in dark mode" compared to the current docs, but happy to choose some other colors as well if you think that's nicer. The "accent" color still adds a bit of color.

msimberg:
To view the documentation locally, download the artifact created in the "Documentation" workflow (see workflow for a link), extract the archive, and serve it e.g. with `python -m http.server --directory <extracted-dir>`. You can also simply open `index.html` from the archive, but e.g. search won't work unless served with a webserver.

docs/requirements.txt
```diff
@@ -2,4 +2,5 @@
 sphinx
 breathe
 recommonmark
-sphinx_material
+# TODO: Do we want to use the not-yet-released branch? Pin the commit?
+sphinx_immaterial @ git+https://github.com/jbms/sphinx-immaterial.git@merge-upstream
```
msimberg, line 6: To do.

docs/conf.py
msimberg: I've removed the unused features.

docs/conf.py
msimberg: Keeping this black and white unless there are objections. I'm also open to changing it later as well.

docs/requirements.txt
msimberg: I've updated this to use a pinned commit from the `merge-upstream` branch since it seems to work fine. I'll keep an eye on the branch and try to update if/when it gets merged.

PR #1322, opened 2024-11-07
Don't throw exception when `PIKA_MPI_ENABLE_POOL` is enabled without MPI initialized

msimberg:
Downgrades the exception to a warning, unless the pool creation mode is `mode_force_create` (in which case it's still an error to not have MPI initialized).


msimberg:
I would like to include this still in 0.30.0, plus test this change with DLA-Future, as soon as possible.

PR #1325, opened 2024-11-11
Remove `PIKA_MOVE` and `PIKA_FORWARD`

msimberg:
The macros were originally introduced to avoid the template instantiation overhead of `std::move` and `std::forward`, as well as the debug bloat. Both clang and GCC should nowadays treat them as builtins, though perhaps not on all supported compiler versions. Since the original change was simply a build-time optimization I think it's fair to go back to the `std::` functions again.

The benefit is that tooling like clang-tidy can warn about improper uses of the functions.

The one downside of using `std::move` over `PIKA_MOVE` (which is simply a cast to a non-const r-value reference) is that `std::move` will happily create `const&&` types, where that is almost never intended for a move.


msimberg:
This PR has lower priority than others, so in case of conflicts let's try to merge others first and push this when it's a bit more quiet on the PR-front.

PR #1328, opened 2024-11-13
Revert #1307 but keep podman-based pre-commit hook in pre-commit-podman

### LINKED ISSUE #1327: pre-commit hook broken for git worktrees
PR #1307 introduced podman to perform format checks using a podman image, but the mounting of the source tree inside the container doesn't work as is for git worktrees
for example, my own setup has a number of branches checked out in different trees as follows
```
elf:~/src/pika/workdir (transform-mpi-lifetimes *$%=): git worktree list
/home/biddisco/src/pika/main                 5827d3a5a655 [main]
/home/biddisco/src/pika/background_threads   74f62b8c112b [background_threads]
/home/biddisco/src/pika/clangformat          318f199d4207 [stringformat]
/home/biddisco/src/pika/cmake_vars           6fb345f34143 [cmake_vars]
/home/biddisco/src/pika/condition_variables  8fff8240051e [condition_variables]
/home/biddisco/src/pika/debug_fmt            3502e9b62ca2 [debug_fmt]
/home/biddisco/src/pika/limiting_scheduler   df3a70046eca [limiting_scheduler]
/home/biddisco/src/pika/moodycamel           9ca5a0afcfca [boost]
/home/biddisco/src/pika/mpi                  ba27a20e7696 [cuda-annotation]
/home/biddisco/src/pika/workdir              38328c003c44 [transform-mpi-lifetimes]
```
but the `.git` entry in 9 out of the 10 is really just a path to follow to find the real `.git` folder 
```
elf:~/src/pika/workdir (transform-mpi-lifetimes *$%=): cat /home/biddisco/src/pika/mpi/.git
gitdir: ../main/.git/worktrees/mpi
```
and only the "main" (actually first, not necessarily main branch) tree created contains the real `.git` folder. When mounting the source tree and then executing git commands from the hook, unfortunately, the actual git folder is not mounted and so commands fail.

msimberg:
Fixes #1327.

Also updates `/bin/bash` to `/usr/bin/env bash` in `pre-commit` hook, as in #1307 (https://github.com/pika-org/pika/pull/1307#discussion_r1825608405).

### DIFF
--- a/tools/pre-commit-podman
+++ b/tools/pre-commit-podman
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2024 ETH Zurich
+#
+# SPDX-License-Identifier: BSL-1.0
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+# ----------------------------------------------------------------
+# simple pre commit hook script to check that *.cpp and *.hpp files
+# are correctly clang-formatted and that CMakeLists.txt and *.cmake
+# files are cmake-formatted
+
+# To install this hook, symlink this hook to your git hooks as follows
+# (note that the extra ../../ in the path is because git runs the hook
+# from the .git/hooks directory, so the symlink has to be redirected)
+# ln -s -f ../../tools/pre-commit-podman .git/hooks/pre-commit
+
+# To use this hook, you must have podman installed.
+
+# @todo : add support for the pika inspect tool to be run as well
+
+function error_msg() {
+    echo "$1 formatted files. run git add -u and git commit again."
+    echo "run git commit --no-verify to bypass."
+    git status
+    exit_code=1
+}
+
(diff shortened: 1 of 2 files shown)


tools/pre-commit-podman
```diff
@@ -0,0 +1,52 @@
...
+
+# ----------------------------------------------------------------
+# simple pre commit hook script to check that *.cpp and *.hpp files
+# are correctly clang-formatted and that CMakeLists.txt and *.cmake
+# files are cmake-formatted
+
+# To install this hook, symlink this hook to your git hooks as follows
+# (note that the extra ../../ in the path is because git runs the hook
+# from the .git/hooks directory, so the symlink has to be redirected)
+# ln -s -f ../../tools/pre-commit .git/hooks/pre-commit
```
biddisco, line 17:
```suggestion
# ln -s -f ../../tools/pre-commit-podman .git/hooks/pre-commit
```

biddisco (changes requested):

tools/pre-commit-podman
msimberg: Thanks, good catch.

PR #1332, opened 2024-11-15
Avoid stack overflow in `task_size` bulk test

msimberg:
stdexec in debug mode is particularly prone to triggering a stack overflow, but pika's std::execution implementation can trigger it as well. Compiling with optimizations also simply delays the stack overflow. This should fix the frequent failures on the performance tests in CI.


tests/performance/local/task_size.cpp
```diff
@@ -190,6 +190,9 @@ double do_work_bulk(
     for (std::uint64_t i = 0; i < tasks_per_thread; ++i)
     {
         sender = std::move(sender) | ex::continues_on(sched) | ex::bulk(num_threads, work);
+        // To avoid stack overflow when connecting, starting, or destroying the operation state,
+        // eagerly start the chain of work periodically using ensure_started.
+        if (i % 100 == 0) { sender = ex::ensure_started(std::move(sender)); }
```
biddisco, line 195: Is the number 100 pulled out of thin air, or is there some calculation behind it? if its an estimate of how much space each task needs, divided into the default stack size (or something of that kind), it might be worth adding a comment to explain it - if the default stack size were to change, this number might need to be re-evauated(?)

msimberg: Pulled entirely out of thin air... Mostly this: "start eagerly as often as we can without introducing significant overheads, and once every 100 iterations seemed reasonable".

I'll definitely add a note about this in the comment. I'm not sure if/how we can reasonably compute a good value here. It depends on stdexec/no-stdexec and optimizations. In this case it had a stack overflow in connect, but it could just as well in the future be in the operation state destructor or in start.

msimberg: @biddisco I updated the message in https://github.com/pika-org/pika/compare/f845306cee31e266067bf114502e5ef8f8505ef9..c53420116a07dd64ff1ff0dc3af06fc9c36a768c. Does that explain it well enough to you?

msimberg: I've added this to the merge queue, but please feel free to suggest further changes even after merging if you'd like.

PR #1335, opened 2024-11-18
Use ctest's built-in junit support

msimberg:
The xsl file previously used for conversion has been removed from the CDash repository. Instead of copying the file over to this repository, we can use ctest's built-in functionality to output a junit file. This functionality exists since 3.21 (see https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-output-junit).


msimberg:
Merging manually as this only affects CircleCI and jobs are failing there without this change.

PR #1339, opened 2024-11-19
Add `RelWithDebInfo` mode sanitizer CI workflows

msimberg:


msimberg:
I've updated this to use `RelWithDebInfo`. `Release` mode with tsan still has issues and it's not clear if it's a false positive or not. Will investigate after #1326.

PR #1346, opened 2024-11-20
Minor `transform_mpi` refactoring

msimberg:
This PR is a proposal for some small refactorings of how the `transform_mpi` adaptor is composed of `dispatch_mpi` and `trigger_mpi`.

There are two goals with the changes:
- handle the "early polling" in one place and one place only: the `trigger_mpi` receiver
- remove nesting of `let_value` and extra type-erasure with `unique_any_sender`

The early check for `MPI_REQUEST_NULL` was done in at least two places, and the early polling was done in `dispatch_mpi`. I've moved the early poll to `trigger_mpi` where I think it fits semantically.


biddisco (approved):
This is good. I had been meaning to do something about the extra unique sender, but somehow blanked it out of my mind. This is much cleaner. thanks.

The polling and REQUEST_NULL return was mosdtly left for debugging but moving it into the trigger makes sense and simplifies the pipeline checks

PR #1354, opened 2024-11-25
Store receiver in `any_operation_state` to avoid dynamic allocation for receiver

### LINKED ISSUE #845: Update `any_receiver` to not require heap allocation
`any_receiver` should be possible to convert into more or less a reference. The non-type-erased receiver should be possible to store in the operation state, with `any_receiver` simply pointing to the operation state. stdexec seems to do something in this direction, and could save a significant number of heap allocations in DLA-Future.

msimberg:
Complement to #1281. Fixes #845.

Adds more indirection so that the concrete receiver type doesn't have to be type-erased. It can instead be stored directly in the operation state type, and then a reference/pointer to a base class is passed to connect instead. This does not avoid the indirection/virtual function dispatch, but only requires passing a pointer-sized receiver to connect.

### DIFF
--- a/libs/pika/execution_base/include/pika/execution_base/any_sender.hpp
+++ b/libs/pika/execution_base/include/pika/execution_base/any_sender.hpp
@@ -12,12 +12,14 @@
 #include <pika/execution_base/operation_state.hpp>
 #include <pika/execution_base/receiver.hpp>
 #include <pika/execution_base/sender.hpp>
+#include <pika/type_support/detail/with_result_of.hpp>
 #include <pika/type_support/pack.hpp>
 
 #include <cstddef>
 #include <cstring>
 #include <exception>
 #include <memory>
+#include <optional>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


msimberg:
@albestro, @rasolca, @RMeli I'm not expecting a review from you, but requested one just in case you're interested in looking. I feel like I should be able to remove one further level of indirection in the receiver, but have not been able to figure out how (if it's possible). I'm happy with the state of this already, but in case you spot something that looks suspicious or could be improved, do let me know.

libs/pika/execution_base/include/pika/execution_base/any_sender.hpp
```diff
@@ -683,39 +673,35 @@ namespace pika::execution::experimental::detail {
...
-        // before the any_(unique_)sender is destroyed. This would be
+        // This helper only exists to make it possible to use any_(unique_)sender in global
+        // variables or in general static that may be created before main. When used as a base for
+        // any_(unique_)_sender, this ensures that the empty vtables for any_operation_state are
+        // created as the first thing when creating an any_(unique_)sender. The empty vtables for
+        // any_operation_state may otherwise be created much later (when the sender is connected and
+        // started), and thus destroyed before the any_(unique_)sender is destroyed. This would be
         // problematic since the any_(unique_)sender can hold previously created
-        // any_receivers and any_operation_states indirectly.
+        // any_operation_states indirectly.
```
albestro, line 698: `(unique_)any_sender` instead of `any_(unique_)sender`?

PR #1356, opened 2024-11-28
Add documentation for `async_rw_mutex`

msimberg:
Part of #49.

### DIFF
--- a/examples/documentation/async_rw_mutex_documentation.cpp
+++ b/examples/documentation/async_rw_mutex_documentation.cpp
@@ -1,0 +1,75 @@
+//  Copyright (c) 2024 ETH Zurich
+//
+//  SPDX-License-Identifier: BSL-1.0
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <pika/async_rw_mutex.hpp>
+#include <pika/execution.hpp>
+#include <pika/init.hpp>
+
+#include <fmt/printf.h>
+
+#include <type_traits>
+#include <utility>
+
+int main(int argc, char* argv[])
+{
+    namespace ex = pika::execution::experimental;
+    namespace tt = pika::this_thread::experimental;
+
+    pika::start(argc, argv);
+    ex::thread_pool_scheduler sched{};
+
+    {
+        ex::async_rw_mutex<int> m{0};
+
+        // This read-write access is guaranteed to not run concurrently with any
+        // other accesses. It will also run first since we requested the sender
+        // first from the mutex.
+        auto rw_access1 =
+            m.readwrite() | ex::continues_on(sched) | ex::then([](auto w) {
+                w.get() = 13;
+                fmt::print("updated value to {}\n", w.get());
+            });
+
+        // These read-only accesses can only read the value, but they can run
+        // concurrently. They'll see the write from the access above.
+        auto ro_access1 =
+            m.read() | ex::continues_on(sched) | ex::then([](auto w) {
+                static_assert(std::is_const_v<
+                    std::remove_reference_t<decltype(w.get())>>);
+                fmt::print("value is now {}\n", w.get());
+            });
+        auto ro_access2 =
+            m.read() | ex::continues_on(sched) | ex::then([](auto w) {
+                static_assert(std::is_const_v<
+                    std::remove_reference_t<decltype(w.get())>>);
+                fmt::print("value is {} here as well\n", w.get());
+            });
+        auto ro_access3 =
+            m.read() | ex::continues_on(sched) | ex::then([](auto w) {
+                static_assert(std::is_const_v<
+                    std::remove_reference_t<decltype(w.get())>>);
+                fmt::print("and {} here too\n", w.get());
+            });
+
+        // This read-write access will run once all the above read-only accesses
+        // are done.
+        auto rw_access2 =
+            m.readwrite() | ex::continues_on(sched) | ex::then([](auto w) {
+                w.get() = 42;
+                fmt::print("value is {} at the end\n", w.get());
+            });
+
+        // Start and wait for all the work to finish.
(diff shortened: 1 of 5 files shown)


examples/documentation/async_rw_mutex_documentation.cpp
RMeli, line 1: The code is very clear and maybe this is overkill, but since I've seen it in DLAF code base I'm wondering if a graphical representation like the following might be useful for newcomers?
```
// rw1 - - - ro1 - - - rw2  
//       \ - ro2 - /
//         \ ro3 /
```

libs/pika/synchronization/include/pika/synchronization/async_rw_mutex.hpp
```diff
@@ -258,31 +276,36 @@ namespace pika::execution::experimental {
...
+    /// The order in which senders signal a receiver is determined by the order in which the senders
+    /// are retrieved from the mutex. Connecting and starting the senders is thread-safe.
     ///
-    /// The order in which senders call set_value is determined by the order in
-    /// which the senders are retrieved from the mutex. Connecting and starting
-    /// the senders is thread-safe.
+    /// The mutex is move-only.
     ///
-    /// Retrieving senders from the mutex is not thread-safe.
+    /// \warning Because access to the wrapped value is granted in the order that it  is requested
```
RMeli, line 297:
```suggestion
    /// \warning Because access to the wrapped value is granted in the order that it is requested
```

RMeli (approved):

examples/documentation/async_rw_mutex_documentation.cpp
msimberg: Yeah, adding that might be worth adding!

msimberg: What about https://github.com/pika-org/pika/pull/1356/commits/66173a74851b038d1db10b87c590d07b23947120? I've added some extra text there explaining the dependencies are implied by the way the senders are accessed from the mutex, rather than explicit dependencies between senders. I find this an important distinction to understand that senders don't one-to-one map to e.g. tasks, but I don't know if it's a useful distinction for users?

RMeli: I think it's useful, and the extra context is helpful.

examples/documentation/async_rw_mutex_documentation.cpp
```diff
@@ -0,0 +1,85 @@
...
+        auto rw_access2 =
+            m.readwrite() | ex::continues_on(sched) | ex::then([](auto w) {
+                w.get() = 42;
+                fmt::print("value is {} at the end\n", w.get());
+            });
+
+        // Start and wait for all the work to finish.
+        tt::sync_wait(ex::when_all(std::move(rw_access1), std::move(ro_access1),
+            std::move(ro_access2), std::move(ro_access3),
+            std::move(rw_access2)));
```
albestro, line 78: Question: in principle, given that DAG, can we `sync_wait` just the last readwrite access?

My doubt is this one. They do not depend on each other as sender, but the mutex implicitly creates a dependency. This "difference" (if it is a difference, or is it just conceptual?) requires that we start each sender independently, or we can just `sync_wait` the last one and all the chain is started as if they were actually dependent?

albestro (approved):
Just a curiosity, but for the rest thanks for the nice documentation!

examples/documentation/async_rw_mutex_documentation.cpp
msimberg: > Question: in principle, given that DAG, can we `sync_wait` just the last readwrite access?

No.
 
> My doubt is this one. They do not depend on each other as sender, but the mutex implicitly creates a dependency. This "difference" (if it is a difference, or is it just conceptual?) requires that we start each sender independently, or we can just `sync_wait` the last one and all the chain is started as if they were actually dependent?

If you `sync_wait` only the last one you haven't said anything about what you want to do with the other accesses (senders). They are still alive, waiting to be connected with receivers, and as such you can't access the last one yet: the senders give access in the order that they are accessed from the mutex. I think you could destroy the previous senders without starting them and then access only the last one with `sync_wait` and that'd work, but it's not an intended use case (I'll have to check what exactly happens in that corner case).

It's an excellent question and is close in spirit to what @RMeli raises in https://github.com/pika-org/pika/pull/1356#discussion_r1863594415 and what I tried to explain a bit in https://github.com/pika-org/pika/pull/1356/commits/66173a74851b038d1db10b87c590d07b23947120 (the part about "implicit" dependencies, and elsewhere in the documentation "senders give access to the wrapped value in the order they are accessed from the mutex").

albestro: Thanks for the explanation! I didn't think enough before commenting, but it makes perfect sense.

The comment you are referring to is the one where I was reasoning on: probably what led me out of track is that even if they are not **EXPLICITLY** dependent, they are **IMPLICITLY** dependent; so they are **DEPENDENT**?! But I see the struggle in wording it.

Not sure it helps, but instead of IMPLICITLY, maybe **INDIRECTLY**? not sure it changes much, but I'm looking for a wording that makes clear that despite at a "logical" level a dependency relation is created between them, they are still completely independent and they have to be started, it goes without saying, independently at the runtime level.

Not sure it gives you enough insight on my doubt. I think the doc is already very good, I wouldn't touch it much/at all.

albestro: The documentation you pointed out was part of my draft comment 😅 That is absolutely a clarifier for my original doubt.

About information spread in two different places, IMHO it might be the case, but I see why doc is like it is now.

I would support having this message also in the example here, because it raise awareness about the problem and it also helps clarifying why it is needed to start all of them (linked to the sentence below the graph with dependencies).

msimberg: What do you think about https://github.com/pika-org/pika/pull/1356/commits/bc2bae4826d020fc36a179bb01d3a8acb754eed7 @albestro?

albestro: That's perfect! 🔝

PR #1365, opened 2024-12-05
Try reenabling `detect_stack_use_after_return` again on address sanitizer CI workflow

msimberg:
The comments in https://github.com/google/sanitizers/issues/1760 seem to suggest that excluding the `__sanitizer*` functions from asan itself, may help avoiding `WARNING: ASan is ignoring requested __asan_handle_no_return` (but there's more, and it may be that the other changes in that issue are actually required to avoid the warning).


msimberg:
Nope, still problematic, at least with only adding `no_sanitize("address")`. Needs more investigation.

PR #1366, opened 2024-12-06
Upgrade clang + CUDA CI pipeline to use CUDA 12

msimberg:
This should allow using C++20 again for both the spec and the build of pika. This also allows removing the `+allow-unsupported-compilers` variant.


msimberg:
Well, now clang is old enough for CUDA to support, but CUDA is too new for clang to support. This may be causing errors. I'll come back to this once other CI PRs have settled a bit.

PR #1368, opened 2024-12-10
Fix the documentation for `set_value`

aurianer:
The documentation is outdated after `set_value` was changed to use member functions. Remove the whole documentation until we can document it properly.


msimberg (changes requested):
Well spotted, thanks! Would you mind removing the whole docstring actually? The second point is wrong, and I'm not sure yet if/how we want to expose these CPOs in the docs. In the meantime I think it's more misleading than useful to keep this docstring. You can do the same for `set_error` etc.

msimberg (approved):

Issue #1204, opened 2024-07-24
Convert `std::execution` functionality to use member functions instead of `tag_invoke`


msimberg:
Later versions of stdexec support both `tag_invoke` and member functions, but the standard will use member functions only. We should try to migrate to use member functions only where it makes sense with reasonable backwards compatibility to the old C++17 implementation in pika.

The following CPOs and customizations can be translated quite simply:
- [x] `set_value_t` (#1295)
- [ ] `set_error_t` (#1369)
- [ ] `set_stopped_t`
- [x] `start_t` (#1392)
- [x] `connect_t` (#1393)
- [ ] `schedule_t`
- [x] `get_env_t` (#1394)

Sender adaptor customization we currently still do with `tag_invoke`, but this is not an as straightforward translation. E.g. `tag_invoke(bulk_t, thread_pool_scheduler, ...)` would need to be translated to `transform_sender(thread_pool_scheduler_domain, bulk_sender)`. We don't have domains implemented for schedulers yet, so we'd need that before tackling this.

PR #1369, opened 2024-12-10
Make `set_error` use member function instead of `tag_invoke`

aurianer:
Part of #1204.

The `non_receiver_6` test is removed from `basic_receiver_test` because stdexec (and now pika) `static_assert` that `set_error` is `noexcept` rather than SFINAE-ing it out. This means we can't detect non-`noexcept` `set_error` member functions as not receivers at compile time. We could change this, but for now we're sticking with the behaviour of stdexec.


cmake/pika_add_config_test.cmake
```diff
@@ -447,7 +449,7 @@ function(pika_check_for_cxx23_static_call_operator)
   pika_add_config_test(
     PIKA_WITH_CXX23_STATIC_CALL_OPERATOR
     SOURCE cmake/tests/cxx23_static_call_operator.cpp
-    FILE ${ARGN}
+    FILE ${ARGN} CXXFLAGS -pedantic-errors
```
msimberg, line 452: Do you know what the failure mode is if this flag isn't supported? Does the test fail or does configuration fail? I'm thinking e.g. about `nvc++` which may not support the flag, but if it does it's perhaps not such a big concern.

msimberg: Reverted in https://github.com/pika-org/pika/pull/1369/commits/39b21c0cf970d74a1e914c60701c2adabbf7db95.

PR #1370, opened 2024-12-10
Remove unused `extra_cxxflags` cmake variable

aurianer:


msimberg (approved):
Same review as on #1371: https://github.com/pika-org/pika/pull/1371#pullrequestreview-2494784297.

PR #1371, opened 2024-12-10
Remove unused cmake variable

aurianer:
I thought I had already remove it in #1259 but looks like it's back


msimberg (approved):
Thanks. You did remove an instance of this in the past, but I managed to reintroduce the same issue in another place with https://github.com/pika-org/pika/pull/1249. #1249 and #1259 were merged around the same time, probably the reason why we didn't notice that one PR removed it and another added it in another place.

PR #1377, opened 2024-12-18
Add documentation section about command line options

RMeli:
Add very bare bone section about CLI options and environment variables in `pika`. The main goal right now is to document `PIKA_COMMANDLINE_OPTIONS`.


docs/usage.rst
```diff
@@ -190,6 +190,24 @@ Since version ``0.20.0``, the ``pika-bind`` helper script is bundled with pika.
...
+====================
+
+pika's behaviour can be controlled with command line options, or environment variables. Not all
+command line options are exposed as environment variables. When both are present, command line
+options take precedence over environment variables.
+
+If a command line option is not exposed as an environment variable, but it is necessary to set it,
+it is possible to use the ``PIKA_COMMANDLINE_OPTIONS`` environment variable.
+
+For example, the following disable thread binding and explicitly sets the number of threads:
```
msimberg, line 205:
```suggestion
For example, the following disables thread binding and explicitly sets the number of threads:
```

msimberg (changes requested):
Only a minor nit, otherwise this looks good to me. Thank you @RMeli!

docs/usage.rst
msimberg: Or... does "the following" refer to "the following option*s*"? I think probably it still makes sense if "the following" is "the following export-thing" (which is singular), but open to either interpretation.

RMeli: I meant the following example/snippet, so the "s" is indeed needed, thanks.

msimberg (approved):

Issue #1125, opened 2024-05-10
Do less dynamic memory allocation in `async_rw_mutex`


msimberg:
`async_rw_mutex` currently allocates many shared states through `shared_ptr`s. Look into reducing the number of allocations.

PR #1379, opened 2024-12-19
Refactor `async_rw_mutex`

msimberg:
This is an attempt at slightly simplifying and optimizing the internals of `async_rw_mutex`. This avoids the needs for a lock to keep track of continuations and instead triggers continuations through operation states. The continuations are linked to each other through an intrusive linked list of operation states. These changes also avoid the need to have a weak shared pointer between shared states.

Overall I'm hoping that the removal of extra reference counting and locks will slightly improve performance, but fundamentally the structure is still the same, requiring the same amount of dynamic allocations (cf. #1125; this PR does _not_ address that) as before (in fact, one more allocation for the value stored by the mutex). So the impact may be minimal in terms of performance. However, I'm also making these changes to make the dependency triggering a bit more understandable.

This also does some minor refactorings, such as:
- make move constructors/assignment operators noexcept
- make readwrite senders move-only


msimberg:
I think I'm done with the refactorings for this time at least. I'd appreciate someone having at least a high level look at this.

I've added some more text and diagrams on how the implementation works. It could probably still be expanded, but I'm hoping it provides at least a better explanation than what was there before. Note that the implementation has changed sufficiently that the new implementation description does not match what was done before.

msimberg:
It looks like the performance regression is a result of the linked operation states being accessed in reverse order compared to before. This was only affecting the GPU backend in DLA-Future, where GPU work is scheduled inline. I've pushed a commit which restores the order of calling continuations and I'm rerunning benchmarks. We shouldn't make this a guarantee, but I'm keeping the order unchanged in this PR to not upset DLA-Future performance for now. We can see if it's possible to relax the order in the future without affecting performance.

msimberg:
Reversing the order of continuations now brings the performance very close to what it was with the old implementation. Some algorithms in DLA-Future still show a tiny performance improvement, but nothing dramatic.

msimberg:
This is again ready for review. Changes since last time are outlined in the previous comments.

msimberg:
I've updated this to fix some race conditions in the (`async_rw_mutex`) intrusive queue handling and added a test that quite reliably triggers a data race with tsan. I've also restored the "unintuitive" LIFO behaviour for continuations as DLA-Future has seen a change in the meantime that avoids the need to do the reversing of the order of continuations to be FIFO again. https://github.com/eth-cscs/DLA-Future/pull/1292 creates a new task for GPU work as well, which means that the order isn't really significant anymore.

I reran the benchmarks with DLA-Future 0.8.0 which does _not_ have the above PR merged yet, and the performance regression reported a few comments up is still there (dlaf_master means DLA-Future 0.8.0):
chol_strong_time_20480_1024

However, with DLA-Future master which does have the PR merged already, there is practically no difference between pika main and this PR with FIFO or LIFO order:
chol_strong_time_20480

Note that this performance difference does not show up at all on e.g. the full eigensolver with any of the version combinations tested, so I'm not worried about this particular "bad" combination of versions having a significant impact on users.

PR #1380, opened 2024-12-19
Expose more configuration options as environment variables

msimberg:
This exposes:
- `PIKA_IGNORE_PROCESS_MASK`
- `PIKA_THREADS`
- `PIKA_SCHEDULER`
- `PIKA_BIND
as environment variables that can be used to set a default for the corresponding configuration options. `PIKA_SCHEDULER` corresponds to `pika.scheduler` and `--pika:queueing`. I feel like the latter should be renamed to `--pika:scheduler`.

All of the above variables should be documented as well.

The following are also exposed in this PR as environment variables, but I don't intend to document them as I feel like they could be removed any time. They're exposed for development purposes only.
- `PIKA_AFFINITY`
- `PIKA_PU_STEP`
- `PIKA_PU_OFFSET`
- `PIKA_NUMA_SENSITIVE`


libs/pika/runtime/tests/unit/CMakeLists.txt
```diff
@@ -23,22 +23,153 @@ foreach(test ${tests})
...
+pika_add_configuration_precedence_test(
+  configuration_precedence 9 -3 "--pika:print-bind" "${two_threads_expected_output}"
+  "PIKA_THREADS=2"
+)
+
+# Set number of threads through environment variable and command line option.
+pika_add_configuration_precedence_test(
+  configuration_precedence 10 -3 "--pika:print-bind --pika:threads=2"
+  "${two_threads_expected_output}" "PIKA_THREADS=1"
+)
```
msimberg, line 175: This is not particularly pretty. Nicer ways of doing this or is this ok?

Are there (important) combinations of options missing that are not tested?

cmake/pika_add_test.cmake
```diff
@@ -54,10 +54,16 @@ function(pika_add_test category name)
     elseif(${name}_THREADS EQUAL -2)
       set(${name}_THREADS "cores")
       set(run_serial TRUE)
+    elseif(${name}_THREADS EQUAL -3)
+      set(${name}_THREADS "default")
+      set(run_serial TRUE)
     endif()
   endif()
```
msimberg, line 61: Maybe these magic numbers could be replaced by the actual names of the command line options?

libs/pika/runtime/tests/unit/CMakeLists.txt
msimberg: I've added tests for ignoring process masks.

cmake/pika_add_test.cmake
msimberg: Not making any changes in this PR.

msimberg:
The test failures come from a combination of #1239 and #1380 being merged without being tested together... Merging this manually, and fixing the new test failures separately.

PR #1383, opened 2024-12-20
Print variables used to build pika in CSCS CI

aurianer:


.gitlab/docker/Dockerfile.spack_build
```diff
@@ -23,6 +23,7 @@ RUN spack -e pika_ci config add "config:flags:keep_werror:all" && \
     if [[ $COMPILER =~ clang ]]; then \
         export PATH="$(spack location --install-dir llvm)/bin:$PATH"; \
     fi && \
+    echo "build_type: ${BUILD_TYPE}  cmake_flags: ${CMAKE_COMMON_FLAGS} ${CMAKE_FLAGS}" && \
     spack -e pika_ci build-env $spack_spec -- \
     bash -c " \
     cmake -B${BUILD_DIR} ${SOURCE_DIR} -DCMAKE_BUILD_TYPE=$BUILD_TYPE $CMAKE_COMMON_FLAGS $CMAKE_FLAGS && \
```
msimberg, line 29: What do you think about:
```suggestion
    spack -e pika_ci build-env $spack_spec -- \
    bash -c " \
    set -x;
    cmake -B${BUILD_DIR} ${SOURCE_DIR} -DCMAKE_BUILD_TYPE=$BUILD_TYPE $CMAKE_COMMON_FLAGS $CMAKE_FLAGS && \
```
? This would give us the actual command that we use to build. This would include `BUILD_DIR`, `SOURCE_DIR`, and `NUM_PROCS` as well. I like your approach as well, but in that case I would print the variables on separate lines and also print these three extra variables.

PR #1412, opened 2025-02-21
Add basic documentation on debugging with pika

msimberg:
Based on #1411.

Naming of options to do based on what will be implemented in #1411.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -302,6 +302,112 @@ following:
 - ``%q``: The parent task id and description.
 - ``%k``: The current task id and description.
 
+.. _debugging:
+
+Debugging
+=========
+
+Writing task based applications can be tricky, and debugging them even more so. This section
+describes a few options and tools that can be helpful when debugging applications using pika.
+
+.. _segfaults_and_stack_overflows:
+
+Segmentation faults and stack overflows
+---------------------------------------
+
+Due to the small default stack sizes of user level threads a common problem is stack overflows. When
+using the ``mmap``-based stack allocation (default on platforms that support it) pika provides a
+configuration option to enable guard pages at the end of a stack. When enabled, a protected page
+will be allocated such that if one attempts to read or write within a page of the end of the stack,
+a segmentation fault will be triggered. The option can be enabled by exporting
+``PIKA_USE_GUARD_PAGES=1``. The option is disabled by default.
+
+Additionally, pika can install signal handlers that print information about failures, such as
+backtraces. These handlers will handle the most common events, such as interrupts, segmentation
+faults, illegal instructions etc. and they can be enabled with the environment variable
+``PIKA_INSTALL_SIGNAL_HANDLERS=1``. The verbosity can be controlled with
+``PIKA_EXCEPTION_VERBOSITY`` (this also controls how much information pika exceptions capture and
+print). The default value of ``1`` will print a backtrace. ``2`` or higher will print additional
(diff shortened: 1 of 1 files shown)


msimberg:
@albestro, @rasolca, @RMeli as usual I appreciate any feedback you might have on the docs and options here.

docs/usage.rst
```diff
@@ -302,6 +302,116 @@ following:
...
+
+   Many MPI implementations also install signal handlers or have options for enabling them.
+   ``libSegFault.so`` (part of ``glibc-tools``, more info in `this blog post
+   <https://www.marcusfolkesson.se/blog/libsegfault/>`_) also provides a way to install a signal
+   handler. These can be useful alternatives to the signal handlers provided by pika. Depending on
+   the type of issue you are debugging, different signal handlers can be more or less helpful as
+   they print slightly different information.
+
+   pika does not install any signal handlers by default. They have to be enabled explicitly using
+   the options described below.
```
RMeli, line 335: This note feels misplaced: should it go after the next session? Mainly asking because of the "also install", but you only says below that pika can install signal handlers.

docs/usage.rst
```diff
@@ -302,6 +302,116 @@ following:
...
+the chances of false positives make sure to build pika with the CMake option
+``PIKA_WITH_SANITIZERS=ON``. This does not enable any sanitizers directly, but disables certain
+functionalities internally to work better with sanitizers. To actually enable sanitizers, enable
+them explicitly by setting ``-fsanitize=`` flags as for any other CMake project. It's highly
+recommended to use ``-fno-omit-frame-pointer`` with sanitizers.
+
+There are known issues that may prevent you from using sanitizers with pika. Under the ``tools``
+subdirectory of the pika repository you can find the most recent suppression files that are used for
+CI runs with sanitizers. Similarly, under ``.github/workflows`` you can find the most recent build
+configurations (including sanitizer options) that with sanitizers, along with blacklists of tests
```
RMeli, line 412:
```suggestion
configurations (including sanitizer options) that work with sanitizers, along with blacklists of tests
```
Use or similar?

docs/usage.rst
```diff
@@ -302,6 +302,116 @@ following:
...
+that are currently known to not pass with sanitizers.
```
RMeli, line 413:
```suggestion
that are currently known to fail with sanitizers.
```
?

msimberg:
Thank you @RMeli for the comments, all very useful! Do you have any opinions on the names of the configuration options/environment variables. The "general signal handlers" option is `PIKA_INSTALL_SIGNAL_HANDLERS`, and in #1411 the SIGSEGV one is called `PIKA_STACKOVERFLOW_DETECTION` (simply because the CMake option used to be called `PIKA_WITH_STACKOVERFLOW_DETECTION`).

After updating #1411 a bit more, I'm leaning towards just setting the special SIGSEGV handler as part of `PIKA_INSTALL_SIGNAL_HANDLERS`, then there would only be one option. The downside of this is that the SIGSEGV handler prints very minimal information (basically only "segmentation fault at address 0xabc, try changing stack sizes").

docs/usage.rst
msimberg: Yeah, I definitely put it too early. I've moved it down a bit in https://github.com/pika-org/pika/pull/1412/commits/33e656a8e0f6196239f37ea386a06cb553e4aaab.

PR #1417, opened 2025-03-14
Add MPI include dir before performing mpix continuations test

biddisco:
Fixes a bad check for MPIX continuations when using openmpi without compiler wrappers


msimberg (approved):
Good catch!

PR #1440, opened 2025-06-12
Use separate package repo in CI

msimberg:
With spack 1.0, the builtin repo will be pulled in separately from the spack tool. Set up CI to explicitly specify which commit to use for the package repo.


.gitlab/pipelines_on_push.yml
```diff
@@ -9,4 +9,6 @@ include:
   - local: '.gitlab/includes/clang17_cuda12_pipeline.yml'
   - local: '.gitlab/includes/gcc12_hip6_pipeline.yml'
   - local: '.gitlab/includes/gcc12_cuda12_pipeline.yml'
+  # TODO: remove
+  - local: '.gitlab/includes/nvhpc24_9_pipeline.yml'
```
msimberg, line 13: To do.

PR #1443, opened 2025-07-11
Set `amdgpu_target` value for all packages in HIP CI configurations

msimberg:


msimberg:
This seems to not concretize. This is not so important so closing for the moment, maybe investigating some other time.

PR #1447, opened 2025-08-04
Make GCC 15 CI pipeline use C++26

msimberg:
Update GCC 14 pipeline to use C++23 so that at least one GCC configuration uses C++23.


msimberg:
Not updating stdexec commit because of #1448. The `task_size --method=bulk` benchmark runs very slowly because it doesn't seem to be parallelized.

msimberg:
Putting this PR on hold. It'd be nice to have this build use stdexec as that's what we're targeting as the default option. However, an older version of stdexec (25.03 tag) doesn't build with GCC 15. A newer version of stdexec (main as of today) has removed algorithm customizations through `tag_invoke`. Waiting until we update the algorithm customizations.

PR #1451, opened 2025-09-24
Require C++20

### LINKED ISSUE #292: Require C++20
This is required to unconditionally use the P2300 reference implementation. nvcc/nvc++ is currently the only compiler holding us back.

msimberg:
Fixes #292.


msimberg:
Boost supports C++20 only from 1.77, so I'm going to bump the minimum requirement for Boost. Boost 1.77 was released in 2021 (https://www.boost.org/releases/1.77.0/) so I think it's fair (and relatively conservative) to require 1.77. The previous minimum, 1.71, was released in 2019 (https://www.boost.org/releases/1.71.0/).

PR #1466, opened 2026-03-02
Fix deprecation warning about empty_env

biddisco:
I get a lot of empty_env warnings on my builds. This PR fixes them, but I didn't check if there should be a version  requirement etc


msimberg:
I'll close this as #1467 will take care of the same warnings.

PR #1475, opened 2026-03-26
Use pika's `execute` and `split` implementation even when stdexec enabled

### LINKED ISSUE #1470: Use our own implementation of execute and split even when stdexec is enabled

aurianer:
Fixes #1471
Fixes #1470


cmake/tests/stdexec_transform_completion_signatures_of.cpp
```diff
@@ -0,0 +1,15 @@
...
+//  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <stdexec/execution.hpp>
+
+int main()
+{
+    // Newer versions of stdexec deprecate transform_completion_signatures_of and provide
+    // __transform_completion_signatures_of_t as a non-deprecated replacement.
+    using type =
+        stdexec::__transform_completion_signatures_of_t<decltype(stdexec::just()), stdexec::env<>>;
```
msimberg: Similar to the comment about `__execute_t` I made previously in `stdexec_forward.hpp`, I don't think we should be using the internal __* names to work around a name being deprecated. This is allowed to change internally in stdexec any time.

Either we directly use the deprecated name and silence warnings about our uses of it, we create our own version of `transform_completion_signatures_of_t`, or we just use the replacement in `exec`.

libs/pika/execution/include/pika/execution/algorithms/execute.hpp
```diff
@@ -8,17 +8,28 @@
...
+    inline constexpr struct execute_t
+    {
+        template <typename Scheduler, typename F>
+        constexpr PIKA_FORCEINLINE auto operator()(Scheduler&& scheduler, F&& f) const
+        {
+            return start_detached(
+                then(schedule(std::forward<Scheduler>(scheduler)), std::forward<F>(f)));
+        }
+    } execute{};
+#else
```
msimberg: Why do we need a different implementation for `PIKA_HAVE_STDEXEC`? I wouldn't be against using this non-customizable version as the only version, but as far as I can tell we don't need both implementations, right?

libs/pika/execution/include/pika/execution/algorithms/split.hpp
```diff
@@ -304,7 +334,8 @@ namespace pika::split_detail {
             if (!start_called.exchange(true))
             {
                 PIKA_ASSERT(os.has_value());
-                pika::execution::experimental::start(*os);
+                pika::execution::experimental::start(
+                    *os);    // NOLINT(bugprone-unchecked-optional-access)
```
msimberg: `// NOLINTNEXTLINE` is typically cleaner for cases like this, you can put it on the line above.

libs/pika/execution/include/pika/execution/algorithms/split.hpp
```diff
@@ -437,6 +492,39 @@ namespace pika::split_detail {
...
+            friend auto operator|(Sender&& sender, closure_type)
+            {
+                return split_t{}(std::forward<Sender>(sender));
+            }
+        };
+
+    public:
+        constexpr auto operator()() const noexcept { return closure_type{}; }
+    } split{};
+#else
```
msimberg: Similar to `execute_t`, why does this need a different implementation when `PIKA_HAVE_STDEXEC`?

libs/pika/execution/include/pika/execution/algorithms/transfer_just.hpp
```diff
@@ -8,16 +8,28 @@
...
+        "transfer_just will be removed in the future, use transfer and just separately instead")
+    inline constexpr struct transfer_just_t
+    {
+        template <typename Scheduler, typename... Ts>
+        constexpr PIKA_FORCEINLINE auto operator()(Scheduler&& scheduler, Ts&&... ts) const
+        {
+            return continues_on(just(std::forward<Ts>(ts)...), std::forward<Scheduler>(scheduler));
+        }
+    } transfer_just{};
+#else
```
msimberg: Same here... why separate? Also this doesn't seem to have anything to do with `execute` and `split`, or am I missing some connection?

cmake/tests/stdexec_transform_completion_signatures_of.cpp
aurianer: yes sorry I missed the replacement in exec thanks! also changed it to use the official namespace `experimental::execution`

libs/pika/execution/include/pika/execution/algorithms/execute.hpp
aurianer: we don't, you are right it's cleaner this way!

libs/pika/execution/include/pika/execution/algorithms/execute.hpp
```diff
@@ -8,27 +8,32 @@
...
-# include <utility>
+#include <utility>
 
 namespace pika::execution::experimental {
-    inline constexpr struct execute_t final : pika::functional::detail::tag_fallback<execute_t>
+    // pika uses its own execute implementation because stdexec's execute is deprecated.
+    inline constexpr struct execute_t
+#if !defined(PIKA_HAVE_STDEXEC)
+        final : pika::functional::detail::tag_fallback<execute_t>
+#endif
```
msimberg, line 25: Sorry, I still don't understand why we define the CPO differently with and without stdexec? I may very well be missing something very basic, but fundamentally there's nothing stopping us from using `tag_fallback` define our CPOs even when stdexec is enabled.

cmake/tests/stdexec_transform_completion_signatures_of.cpp
```diff
@@ -0,0 +1,21 @@
+//  Copyright (c) 2024 ETH Zurich
```
msimberg, line 1: I think this file is confusingly named: it checks if `::experimental::execution::transform_completion_signatures` is available, not `stdexec::transform_completion_signatures_of` as the name would imply (to me at least). Probably the test file and the CMake feature detection variable should just drop the `_of` suffix?

libs/pika/execution/include/pika/execution/algorithms/execute.hpp
aurianer: I thought member function was needed here but I was actually wrong, all the body is calling pika's own functionality and it should work like this, thank you!

cmake/tests/stdexec_transform_completion_signatures_of.cpp
aurianer: yes I changed what was tested without changing the name, thanks for spotting that!

libs/pika/execution_base/include/pika/execution_base/stdexec_forward.hpp
```diff
@@ -42,5 +26,47 @@ namespace pika::execution::experimental {
...
+        // template-template API while forwarding to the new function-object API.
+        template <template <class...> class F>
+        struct template_alias_to_fn
+        {
+            template <class... Args>
+            consteval auto operator()() const noexcept
+            {
+                return F<Args...>{};
+            }
+        };
```
aurianer, line 58: @msimberg I added this since using the public `transform_completion_signatures` doesn't accept template templates. I've looked what is done in stdexec and they use the internal meta function with __ which accepts template templates but since we said we don't want to use the internals, this is what I added, what do you think?

msimberg (approved):
Thanks for the updates @aurianer, looks great!

msimberg:
Codacy is wrong, ignoring that.

msimberg:
I think most CI failures were not related to changes here. The clang13 failure looks real: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/479009878135925/5304355110917878/-/jobs/14055830094#L329. Would you mind having a look? I'm definitely not opposed to removing the clang 13 configuration. It's already more than 4 years old, so if getting this running on clang 13 is more than a trivial amount of work I think we can just drop it (but note the change in docs/usage.rst). We should likely add some newer clang versions to CI, but that's not for this PR...

aurianer:
The problem was not with clang13 but with the older stdexec commit vs newer one. Before we could connect a non const& completion_signatures to a const& value type. The new commit I pinned is actually checking this more strictly, so I've added const ref completion signatures too as a fix. In stdexec, value_types and error_types are derived from completion_signatures so there wouldn't be this mismatch but with pika's implementation we have the completion signatures and value_types mismatch, I guess if we want to change that in pika I would prefer to do it in a separate PR :)
The workaround in my last commit adding the const ref completion signatures seems to solve the problem

aurianer:
the errors are unrelated to that PR, merging it as is since cscs/ci passes, I'll fix things in another PR if we notice anything

